PR #11, opened 2019-02-04
DNM: samples: images: remove static bt0 IP assignment

mike-scott:
We are migrating to RADVD containers which will assign the IP
to interfaces as they are brought up.  This needs to be removed.


mike-scott:
DO NOT MERGE until we have docker-compose / doc changes up for RADVD/Jool kernel module

mike-scott:
Closing.  Turns out we can keep this around for legacy support.

PR #20, opened 2019-03-21
Update to newer version with HSM support

mbolivar:
Bring in changes which allow PKCS#11-based provisioning using an HSM.

Signed-off-by: Marti Bolivar <marti@foundries.io>


doanac:
and of course a change to this, leads to a change needed here:

 https://github.com/foundriesio/lmp-manifest/blob/master/default.xml

I've always let ricardo handle that piece.

PR #34, opened 2019-05-28
recipes-bsp: u-boot: include same patches for u-boot-tools

mike-scott:
To make sure our mkimage patches stay in sync with the u-boot patches
we need to apply them to both u-boot and u-boot-tools packages.

This way other layers can expect certain changes to be in place.


mike-scott:
@ricardosalveti how about `u-boot-lmp-common.inc`?

PR #35, opened 2019-06-10
lmp.inc: set default username/password variables

mike-scott:
These can be overriden in local.conf


mike-scott:
I changed the default from "osf" -> "fio" here.    Maybe I should keep "osf" for now?  Or should I also put up documentation changes?

mike-scott:
Btw, still testing this.

doanac:
i like the idea of "fio" but documentation and CI depend of "osf". We just need to be careful about doing this

mike-scott:
@doanac yep, put the user/pwd back to "osf" by default to avoid any complications.

mike-scott:
This can wait for @ricardosalveti to look at.  It's not critical.

PR #44, opened 2019-07-31
Support for aktualizr-lite/get

ricardosalveti:
Include the wip patches required for supporting aktualizr-lite and aktualizr-get.

Aktualizr-lite can be used by default by simply setting SOTA_CLIENT = "aktualizr-lite" in conf/local.conf.

### DIFF
--- a/recipes-samples/images/lmp-image-common.inc
+++ b/recipes-samples/images/lmp-image-common.inc
@@ -23,6 +23,7 @@ REQUIRED_DISTRO_FEATURES = "pam systemd"
 
 # Base packages
 CORE_IMAGE_BASE_INSTALL += " \
+    aktualizr-get \
     aktualizr-host-tools \
     bluetooth-attach \
     haveged \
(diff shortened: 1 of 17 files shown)


recipes-samples/images/lmp-image-common.inc
```diff
@@ -23,6 +23,7 @@ REQUIRED_DISTRO_FEATURES = "pam systemd"
 
 # Base packages
 CORE_IMAGE_BASE_INSTALL += " \
+    aktualizr-get \
```
doanac: curious why this was here, but not also aktualizr-lite?

ricardosalveti: Just because aktualizr-get is just a tool, while aktualizr-lite would end up replacing the functionality provided by aktualizr itself. The way I see we should have either aktualizr or aktualizr-lite, not both, right?

And I'm now checking if this can be set by the SOTA_CLIENT variable.

PR #45, opened 2019-07-31
Support for docker-credential-helper-fio

ricardosalveti:
As suggested by @doanac.


recipes-support/docker-credential-helper-fio/docker-credential-helper-fio_0.1.bb
```diff
@@ -0,0 +1,17 @@
+SUMMARY = "Docker-credential helper to handle hub.foundries.io for registered devices"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit allarch
+
+RDEPENDS_${PN} = "aktualizr-get"
```
doanac: Should that be rdepends or just depends?

ricardosalveti: Depends is for build time dependencies, rdepends is mostly used by the package as a runtime dependency, which is what we need here.

PR #49, opened 2019-08-31
images: implement feature includes

mike-scott:
The LmP images are not easy for subscribers to inherit and modify for
their own purposes.

Let's split them up into feature-based include files such as:
lmp-feature-bluetooth.inc
lmp-feature-docker.inc
...

These can be included at the top a subscriber custom image defined in
their custom layer as recipes-samples/images/custom-image.bb so:

```
require lmp-image-common.inc
require lmp-feature-ota.inc
require lmp-feature-docker.inc
require lmp-feature-bluetooth.inc
require lmp-feature-wifi.inc
require lmp-feature-nat64.inc
require lmp-feature-debug.inc
require lmp-feature-sbin-path-helper.inc
require lmp-feature-sysctl-hang-crash-helper.inc
require lmp-feature-sysctl-net-queue-pfifo-fast.inc
BT_6LOWPAN_NETWORK = "fe80:0:0:0:d4e7::1/80"
require lmp-feature-bt-6lowpan.inc
require lmp-service-bluetooth-disable.inc
require lmp-service-ostree-pending-reboot.inc
```

### DIFF
--- a/recipes-samples/images/lmp-feature-bluetooth.inc
+++ b/recipes-samples/images/lmp-feature-bluetooth.inc
@@ -1,0 +1,5 @@
+# Bluetooth packages
+CORE_IMAGE_BASE_INSTALL += " \
+    bluetooth-attach \
+    bluez5-noinst-tools \
+"
--- a/recipes-samples/images/lmp-feature-sysctl-net-queue-pfifo-fast.inc
+++ b/recipes-samples/images/lmp-feature-sysctl-net-queue-pfifo-fast.inc
@@ -1,0 +1,4 @@
+# packages
+CORE_IMAGE_BASE_INSTALL += " \
+    sysctl-net-queue-pfifo-fast \
+"
--- a/recipes-samples/images/lmp-gateway-image.bb
+++ b/recipes-samples/images/lmp-gateway-image.bb
@@ -2,6 +2,22 @@ SUMMARY = "Basic console-based gateway image"
 
 require lmp-image-common.inc
 
+require lmp-feature-ota.inc
+require lmp-feature-docker.inc
+require lmp-feature-bluetooth.inc
+require lmp-feature-wifi.inc
+require lmp-feature-nat64.inc
+require lmp-feature-debug.inc
+require lmp-feature-sbin-path-helper.inc
+require lmp-feature-sysctl-hang-crash-helper.inc
+require lmp-feature-sysctl-net-queue-pfifo-fast.inc
+
+BT_6LOWPAN_NETWORK = "fe80:0:0:0:d4e7::1/80"
+require lmp-feature-bt-6lowpan.inc
+
+require lmp-service-bluetooth-disable.inc
+require lmp-service-ostree-pending-reboot.inc
+
 IMAGE_FEATURES += "ssh-server-openssh"
 
 # Enough free space for a full image update
@@ -13,8 +29,6 @@ CORE_IMAGE_BASE_INSTALL += " \
     cpufrequtils \
     ldd \
     gptfdisk \
-    hostapd \
-    htop \
     iptables \
     kernel-modules \
     networkmanager-nmtui \
@@ -24,25 +38,15 @@ CORE_IMAGE_BASE_INSTALL += " \
 "
 
 CORE_IMAGE_BASE_INSTALL += " \
-    docker \
-    bluez5-noinst-tools \
     ethtool \
     git \
-    jool \
     less \
-    linux-firmware-ar3k \
-    linux-firmware-ath9k \
-    linux-firmware-ath10k \
-    linux-firmware-qca \
-    linux-firmware-wl18xx \
     openssh-sftp-server \
     packagegroup-core-full-cmdline-utils \
     packagegroup-core-full-cmdline-extended \
     packagegroup-core-full-cmdline-multiuser \
     python3-compression \
     python3-distutils \
-    python3-docker \
-    python3-docker-compose \
     python3-json \
     python3-multiprocessing \
     python3-netclient \
@@ -50,19 +54,4 @@ CORE_IMAGE_BASE_INSTALL += " \
     python3-shell \
     python3-unixadmin \
     pciutils \
-    strace \
-    tcpdump \
-    vim-tiny \
-    wpanusb \
-"
-
-fakeroot do_populate_rootfs_src () {
-    # Disable bluetooth service by default (allow to be contained in docker)
-    ln -sf /dev/null ${IMAGE_ROOTFS}/etc/systemd/system/bluetooth.service
-}
-
-IMAGE_PREPROCESS_COMMAND += "do_populate_rootfs_src; "
-
-EXTRA_USERS_PARAMS += "\
-usermod -a -G docker ${LMP_USER}; \
 "
--- a/recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
+++ b/recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
@@ -1,0 +1,36 @@
+SUMMARY = "Setup BT 6LoWPAN network / modules"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit allarch
+
+SRC_URI = " \
+	file://bt-6lowpan.network.in \
+	file://modules-6lowpan.conf \
+"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+# Allow build time customizations by the user
+BT_6LOWPAN_INTERFACE ?= "bt0"
+BT_6LOWPAN_NETWORK ?= ""
+
+do_compile() {
+	if [ ! -z "${BT_6LOWPAN_NETWORK}" ]; then
+		sed -e 's/@@BT_6LOWPAN_NETWORK@@/${BT_6LOWPAN_NETWORK}/' \
+			${S}/bt-6lowpan.network.in > bt-6lowpan.network
+	fi
+}
+
+do_install() {
+	if [ -f "${S}/bt-6lowpan.network" ]; then
+		install -d ${D}${systemd_unitdir}/network
+		install -m 0644 ${B}/bt-6lowpan.network ${D}${systemd_unitdir}/network/60-bt-6lowpan.network
+	fi
+	install -d ${D}${libdir}/modules-load.d
+	install -m 0644 ${WORKDIR}/modules-6lowpan.conf ${D}${libdir}/modules-load.d/6lowpan.conf
+}
+
+FILES_${PN} += "${libdir}/modules-load.d"
--- a/recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
+++ b/recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
@@ -1,0 +1,38 @@
+SUMMARY = "OStree Pending Reboot service"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit allarch systemd
+
+SRC_URI = " \
+	file://ostree-pending-reboot.service \
+	file://ostree-pending-reboot.timer.in \
+"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+# Value is in minutes (default to check for reboot every 5 minutes)
+OSTREE_PENDING_REBOOT_CHECK_MINUTES ?= "5"
+
+PACKAGES += "${PN}-timer"
+SYSTEMD_PACKAGES = "${PN} ${PN}-timer"
+SYSTEMD_SERVICE_${PN} = "ostree-pending-reboot.service"
+SYSTEMD_SERVICE_${PN}-timer = "ostree-pending-reboot.timer"
+SYSTEMD_AUTO_ENABLE_${PN}-timer = "enable"
+
+do_compile() {
+	sed -e 's/@@OSTREE_PENDING_REBOOT_CHECK_MINUTES@@/${OSTREE_PENDING_REBOOT_CHECK_MINUTES}/' \
+		${WORKDIR}/ostree-pending-reboot.timer.in > ostree-pending-reboot.timer
+}
+
+do_install () {
+	install -d ${D}${systemd_system_unitdir}
+	install -m 0644 ${WORKDIR}/ostree-pending-reboot.service ${D}${systemd_system_unitdir}
+	install -m 0644 ${B}/ostree-pending-reboot.timer ${D}${systemd_system_unitdir}
+}
+
+FILES_${PN} += "${systemd_system_unitdir}/ostree-pending-reboot.service"
+FILES_${PN} += "${systemd_system_unitdir}/ostree-pending-reboot.timer"
+FILES_${PN} += "${systemd_unitdir}/system-preset"
--- a/recipes-support/sbin-path-helper/sbin-path-helper_0.1.bb
+++ b/recipes-support/sbin-path-helper/sbin-path-helper_0.1.bb
@@ -1,0 +1,19 @@
+SUMMARY = "Add all sbin dirs to PATH for root user"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit allarch
+
+SRC_URI = "file://path-sbin.sh"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install () {
+	# Useful for development
+	install -d ${D}${sysconfdir}/profile.d
+	install -m 0644 ${WORKDIR}/path-sbin.sh ${D}${sysconfdir}/profile.d/path-sbin.sh
+}
+
+FILES_${PN} += "${sysconfdir}/profile.d/path-sbin.sh"
--- a/recipes-support/sysctl-hang-crash-helper/sysctl-hang-crash-helper_0.1.bb
+++ b/recipes-support/sysctl-hang-crash-helper/sysctl-hang-crash-helper_0.1.bb
@@ -1,0 +1,18 @@
+SUMMARY = "sysctl hang/crash helper settings"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit allarch
+
+SRC_URI = "file://sysctl-panic.conf"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install () {
+	install -d ${D}${libdir}/sysctl.d
+	install -m 0644 ${WORKDIR}/sysctl-panic.conf ${D}${libdir}/sysctl.d/60-panic.conf
+}
+
+FILES_${PN} += "${libdir}/sysctl.d/60-panic.conf"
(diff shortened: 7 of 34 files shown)


mike-scott:
@ricardosalveti WIP PR for splitting up the images into feature includes

recipes-samples/images/lmp-feature-bluetooth.inc
```diff
@@ -0,0 +1,24 @@
+SRC_URI += "\
```
ricardosalveti: Probably better to rename this to lmp-feature-bt-6lowpan as the only generic bluetooth piece is the part that disables the bluetooth systemd service.

recipes-samples/images/lmp-feature-bluetooth.inc
```diff
@@ -0,0 +1,24 @@
...
+    file://bt-6lowpan.network \
+    file://modules-6lowpan.conf \
+"
+
+# Bluetooth packages
+CORE_IMAGE_BASE_INSTALL += " \
+    bluetooth-attach \
```
ricardosalveti: Actually, this is also generic. Maybe split this into 2?

recipes-samples/images/lmp-gateway-image.bb
```diff
@@ -53,16 +47,4 @@ CORE_IMAGE_BASE_INSTALL += " \
     strace \
     tcpdump \
```
ricardosalveti: Can you also create another for the debug/development tools such as strace/tcpdump/vim?

ricardosalveti:
Nice, thanks for starting this.

mike-scott:
@ricardosalveti There is an existing class that I think I'd like to use instead of the direct includes: it's called packagegroup
https://github.com/openembedded/openembedded-core/blob/master/meta/classes/packagegroup.bbclass

That way images can add packagegroups.

ricardosalveti:
> @ricardosalveti There is an existing class that I think I'd like to use instead of the direct includes: it's called packagegroup
> https://github.com/openembedded/openembedded-core/blob/master/meta/classes/packagegroup.bbclass
> 
> That way images can add packagegroups.

Sure, that would be a better way but we first need to package all the configuration files that are installed via this recipe.

mike-scott:
@ricardosalveti So, I changed my mind.  Using the packagegroup feature can be added later.  For now, it seems really easy to just leave these as include files that a subscriber can use.  We can point them at the https://github.com/mike-scott/meta-lmp/tree/image-breakdown/recipes-samples/images directory to see a full list of includes.

mike-scott:
This probably isn't quite ready for prime time as it needs testing at the very least.

mike-scott:
BUGFIXES incoming in the morning.  Found a few errors during testing.

recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
```diff
@@ -0,0 +1,32 @@
...
+do_compile() {
+	if [ ! -z "${BT_6LOWPAN_NETWORK}" ]; then
+		sed -e 's/@@BT_6LOWPAN_NETWORK@@/${BT_6LOWPAN_NETWORK}/' \
+			${WORKDIR}/bt-6lowpan.network.in > ${WORKDIR}/bt-6lowpan.network
+	fi
+}
+
+do_install() {
+	if [ -f "${WORKDIR}/bt-6lowpan.network" ]; then
+		install -m 0644 ${WORKDIR}/bt-6lowpan.network ${IMAGE_ROOTFS}${exec_prefix}/lib/systemd/network/60-bt-6lowpan.network
```
ricardosalveti: When installing just install to ${D}${systemd_unitdir}/network... instead, as this is part of the recipe (it was before in IMAGE_ROOTFS because it was part of the recipe responsible for building the image).

recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
```diff
@@ -0,0 +1,32 @@
...
+	fi
+	install -m 0644 ${WORKDIR}/modules-6lowpan.conf ${IMAGE_ROOTFS}${exec_prefix}/lib/modules-load.d/6lowpan.conf
```
ricardosalveti: Same here but to ${D}${libdir}/modules-load.... instead.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,23 @@
+SUMMARY = "OStree Pending Reboot service"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit allarch systemd
+
+RDEPENDS_${PN} = "aktualizr-get" # TODO: ??
```
ricardosalveti: The service itself does not require aktualizr, it just acts due an action caused by it.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,23 @@
...
+
+SRC_URI = "file://ostree-pending-reboot.service"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+SYSTEMD_SERVICE_${PN} = "ostree-pending-reboot.service"
+SYSTEMD_AUTO_ENABLE_${PN} = "enable"
```
ricardosalveti: So this service was initially created to be used together with a systemd timer job, so it shouldn't really be started/enabled automatically (as it will not even start due ConditionPathExists being false during boot).

You could have a timer created and installer by this recipe as well, and enable that one by default (or via configs that can be set by the user).

recipes-support/sbin-path-helper/sbin-path-helper_0.1.bb
```diff
@@ -0,0 +1,17 @@
+SUMMARY = " Add all sbin dirs to $PATH for root user"
```
ricardosalveti: Extra espace before 'Add', probably good to also remove $ from PATH here to avoid it being converted during parsing.

recipes-support/sbin-path-helper/sbin-path-helper_0.1.bb
```diff
@@ -0,0 +1,17 @@
...
+
+SRC_URI = "file://path-sbin.sh"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install () {
+	# Useful for development
+	install -d ${IMAGE_ROOTFS}${sysconfdir}/profile.d
```
ricardosalveti: As with the 6lowpan recipe, install at ${D}${sysconfdir} instead.

recipes-support/sysctl-hang-crash-helper/sysctl-hang-crash-helper_0.1.bb
```diff
@@ -0,0 +1,15 @@
...
+inherit allarch
+
+SRC_URI = "file://sysctl-panic.conf"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install () {
+	install -m 0644 ${WORKDIR}/sysctl-panic.conf ${IMAGE_ROOTFS}${exec_prefix}/lib/sysctl.d/60-panic.conf
```
ricardosalveti: Better to install the file at '/usr/lib/sysctl.d' (${D}${libdir}/sysctl.d...)

Make sure to create that folder first as well.

recipes-support/sysctl-net-queue-pfifo-fast/sysctl-net-queue-pfifo-fast_0.1.bb
```diff
@@ -0,0 +1,15 @@
...
+inherit allarch
+
+SRC_URI = "file://sysctl-net-queuing.conf"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+do_install () {
+	install -m 0644 ${WORKDIR}/sysctl-net-queuing.conf ${IMAGE_ROOTFS}${exec_prefix}/lib/sysctl.d/90-net-queuing.conf
```
ricardosalveti: Same as I said for the hang crash recipe.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
mike-scott: I think we wanted this to be enabled / started by default if the user image chooses it.  So I'll get the timer portion in as well here.

ricardosalveti: What you want is the timer to be enabled by default, the reboot.service one doesn't require that as it gets triggered by the timer job.

mike-scott:
@ricardosalveti Fixed the patches:
- Moved previous fixes into their respective patches
- Removed the use of ${IMAGE_ROOTFS} and moved to standard ${S} -> ${D} adding FILES_${PN} entries where needed.
- Added ostree-pending-reboot.timer.in template which is customized via the OSTREE_PENDING_REBOOT_CHECK_MINUTES variable.

recipes-samples/images/lmp-gateway-image.bb
```diff
@@ -24,45 +38,20 @@ CORE_IMAGE_BASE_INSTALL += " \
...
     python3-compression \
     python3-distutils \
-    python3-docker \
-    python3-docker-compose \
     python3-json \
     python3-multiprocessing \
     python3-netclient \
     python3-pkgutil \
     python3-shell \
     python3-unixadmin \
```
mike-scott: I was curious if we could categorize these python3 modules?

recipes-samples/images/lmp-feature-sysctl-net-queue-pfifo-fast.inc
```diff
@@ -0,0 +1,4 @@
+# packages
+CORE_IMAGE_BASE_INSTALL += " \
+    sysctl-net-queue-pfifo-fast \
+"
```
mike-scott: We could rename this include to something like:
`lmp-feature-sysctl-set-default-packet-scheduler.inc`

With a default in the package of something like this:
`LMP_DEFAULT_PACKET_SCHEDULER ?= "fq_codel"`

For the gateway image the include would look like this:
```
...
require lmp-feature-sysctl-hang-crash-helper.inc

LMP_DEFAULT_PACKET_SCHEDULER = "pfifo_fast"
require lmp-feature-sysctl-set-default-packet-scheduler.inc
```

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,35 @@
...
+}
+
+do_install () {
+	install -d ${D}${systemd_system_unitdir}
+	install -m 0644 ${S}/ostree-pending-reboot.service ${D}${systemd_system_unitdir}
+	install -m 0644 ${S}/ostree-pending-reboot.timer ${D}${systemd_system_unitdir}
+}
+
+FILES_${PN} += "${systemd_system_unitdir}/ostree-pending-reboot.service"
+FILES_${PN}-timer += "${systemd_system_unitdir}/ostree-pending-reboot.timer"
```
mike-scott: Dang.  This doesn't work.  Will need to fix that.

mike-scott: Fixed in latest version

mike-scott:
- Rebased
- Fixed the ostree-pending-reboot patch

recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
```diff
@@ -0,0 +1,34 @@
...
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+# Allow build time customizations by the user
+BT_6LOWPAN_INTERFACE ?= "bt0"
+BT_6LOWPAN_NETWORK ?= ""
+
+do_compile() {
+	if [ ! -z "${BT_6LOWPAN_NETWORK}" ]; then
+		sed -e 's/@@BT_6LOWPAN_NETWORK@@/${BT_6LOWPAN_NETWORK}/' \
+			${S}/bt-6lowpan.network.in > ${S}/bt-6lowpan.network
```
ricardosalveti: Here you don't need the final ${S} as the function is already called at the right folder.

recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
```diff
@@ -0,0 +1,34 @@
...
+	fi
+}
+
+do_install() {
+	if [ -f "${S}/bt-6lowpan.network" ]; then
+		install -d ${D}${systemd_unitdir}/network
+		install -m 0644 ${S}/bt-6lowpan.network ${D}${systemd_unitdir}/network/60-bt-6lowpan.network
```
ricardosalveti: Here it is better to use ${B} instead of ${S} as that is the variable set for base files produced during the build step.

recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
```diff
@@ -0,0 +1,34 @@
...
+	fi
+	install -d ${D}${libdir}/modules-load.d
+	install -m 0644 ${S}/modules-6lowpan.conf ${D}${libdir}/modules-load.d/6lowpan.conf
```
ricardosalveti: Here since modules-6lowpan.conf gets provided as a file, better to use ${WORKDIR} instead (${S} works the same way but it can later mean a different thing if the recipe starts fetching sources from a git repository, for example).

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,39 @@
...
+"
+
+S = "${WORKDIR}"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+# Value is in minutes (default to check for reboot every 1 minute)
+OSTREE_PENDING_REBOOT_CHECK_MINUTES ?= "1"
+
+PACKAGES = "${PN} ${PN}-timer"
```
ricardosalveti: PN is already set into PACKAGES by default, so better use 'PACKAGES += "${PN}-timer"' instead.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,39 @@
...
+SYSTEMD_PACKAGES = "${PN} ${PN}-timer"
+SYSTEMD_SERVICE_${PN} = "ostree-pending-reboot.service"
+SYSTEMD_AUTO_ENABLE_${PN} = "disable"
```
ricardosalveti: I don't think we need to manually disable the pending-reboot.service.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,39 @@
...
+
+do_compile() {
+	sed -e 's/@@OSTREE_PENDING_REBOOT_CHECK_MINUTES@@/${OSTREE_PENDING_REBOOT_CHECK_MINUTES}/' \
+		${WORKDIR}/ostree-pending-reboot.timer.in > ${WORKDIR}/ostree-pending-reboot.timer
+}
+
+do_install () {
+	install -d ${D}${systemd_system_unitdir}
+	install -m 0644 ${S}/ostree-pending-reboot.service ${D}${systemd_system_unitdir}
+	install -m 0644 ${S}/ostree-pending-reboot.timer ${D}${systemd_system_unitdir}
```
ricardosalveti: Use ${B}/ostree-pending-reboot.timer and ${WORKDIR}/ostree-pending-reboot.service instead.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,39 @@
...
+PACKAGES = "${PN} ${PN}-timer"
+SYSTEMD_PACKAGES = "${PN} ${PN}-timer"
+SYSTEMD_SERVICE_${PN} = "ostree-pending-reboot.service"
+SYSTEMD_AUTO_ENABLE_${PN} = "disable"
+SYSTEMD_SERVICE_${PN}-timer = "ostree-pending-reboot.timer"
+SYSTEMD_AUTO_ENABLE_${PN}-timer = "enable"
+
+do_compile() {
+	sed -e 's/@@OSTREE_PENDING_REBOOT_CHECK_MINUTES@@/${OSTREE_PENDING_REBOOT_CHECK_MINUTES}/' \
+		${WORKDIR}/ostree-pending-reboot.timer.in > ${WORKDIR}/ostree-pending-reboot.timer
```
ricardosalveti: Also no need to > ${WORKDIR}, just to the file itself is fine.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
```diff
@@ -0,0 +1,39 @@
...
+
+do_install () {
+	install -d ${D}${systemd_system_unitdir}
+	install -m 0644 ${S}/ostree-pending-reboot.service ${D}${systemd_system_unitdir}
+	install -m 0644 ${S}/ostree-pending-reboot.timer ${D}${systemd_system_unitdir}
+}
+
+FILES_${PN} += "${systemd_system_unitdir}/ostree-pending-reboot.service"
+FILES_${PN} += "${systemd_system_unitdir}/ostree-pending-reboot.timer"
+FILES_${PN} += "${systemd_unitdir}/system-preset"
```
ricardosalveti: I believe this is not needed, since it gets added by systemd.bbclass.

recipes-samples/images/lmp-gateway-image.bb
ricardosalveti: Most of them are there because of ansible (iirc sftp-server as well), so we might want to split those to a different file at least.

recipes-support/ostree-pending-reboot/ostree-pending-reboot_0.1.bb
mike-scott: I was getting an error here due to files ending up in the image that weren't "used" by any packages.. so I added this to fix that.

recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
mike-scott: If that's the case, do you want me to change to ${WORKDIR} for the following lines in sbin-path-helper, sysctl-hang-crash-helper and sysctl-net-queue-pfifo-fast? (these are provided as files):
`install -m 0644 ${S}/path-sbin.sh ${D}${sysconfdir}/profile.d/path-sbin.sh`
`install -m 0644 ${S}/sysctl-panic.conf ${D}${libdir}/sysctl.d/60-panic.conf`
`install -m 0644 ${S}/sysctl-net-queuing.conf ${D}${libdir}/sysctl.d/90-net-queuing.conf`

mike-scott:
Updated per @ricardosalveti comments.  Here's the diff: https://pastebin.com/raw/agysHzjJ

recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
mike-scott: I went ahead and followed the same methodology for the above in the new patchset.

mike-scott:
Updated to 5 minute default reboot check.  Diff: https://pastebin.com/raw/NqUShBZb

mike-scott:
Updated to fix build issue w/ bt-6lowpan-setup package.
Diff:
```
diff --git a/recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb b/recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
index 19d2fcf..331a94e 100644
--- a/recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
+++ b/recipes-support/bt-6lowpan-setup/bt-6lowpan-setup_0.1.bb
@@ -32,3 +32,5 @@ do_install() {
 	install -d ${D}${libdir}/modules-load.d
 	install -m 0644 ${WORKDIR}/modules-6lowpan.conf ${D}${libdir}/modules-load.d/6lowpan.conf
 }
+
+FILES_${PN} += "${libdir}/modules-load.d"
```

PR #51, opened 2019-09-03
lmp-machine-custom changes for i.MX7ULP

mike-scott:
- Don't set unknown SoCs OSTREE kernel args by default
- Add i.MX7ULP OSTREE kernel arg defaults


mike-scott:
@ricardosalveti I feel like we should be able to setup the console entry in a better way (from the SoC SERIAL_CONSOLES define) but for now this fixes the console output for i.MX7ULP.

PR #53, opened 2019-09-13
Changes for the 47 update

ricardosalveti:
Making LmP compatible with latest OE-Core/Meta-OE and other layers.

This also introduces u-boot-fio, which is currently just used by apalis-imx6.

Docker-app also included as part of the build, and the switch to aktualizr-lite by default.


ricardosalveti:
@doanac I believe only docker-app that needs better testing, rest should be good to go.

doanac:
@rsalveti - looks good. I'm testing with "docker-app app render <docker app file>" which is what we use inside aktualizr and aktualizr-lite

doanac:
@rsalveti - one sec. it worked on my minnowboard, but I have an rpi3-64 issue:
~~~
docker-app app render /tmp/shellhttpd.dockerapp 
render failed: standard_init_linux.go:211: exec user process caused "exec format error"
~~~

recipes-sota/lmp-device-register/lmp-device-register_git.bb
```diff
@@ -4,7 +4,7 @@ LIC_FILES_CHKSUM = "file://COPYING.MIT;md5=838c366f69b72c5df05c96dff79b35f2"
 
 DEPENDS = "boost curl ostree glib-2.0"
 
-SRCREV = "579afb7488534acd04d7206ce39422553d62a801"
+SRCREV = "670ab2d040df3413938377816fce8b8933a4be91"
```
mike-scott: This doesn't include @doanac latest changes.  Is that ok?

ricardosalveti: Yes, asked that to @doanac earlier today and he wanted to update that on another update.

PR #54, opened 2019-09-16
update kernel to v5.3

mike-scott:
- Dropped the downstream RPi kernel tree for RPi4
- Update linux-lmp to v5.3

### DIFF
--- a/classes/lmp-machine-custom.bbclass
+++ b/classes/lmp-machine-custom.bbclass
@@ -34,12 +34,6 @@ OSTREE_KERNEL_ARGS_raspberrypi4_sota ?= "video=HDMI-A-1:1280x720@60 cma=256M vc_
 ## U-Boot entrypoints for rpi
 UBOOT_DTB_LOADADDRESS_rpi = "0x02600000"
 UBOOT_DTBO_LOADADDRESS_rpi = "0x026d0000"
-## RPI4: Force rpi upstream kernel for now until it is in a better shape
-PREFERRED_PROVIDER_virtual/kernel_raspberrypi4 = "linux-lmp-dev"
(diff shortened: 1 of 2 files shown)


mike-scott:
@ricardosalveti Just getting this started for v48.

NOTE: the u-boot script is hard-coded to copy 0x1000 of the DTB memory.  Looking around for a better option to generate the size.

recipes-bsp/u-boot/u-boot-ostree-scr/qemuriscv64/boot.cmd
```diff
@@ -4,6 +4,12 @@ then
     exit
 fi
 
+# Recent kernel commit in 5.3:
+# https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/arch/riscv/kernel?h=v5.3&id=671f9a3e2e24cdeb2d2856abee7422f093e23e29
+# changed page table setup which now crushes the qemu loaded DTB.
+# Let's copy the DTB to the established load addr instead where it's safe.
+cp ${fdt_addr} ${fdt_addr_r} 1000
```
mike-scott: Ugh.  I actually don't like this being here.  This file won't get updated from version to version.  I just tried an update from 621 -> 627 and it didn't work because this file stays outside of the rootfs and isn't managed by OTA.

mike-scott: Nm, since this is only for QEMU riscv targets atm, this is probably fine.  Noting this in a JIRA.

ricardosalveti:
Closing as this is now part of #90

PR #62, opened 2019-09-19
conf/distro: introduce lmp-base distro configuration

mike-scott:
The LmP base distro should be used when no ostree/aktualizr functionality
is needed by an image build.

Example of this would be for manufacturing tools, etc.

### DIFF
--- a/conf/distro/lmp-base.conf
+++ b/conf/distro/lmp-base.conf
@@ -1,0 +1,20 @@
+require conf/distro/include/lmp.inc
+
+DISTRO = "lmp-base"
+DISTROOVERRIDES = "lmp:lmp-base"
+DISTRO_NAME = "Linux-microPlatform Base (no ostree)"
+
+IMAGE_LINGUAS ?= "en-us"
+
+# No graphical feature as part of the base platform
+DISTRO_FEATURES_remove = "wayland x11"
+
+# By default we only support initramfs. We don't build live as that
+# pulls in a lot of dependencies for the live image and the installer, like
+# udev, grub, etc.  These pull in gettext, which fails to build with wide
+# character support.
+IMAGE_FSTYPES = "cpio.gz"
+QB_DEFAULT_FSTYPE = "${IMAGE_FSTYPES}"
+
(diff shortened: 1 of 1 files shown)


conf/distro/lmp-base.conf
```diff
@@ -0,0 +1,18 @@
...
+DISTRO_FEATURES_remove = "wayland x11"
+
+# By default we only support initramfs. We don't build live as that
+# pulls in a lot of dependencies for the live image and the installer, like
+# udev, grub, etc.  These pull in gettext, which fails to build with wide
+# character support.
+IMAGE_FSTYPES = "cpio.gz"
+QB_DEFAULT_FSTYPE = "${IMAGE_FSTYPES}"
+
+MACHINE_ESSENTIAL_EXTRA_RDEPENDS = ""
```
mike-scott: I should probably comment here something like:
By default we don't have any extra machine dependencies

conf/distro/lmp-base.conf
```diff
@@ -0,0 +1,18 @@
...
+IMAGE_LINGUAS ?= "en-us"
+
+# No graphical feature as part of the base platform
+DISTRO_FEATURES_remove = "wayland x11"
+
+# By default we only support initramfs. We don't build live as that
+# pulls in a lot of dependencies for the live image and the installer, like
+# udev, grub, etc.  These pull in gettext, which fails to build with wide
+# character support.
+IMAGE_FSTYPES = "cpio.gz"
```
mike-scott: @ricardosalveti is this ok for you?  It was semi-specific to the mfgtool setup.

conf/distro/lmp-base.conf
```diff
@@ -0,0 +1,18 @@
+require conf/distro/include/lmp.inc
+
+DISTRO = "lmp-base"
```
ricardosalveti: Can you please add 'DISTROOVERRIDES = "lmp:lmp-base"' here as well? That way the lmp-base distro will also use the lmp specific overrides.

mike-scott: will do

PR #66, opened 2019-09-25
images/feature-debug: add more debug tools

mike-scott:
To help with HW bringup and debugging let's add the following
tools to the debug feature:
- screen
- tmux
- minicom
- devmem2

### DIFF
--- a/recipes-samples/images/lmp-feature-debug.inc
+++ b/recipes-samples/images/lmp-feature-debug.inc
@@ -4,4 +4,8 @@ CORE_IMAGE_BASE_INSTALL += " \
     strace \
     tcpdump \
     vim-tiny \
+    screen \
+    tmux \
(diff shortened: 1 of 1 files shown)


mike-scott:
My thoughts here were to have a serial tool that the user is familiar with (we offer several), add a package manager to enable on-the-fly loading of packages, and devmem2 to allow for reading registers etc.

doanac:
handy - but have we considered making a "dev" container instead? Just to help push people in the "containers can do it all" direction?

mike-scott:
@doanac you would have to have that container pre-installed as in some cases you're debugging why the network isn't working, etc.

doanac:
good point.

mike-scott:
@rsalveti removed opkg

PR #73, opened 2019-09-27
lmp-feature-factory: add docker-cli-config

ricardosalveti:
hub.foundries.io docker cli configuration is required for factory users.


doanac:
seems easier than the other approach we discussed

PR #79, opened 2019-10-11
lmp-machine-custom: rpi: move to downstream kernel

mike-scott:
In an effort to make supporting new kernels on our reference hardware
easier, we are dropping all of the RPi patches from the FIO upstream
kernel.  Let's move the RPi targets back to the current 4.19 downstream
kernel.


ricardosalveti:
Closing as this is now part of #90

PR #85, opened 2019-10-16
python3-docker: Add support for our credential helper

doanac:
Our credential helper configuration for the LMP is stored under
/usr/lib/docker/config.json. This adds the path to the docker API
so that docker-compose can use this helper.

### DIFF
--- a/recipes-devtools/python/python3-docker/0001-config-Include-usr-lib-docker-in-search-path.patch
+++ b/recipes-devtools/python/python3-docker/0001-config-Include-usr-lib-docker-in-search-path.patch
@@ -1,0 +1,25 @@
+From 742156db38a81d17155870a9614e52f37fa90d1e Mon Sep 17 00:00:00 2001
+From: Andy Doan <andy@foundries.io>
+Date: Wed, 16 Oct 2019 13:47:30 -0500
+Subject: [PATCH] config: Include /usr/lib/docker in search path
+
+Signed-off-by: Andy Doan <andy@foundries.io>
+---
+ docker/utils/config.py | 1 +
+ 1 file changed, 1 insertion(+)
+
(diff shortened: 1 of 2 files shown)


doanac:
gonna try a quick test in my factory to verify

recipes-devtools/python/python3-docker/0001-config-Include-usr-lib-docker-in-search-path.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/docker/utils/config.py b/docker/utils/config.py
+index 82a0e2a..d382f52 100644
+--- a/docker/utils/config.py
++++ b/docker/utils/config.py
+@@ -12,6 +12,7 @@ log = logging.getLogger(__name__)
+ 
+ def find_config_file(config_path=None):
+     paths = list(filter(None, [
++        '/usr/lib/docker', # 0
+         config_path,  # 1
```
mike-scott: Where does it normally look for the config?  IE: `config_path` and `config_path_from_environment`?  Is there a way we can just symlink our config to one of those 2 places?

doanac: @ricardosalveti originally set the location. I think he chose /usr/lib/docker because its immutable. So that then makes us update docker related client code since its not standard in normal desktops/servers.

ricardosalveti: Right, by default there is no system wide config, which is why we had to define ours.

mike-scott:
If @ricardosalveti is happy then so am I.

doanac:
docker-compose pull works in my factory now.

PR #90, opened 2019-10-21
Cleanups and Zeus updates

ricardosalveti:
* Making meta-lmp compatible with latest OE-zeus
* New kernel-lmp-fitimage class created to extend the upstream kernel-fitimage one, supporting external device tree and overlays as a start (later it will also be used for additional features)
* Migrate BSP settings from meta-updater to meta-lmp, in order to have full control of the options used
* RPi migrated to the downstream kernel
* Linux-lmp updated to 5.3


ricardosalveti:
Testing done as part of https://github.com/foundriesio/lmp-manifest/pull/40 (which is consuming meta-lmp from the same orig tree/branch).

PR #92, opened 2019-10-24
kernel-lmp-fitimage: add support for fpga

ricardosalveti:
Add support to store and load fpga binaries as part of the kernel
fitimage.

FPGA support depends on FPGA_BINARY and FPGA_LOADADDRESS, which needs to
be defined by the user.

### DIFF
--- a/classes/kernel-lmp-fitimage.bbclass
+++ b/classes/kernel-lmp-fitimage.bbclass
@@ -2,6 +2,9 @@
 
 inherit kernel-fitimage
 
+# Default value for deployment filenames
+FPGA_BINARY ?= "fpga.bin"
+
 #
 # Emit the fitImage ITS configuration section
 #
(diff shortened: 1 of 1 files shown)


ricardosalveti:
To make it work reliably the layer will have to extend the linux recipe and add "do_compile[depends] += " fpga-recipe:do_deploy"", besides defining FPGA_BINARY and FPGA_LOADADDRESS.

classes/kernel-lmp-fitimage.bbclass
```diff
@@ -2,15 +2,19 @@
...
+FPGA_BINARY ?= "fpga.bin"
+
 #
 # Emit the fitImage ITS configuration section
 #
 # $1 ... .its filename
 # $2 ... Linux kernel ID
 # $3 ... DTB image name
 # $4 ... ramdisk ID
-# $5 ... config ID
```
mike-scott: Are we dropping config ID? Or are we calling that Setup ID now?

ricardosalveti: Original comment was wrong, $5 is actually setup id.

mike-scott: ok

PR #111, opened 2019-11-28
fiovb host

ldts:
Use fiovb_printenv and fiovb_setenv to access the secure functionality (just as fw_printenv and fw_sentenv).

This PR also creates the necessary symlinks during the install phase

requires actualzr update.

### DIFF
--- a/recipes-security/optee/optee-fiovb_git.bb
+++ b/recipes-security/optee/optee-fiovb_git.bb
@@ -26,5 +26,7 @@ EXTRA_OEMAKE = "OPTEE_CLIENT_EXPORT=${OPTEE_CLIENT_EXPORT} \
 
 do_install () {
     install -d ${D}${bindir}
-    install -m 0755 ${S}/out/ca/fiovb ${D}${bindir}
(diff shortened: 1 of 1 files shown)


recipes-security/optee/optee-fiovb_git.bb
```diff
@@ -27,4 +27,6 @@ EXTRA_OEMAKE = "OPTEE_CLIENT_EXPORT=${OPTEE_CLIENT_EXPORT} \
 do_install () {
     install -d ${D}${bindir}
     install -m 0755 ${S}/out/ca/fiovb ${D}${bindir}
+    ln -sf ${D}${bindir}fiovb ${D}${bindir}/fiovb_printenv
+    ln -sf ${D}${bindir}fiovb ${D}${bindir}/fiovb_setenv
```
ricardosalveti: Better to do a relative link instead for both here, as they will all be under the same root folder.

ldts: ok. maybe I will just do the same installation process that  other metas did (ie, meta-freescale...renaming the binary)

PR #114, opened 2019-12-02
General updates for 58

ricardosalveti:
* Bump aktualizr to 2019.10
* Rollback support add for beaglebone and rpi targets
* Latest kernel stable updates


recipes-sota/aktualizr/aktualizr-fiovb-env-rollback.bb
```diff
@@ -0,0 +1,23 @@
...
+
+do_install() {
+    install -m 0700 -d ${D}${libdir}/sota/conf.d
+    install -m 0644 ${WORKDIR}/sota-fiovb-env.toml ${D}${libdir}/sota/conf.d/30-rollback.toml
+}
+
+FILES_${PN} = " \
+    ${libdir}/sota/conf.d \
+    ${libdir}/sota/conf.d/30-rollback.toml \
+"
```
doanac: so this is an opt-in thing, or its enabled by default?

otherwise it looks good to me from an aktualizr standpoint

ricardosalveti: It depends on what is available as part of SOTA_CLIENT_FEATURES, which can be extended depending on the machine used.

Same thing as we have for the u-boot env logic, it needs ubootenv to be part of SOTA_CLIENT_FEATURES, which is currently only set at beagle and rpi (part of this same pr).

PR #120, opened 2019-12-06
optee-os: imx: psci reset through watchdog

ldts:
the psci system reset is delegated to a watchdog needing external initialization (usually by the linux kernel). Since linux could disable the watchdog at any time before the actual reset sequence (ie, systemd requests), optee needs to guarantee that the watchdog will be ready to process the psci request.
This is achieved by
1) making sure the watchdog clock is active during device init - so no normal world initialization is required.
2) reserving the watchdog via overlay and dtb generation - so the kernel will not probe it and create the device node.

### DIFF
--- a/recipes-security/optee/optee-os/0001-imx-imx7ulp-watchdog-set-into-ready-state-prior-rese.patch
+++ b/recipes-security/optee/optee-os/0001-imx-imx7ulp-watchdog-set-into-ready-state-prior-rese.patch
@@ -1,0 +1,52 @@
+From 29fe1792700218bdb22acff46c385bc3ffddd2de Mon Sep 17 00:00:00 2001
+From: Jorge Ramirez-Ortiz <jorge@foundries.io>
+Date: Fri, 6 Dec 2019 12:38:53 +0100
+Subject: [PATCH] imx: imx7ulp: watchdog: set into ready state prior reset
+
+Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>
+---
+ core/drivers/imx_wdog.c | 21 +++++++++++++++++++++
+ 1 file changed, 21 insertions(+)
+
+diff --git a/core/drivers/imx_wdog.c b/core/drivers/imx_wdog.c
+index 83f461c2..72c8841b 100644
+--- a/core/drivers/imx_wdog.c
++++ b/core/drivers/imx_wdog.c
+@@ -44,6 +44,24 @@
+ static bool ext_reset;
+ static vaddr_t wdog_base;
+ 
++#ifdef CFG_MX7ULP
++static void imx_wdt_reinit(void)
++{
++	uint32_t val;
++
++	/* unlock the wdog for reconfiguration */
++	io_write32(wdog_base + WDOG_CNT, UNLOCK_SEQ0);
++	io_write32(wdog_base + WDOG_CNT, UNLOCK_SEQ1);
++
++	/* set an initial timeout value in TOVAL */
++	io_write32(wdog_base + WDOG_TOVAL, 1000);
++
++	/* enable 32bit command sequence and reconfigure */
++	val = (1 << 13) | (1 << 8) | (1 << 5);
++	io_write32(wdog_base + WDOG_CS, val);
++}
++#endif
++
+ void imx_wdog_restart(void)
+ {
+ 	uint32_t val;
+@@ -54,6 +72,9 @@ void imx_wdog_restart(void)
+ 	}
+ 
+ #ifdef CFG_MX7ULP
++	/* make sure the wdt is in the ready state */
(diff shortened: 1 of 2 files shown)


ldts:
another way we could work around this inconsistency in OPTEE is if we allow the kernel driver to go through the probe function -so clocks/registers are initialized- but then return -EPERM if secure-status is defined in the node (so the device is not registered).

[1] https://github.com/torvalds/linux/blob/master/Documentation/devicetree/bindings/arm/secure.txt

ldts:
regardless of optee I think we should have this fix in anyways. the current driver is just not right.

recipes-security/optee/optee-os/0001-imx7ulp-watchdog-add-initialization-function.patch
```diff
@@ -0,0 +1,64 @@
+From 3614eb7282e35b3e2ac02aa5b58424b03ab0d5a0 Mon Sep 17 00:00:00 2001
+From: Jorge Ramirez-Ortiz <jorge@foundries.io>
+Date: Tue, 10 Dec 2019 21:21:02 +0100
+Subject: [PATCH] imx7ulp: watchdog: add initialization function.
+
+atarting the watchdog clock allows it to operate without any normal
```
ricardosalveti: s/atarting/starting/g

recipes-security/optee/optee-os/0001-imx7ulp-watchdog-add-initialization-function.patch
```diff
@@ -0,0 +1,64 @@
...
+--- a/core/drivers/imx_wdog.c
++++ b/core/drivers/imx_wdog.c
+@@ -44,6 +44,20 @@
+ static bool ext_reset;
+ static vaddr_t wdog_base;
+ 
++#ifdef CFG_MX7ULP
++static void imx_wdog_initialize(void)
++{
++	/* unlock the wdog for reconfiguration */
```
ricardosalveti: Can you add an extra check for wdog_base here? Just to be safe in case code ends up moving around later on.

recipes-security/optee/optee-os/0001-imx7ulp-reserve-a-watchdog-for-OP-TEE.patch
```diff
@@ -0,0 +1,42 @@
...
++++ b/core/arch/arm/plat-imx/main.c
+@@ -140,6 +140,22 @@ void console_init(void)
+ #endif
+ }
+ 
++/* reserve the following drivers for exclusive OP-TEE access */
++ const char *main_get_optee_exclusive_node_name(unsigned int i)
++{
++	static const char *exclusive_drivers[] = {
++#if defined CFG_MX7ULP
```
ricardosalveti: Care to add a comment here explaining why this driver needs to be handled by optee only?

recipes-security/optee/optee-os/0001-arm-generic-allow-platforms-to-reserve-devices-for-e.patch
```diff
@@ -0,0 +1,199 @@
...
++static int add_descriptor_node(struct dt_descriptor *dt, const char *driver)
++{
++	unsigned int i, j;
++	char root_name[50];
++	char buffer[100];
++	char *item[10];
++	char *root;
++	int offs;
++
++	/* strok will modify the buffer */
```
ricardosalveti: s/strok/strtok/g

recipes-security/optee/optee-os/0001-arm-generic-allow-platforms-to-reserve-devices-for-e.patch
```diff
@@ -0,0 +1,199 @@
...
++	}
++
++	if (i >= ARRAY_SIZE(item))
++		return -1;
++
++	if (fdt_path_offset(dt->blob, driver) >= 0)
++		return 0;
++
++	/* add the root node name */
++	snprintf(root_name, 50, "/%s", root);
```
ricardosalveti: Better to use sizeof instead of hardcoding 50.

recipes-security/optee/optee-os/0001-arm-generic-allow-platforms-to-reserve-devices-for-e.patch
```diff
@@ -0,0 +1,199 @@
...
++
++	return start;
++}
++
++static int add_descriptor_node(struct dt_descriptor *dt, const char *driver)
++{
++	unsigned int i, j;
++	char root_name[50];
++	char buffer[100];
++	char *item[10];
```
ricardosalveti: Care to add a comment on how you decided to set up the sizes for all these items?

PR #123, opened 2019-12-10
Drop some recipes already in meta virtualization layer

liuming50:
The docker-compose recipe in meta-virtualization has been upgraded to 1.25.0, so we could drop some recipes from meta-lmp.


ricardosalveti:
Merged manually due the new meta-lmp-base layer.

Thanks for the patches and for upstreaming the recipes!

PR #128, opened 2019-12-17
meta-lmp-base: optee-os: imx7ulp: fix incorrect register setting

ldts:
During the previous patch preparation we incorrectly wrote to BIT(6) -
interrupt status flag - instead of BIT(5) (enable support for 32bit
timers)

Upon stress testing this was causing that "sometimes" reset requests
would act inmediately instead of via countdown (perhaps due to a
violation in writing to this register).

This commit fixes the situation generated by our previous commit.


ldts:
@ricardosalveti @mike-scott during these morning tests I noticed that uboot psci resets would _sometimes_ not trigger the watchdog  countdown and just reset the board (as we saw before when register access violations were involved). This commit fixes the error I introduced in my previous commit (consequence of the clean up before creating the PR and the minimal unit testing I did after that cleanup)

PR #135, opened 2020-01-20
aktualizr: relies on nss-lookup.target

liuming50:
aktualizr-lite.service needs run after nss-lookup.target to ensure DNS
resolving is ready.

### DIFF
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr/aktualizr-lite.service
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr/aktualizr-lite.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=Aktualizr Lite SOTA Client
-After=network.target
+After=network.target nss-lookup.target
 ConditionPathExists=|/var/sota/sota.toml
 ConditionPathExists=|/usr/lib/sota/conf.d/10-lite-public-stream.toml
 


meta-lmp-base/recipes-sota/aktualizr/aktualizr/aktualizr-lite.service
```diff
@@ -1,6 +1,6 @@
 [Unit]
 Description=Aktualizr Lite SOTA Client
-After=network.target
+After=network.target nss-lookup.target
```
ricardosalveti: My only concern here is that nss-lookup.target will have a side effect similar to network-online, which blocks the job to be executed until network is available and configured.

The problem for aktualizr itself is that the bootloader update confirmation process gets executed as part of aktualizr-lite, and in case the device ends up with a broken network setup (caused by an external change for example), it will not be able to confirm the update, making it possible to end up rolling back to the previous image even when it shouldn't.

Not sure if we can improve the situation via the systemd job, maybe changing aktualizr-lite instead to make it retry reaching to the remote server more often (until it is able to reach it at least once)? Or even receiving the signal that the network is fully configured so it can retry the update part?

@doanac might also have other ideas here.

liuming50: Yes, that is a good point, it'd better to be fixed inside aktualizr.

ricardosalveti: Will close this pull request then, as we should provide the fix via aktualizr itself.

liuming50: @ricardosalveti Yes, I am fine with closing it, but would like to mention that I had sent a same patch to meta-updater, they accepted and merged it, have you managed time to ping aktualizr guys about your solution?

ricardosalveti: Hm, the same issue I described here would also happen with upstream aktualizr, as the bootloader confirmation logic gets executed as aktualizr starts, and that would never happen without a valid network in place.

This is not a critical issue for us in ak-lite when running in daemon mode as it will retry after a specific amount of time, which should work fine if networks is functional at that point. The change we should make is just to try more often until it is able to reach the remote server at least once (or when it gets a dns/network error).

PR #137, opened 2020-01-25
Opee drop python2 support

liuming50:


ricardosalveti:
Great, thanks for the cleanup, giving it a test locally now.

meta-lmp-base/recipes-security/optee/optee-os/0001-scripts-convert-remainging-scripts-to-python3.patch
```diff
@@ -0,0 +1,227 @@
+From bbaeed4dc6258006e846543197b8aff95d80abbf Mon Sep 17 00:00:00 2001
```
mike-scott: We're using a version of OP-TEE that already has python3 compliant scripts.  I don't think this additional patch is needed.

ricardosalveti: Not every script was converted (based on the hash we used), so we might indeed need the patch to make sure every python script is compatible with python 3.

mike-scott: You are quite right.   I was looking at the wrong hash.  We do need this.  Apologies.

PR #139, opened 2020-02-18
base: optee-os: use fio tree

ldts:


ldts:
@mike-scott I also shared a optee-tree branch for the meta-subscriber-override (they need to go hand in hand)

meta-lmp-base/recipes-security/optee/optee-os_git.bb
```diff
@@ -50,7 +34,7 @@ EXTRA_OEMAKE = "PLATFORM=${OPTEEMACHINE} O=out/arm \
                 CROSS_COMPILE_core=${HOST_PREFIX} \
                 CFG_WERROR=y DEBUG=0 LDFLAGS= \
                 LIBGCC_LOCATE_CFLAGS=--sysroot=${STAGING_DIR_HOST} \
-                CFG_TEE_CORE_LOG_LEVEL=2 CFG_TEE_TA_LOG_LEVEL=2 \
+                CFG_TEE_CORE_LOG_LEVEL=2 CFG_TEE_TA_LOG_LEVEL=2 CFG_CAAM_DBG=0x001 \
```
mike-scott: This CFG_CAAM_DBG=0x001 setting is new.  Is that to limit the debug output or turn it on?

ldts: to limit: is a bitmask (12 bits IIRC)  we could set it up to 0 but is good to remember we have this tweak IMO

mike-scott: Ok

meta-lmp-base/recipes-security/optee/optee-os_git.bb
```diff
@@ -1,32 +1,16 @@
 SUMMARY = "OP-TEE Trusted OS"
+HOMEPAGE = "https://github.com/foundries.io/optee_os.git"
```
mike-scott: Looks like there's extra white-space on the end of this line.

ricardosalveti: Address is also incorrect, should be "https://github.com/foundriesio/optee_os"

meta-lmp-base/recipes-security/optee/optee-os_git.bb
ricardosalveti: Can we move this to an append specifically for imx-based devices?

Like 'EXTRA_OEMAKE_append_imx = " etc".

ricardosalveti:
Just some general comments when looking at the patch-set available at https://github.com/foundriesio/optee_os/commits/3.6.0%2Bfio:

- https://github.com/foundriesio/optee_os/commit/da88f66e13b9f4c5ed3abd17708afe318894521d
  Do we really need this patch to be included? I had the impression the new patch-set you added based on the upstream CAAM + HUK support was the logic we wanted/needed. If so we are fine to remove our previous set, which includes this commit and https://github.com/foundriesio/optee_os/commit/e85af345e55ab5ac35f087ff6eb93592a4b89c7c
- https://github.com/foundriesio/optee_os/commit/3c9be191299ca32db67fd25dcbaffd0fe2a2012a
  This was previously applied only for qemuarm, so I would prefer not having it applied here as it will affect every other target we have. I'm also testing without this patch on qemu to make sure if this is indeed required.
- https://github.com/foundriesio/optee_os/commit/b44952d9d8e0c1945adbceffa6012b6aa3a646b0
  Would be nice to also enable this for the imx6-based devices, but we would also need to test on apalis (that way we avoid having multiple implementations in place, and just use the one that is closer to upstream, which will help comparing when we migrate to 3.8).
- https://github.com/foundriesio/optee_os/commit/6217ad5f2362216b0da86f051a09afebb56e9de9
  Missing signed-off line

Feel free to do a force push to https://github.com/foundriesio/optee_os/commits/3.6.0%2Bfio as you update with the required changes.

ricardosalveti:
Looking a bit more at https://github.com/foundriesio/optee_os/commit/3c9be191299ca32db67fd25dcbaffd0fe2a2012a and can still make the kernel to crash without that patch reverted on qemuarm64.

It seems we have a conflict when defining the secure memory region. Qemu will always create a secram area and this extra patch makes op-tee to also define its own secure memory region under reserver memory, which seems to be what is causing the boot problem.

Without the revert:
```
=> fdt print /secram@e000000
secram@e000000 {
        secure-status = "okay";
        status = "disabled";
        reg = <0x00000000 0x0e000000 0x00000000 0x01000000>;
        device_type = "memory";
};
=> fdt print /reserved-memory
reserved-memory {
        ranges;
        #size-cells = <0x00000002>;
        #address-cells = <0x00000002>;
        optee_core@0xe100000 {
                no-map;
                reg = <0x00000000 0x0e100000 0x00000000 0x00f00000>;
        };
        optee_shm@0x42000000 {
                no-map;
                reg = <0x00000000 0x42000000 0x00000000 0x00200000>;
        };
};
```

And when testing from u-boot I can get it to boot by removing /secram@e000000 or /reserved-memory/optee_core@0xe100000, so that might be a way forward (by adding fdt rm as part of the u-boot bootscript, we just need to find what is the correct/right solution here).

meta-lmp-base/recipes-security/optee/optee-os_git.bb
ldts: @ricardosalveti yes, you are right. ok

meta-lmp-base/recipes-security/optee/optee-os_git.bb
ldts: yes

ldts:
re: foundriesio/optee_os@da88f66 
right, we do not need this patch anymore  for imx7ulp; but I wasnt sure if some other target would still be using it (hence why I kept it).

re foundriesio/optee_os@3c9be19
if qemu crashes without the patch reverted I guess it means we dont need it. should we then just remove it and investigate the qemu crash?

re: foundriesio/optee_os@b44952d
sure, but I need to setup apalis first. can we not enable this once things are merged? then we can remove foundriesio/optee_os@da88f66 and enable Apalis.

ricardosalveti:
> re: [foundriesio/optee_os@da88f66](https://github.com/foundriesio/optee_os/commit/da88f66)
> right, we do not need this patch anymore for imx7ulp; but I wasnt sure if some other target would still be using it (hence why I kept it).

I think it is fine to remove it, let's have a clean base and we can add support for other targets based on the supported path as we move forward.

> re [foundriesio/optee_os@3c9be19](https://github.com/foundriesio/optee_os/commit/3c9be19)
> if qemu crashes without the patch reverted I guess it means we dont need it. should we then just remove it and investigate the qemu crash?

Just remove it for now, I'm going to propose a work around at the u-boot boot script to remove the fdt entry that causes the crash, less things to maintain.

> re: [foundriesio/optee_os@b44952d](https://github.com/foundriesio/optee_os/commit/b44952d)
> sure, but I need to setup apalis first. can we not enable this once things are merged? then we can remove [foundriesio/optee_os@da88f66](https://github.com/foundriesio/optee_os/commit/da88f66) and enable Apalis.

As we still need to do some work on Apalis, let's just make a clean rebase that only really supports ulp for now, and we can enable apalis properly as we move to 3.8.

ldts:
I updated the commit. please could you review?

meta-lmp-base/recipes-security/optee/optee-os_git.bb
ldts: done

PR #143, opened 2020-02-25
base: introduce docker-resource-control package

mike-scott:
Aktualizr-lite uses docker.service to pull down docker-app updates.
Without restrictions, this can cause CPU usage to peak during this
time.

Let's introduce a package: docker-resource-control to place limits
on the docker.service.  User's can configure the actual values.

NOTE: This does not affect containers only the docker.service.

### DIFF
--- a/meta-lmp-base/recipes-support/docker-resource-control/docker-resource-control/10-resource-control.conf.in
+++ b/meta-lmp-base/recipes-support/docker-resource-control/docker-resource-control/10-resource-control.conf.in
@@ -1,0 +1,3 @@
+[Service]
+CPUAccounting=true
+CPUQuota=@@CPU_QUOTA@@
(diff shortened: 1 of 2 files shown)


mike-scott:
NOTE: Once approved I will also add the following block to conf/local.conf (under the aktualizr resource settings) in the manifest:
```
# Set resource limits for docker service
# Uncomment these lines to change the default parameters.
# NOTES: This percent value is divided by the total # of cores
#        For 4 cores with 50% max per core this value would be "200%".
#        Requires CONFIG_CFS_BANDWIDTH=y in kernel defconfig.
#        This value does not affect running containers.
#RESOURCE_CPU_QUOTA_pn-docker-resource-control = "20%"
#IMAGE_INSTALL_append = " docker-resource-control"
```

mike-scott:
This can easily be tested on a live system (in my case a RaspberryPi 3 w/ 4 cores).
Let's set the quota to around 15% max on a 4 core system and then do a docker run on a stress test image which will run for 30 seconds:
```
$ sudo -s
# echo "60000" > /sys/fs/cgroup/cpu,cpuacct/system.slice/docker.service/cpu.cfs_quota_us
# exit
$ docker run --rm -it progrium/stress --cpu 4 --io 1 --vm 2 --vm-bytes 128M --timeout 30s
```

You should see:
- the docker pull portion runs fairly slow but steady (limited to 15% per core on a 4 core system)
- the docker run will jump to 100%

mike-scott:
archive this for now

PR #145, opened 2020-02-26
bsp: mfgtool-files: change extra depends to soc level

ricardosalveti:
imx-boot is a generic dependency for mx8-based targets, so change the
extra machine dependencies to be soc-based and not machine, in order to
cover a wider range of targets.


mike-scott:
@ricardosalveti This change works for me when tested.

PR #153, opened 2020-03-09
Base image cleanups

ricardosalveti:
* Drop ostree tarball, tar.xz and ext4.gz (not really useful and slows down the build process)
* Update default profile / .bashrc to include the extended PATH and improved PS1
* Extend busybox ash and less configs for a more rich experience


meta-lmp-base/recipes-samples/images/lmp-gateway-image.bb
```diff
@@ -10,7 +10,6 @@ require lmp-feature-nat64.inc
 require lmp-feature-jobserv.inc
 require lmp-feature-debug.inc
 require lmp-feature-softhsm.inc
-require lmp-feature-sbin-path-helper.inc
```
mike-scott: Do you also want to delete the `lmp-feature-sbin-path-helper.inc` file here as well?

mike-scott:
@rsalveti Looks good, just the 1 question

meta-lmp-base/recipes-samples/images/lmp-gateway-image.bb
ricardosalveti: Didn't delete just yet as we might have factory users importing such .inc (it won't cause any bad side effect, it just won't be necessary anymore).

Was thinking on keeping it for a few more releases and then we can safely remove it, what do you think?

mike-scott: Good point.

PR #156, opened 2020-03-15
recipes-containers: docker-apps: move to 0.6.0

mike-scott:
- rollback to 0.6.0 to avoid pulling down a cnab container which is used to
  render the docker-compose.yml
- the cnab and invoc containers gets pruned by aktualizr-lite at the end of
  every update meaning they will be re-downloaded on the next update
- this complexity isn't really needed as docker-app project seems to have
  stalled out and may never realize the correct abstractions to make all of
  this worth it.
- simplify, simplify, simplify


mike-scott:
> I'm not sure if the unused parameters patch will apply cleanly, but it would be nice to keep that functionality around.

The `matchAndMergeParametersDefinition` function isn't implemented in docker-app for 0.6.0.  Actually, the `internal/commands/parameters.go` file doesn't exist either.  Was there somewhere else I can apply it?

mike-scott:
Fixed an issue with the .bb file where I was still referencing the 2 patches.

doanac:
> The matchAndMergeParametersDefinition function isn't implemented in docker-app for 0.6.0. Actually, the internal/commands/parameters.go file doesn't exist either. Was there somewhere else I can apply it?

I just tested v0.6.0 - it doesn't fail on unused parameters. So we are good to go. This was "fixed" in later versions of docker app I guess

meta-lmp-base/recipes-containers/docker-app/docker-app_git.bb
```diff
@@ -5,26 +5,25 @@ LICENSE = "Apache-2.0"
...
+PV = "v0.6.0"
 
 inherit go goarch
 
 do_compile() {
 	cd ${S}/src/${GO_IMPORT}
 	BUILD_TAG=`git describe --always --abbrev=10`
 	BUILD_COMMIT=`git rev-parse --short HEAD`
-	DOCKER_APP_LDFLAGS="-X ${GO_IMPORT}/internal.GitCommit=${BUILD_COMMIT} \
+	DOCKER_APP_LDFLAGS="-s -w -X ${GO_IMPORT}/internal.GitCommit=${BUILD_COMMIT} \
```
ricardosalveti: Why the change here to strip during build? OE already calls strip, which is why the insane skip wasn't required before.

mike-scott: I was getting an stripping error when I built this as a test

mike-scott: @ricardosalveti testing the new version of this PR locally

PR #161, opened 2020-04-04
Caam locked

ldts:


ldts:
there is no imx7ulp target - would have to be done in the meta-override per product. Perhaps initially we can enable it for all imx platforms?

meta-lmp-base/recipes-security/optee/optee-os_git.bb
```diff
@@ -9,7 +9,7 @@ DEPENDS = "python3-pycrypto-native python3-pyelftools-native"
 SRC_URI = "git://github.com/foundriesio/optee_os.git;branch=${SRCBRANCH}"
 
 PV = "3.6.0+git"
-SRCREV = "a667673ba5ba2cf8740764a1c50d4ccf69ef8403"
+SRCREV = "a401165c3b3e1c5000b13970e98997d56c3ed281"
```
mike-scott, line 12: Can you update with cef3803949f821fe967e2bce32980f1316152c99 here? (Not that optee_os PR is merged)

meta-lmp-base/recipes-security/optee/optee-os-mfgtool_git.bb
```diff
@@ -1,3 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/optee-os:"
 
 require optee-os_${PV}.bb
+
+EXTRA_OEMAKE_append_imx = " \
```
ricardosalveti, line 5: Can you add this change to meta-lmp-bsp/recipes-security/optee/optee-os-mfgtool_%.bbappend instead?

That way the base recipe keeps without any bsp specific information, for users consuming our layer.

meta-lmp-base/recipes-security/optee/optee-os_git.bb
ricardosalveti, line 12: Rev is different now after we rebase & merged your change, so needs update as well.

PR #164, opened 2020-04-29
base: recipes-support: fioconfig: add toml helper support

mike-scott:
ficonfig has been updated to include an aktualizr-toml-update
helper script.  By using fioctl devices config updates command,
we can now easily set the docker-apps and tags settings for
aktualizr-lite.

### DIFF
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
@@ -7,6 +7,7 @@ LIC_FILES_CHKSUM = "file://src/${GO_IMPORT}/LICENSE;md5=2a944942e1496af1886903d2
 GO_IMPORT = "github.com/foundriesio/fioconfig"
 SRC_URI = "git://${GO_IMPORT} \
 	file://fioconfig.service \
+	file://fioconfig.path \
 	file://fioconfig-extract.service \
 "
 SRCREV = "1db6a4fdee49a69a64c4c0ac6f92bc2aa84c874f"
@@ -25,12 +26,13 @@ do_compile() {
 }
 
 SYSTEMD_PACKAGES += "${PN}"
-SYSTEMD_SERVICE_${PN} = "fioconfig.service fioconfig-extract.service"
+SYSTEMD_SERVICE_${PN} = "fioconfig.service fioconfig-extract.service fioconfig.path"
 
 
 do_install_append() {
 	install -d ${D}${systemd_system_unitdir}
 	install -m 0644 ${WORKDIR}/fioconfig.service ${D}${systemd_system_unitdir}/
+	install -m 0644 ${WORKDIR}/fioconfig.path ${D}${systemd_system_unitdir}/
 	install -m 0644 ${WORKDIR}/fioconfig-extract.service ${D}${systemd_system_unitdir}/
 	install -d ${D}${datadir}/fioconfig/handlers
 	install -m 0755 ${S}/src/${GO_IMPORT}/contrib/aktualizr-toml-update ${D}${datadir}/fioconfig/handlers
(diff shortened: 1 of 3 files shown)


meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
```diff
@@ -9,7 +9,7 @@ SRC_URI = "git://${GO_IMPORT} \
 	file://fioconfig.service \
 	file://fioconfig-extract.service \
 "
-SRCREV = "6d9f25f1c402eaa23bb98c0ce04355cd96e25647"
+SRCREV = "14512a50207a882522dcaa2e0c116f5a4da264f1"
```
doanac, line 12: I just added another commit to master you may as well pull in:

 https://github.com/foundriesio/fioconfig/commit/1db6a4fdee49a69a64c4c0ac6f92bc2aa84c874f

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
```diff
@@ -31,6 +31,8 @@ do_install_append() {
 	install -d ${D}${systemd_system_unitdir}
 	install -m 0644 ${WORKDIR}/fioconfig.service ${D}${systemd_system_unitdir}/
 	install -m 0644 ${WORKDIR}/fioconfig-extract.service ${D}${systemd_system_unitdir}/
+	install -d ${D}${bindir}
+	install -m 0755 ${S}/src/${GO_IMPORT}/contrib/aktualizr-toml-update ${D}${bindir}
```
doanac, line 35: this is shear curiosity - my version of this same change had 0655 for the file mode. I didn't find any clear guidance on what you should mark binaries with?

ricardosalveti: As this is a script it should have +x by default, otherwise I'm surprised it worked.

ricardosalveti: I believe it is probably better to have this script available via /usr/share or /usr/lib instead, in a folder that is specific to fioconfig, otherwise it could later create conflicts with other binaries, besides the confusion on the script scope.

mike-scott: @doanac 
A value of `6` for `user` (vs. `5` for `group` and `other`) would be: `read + write`
Where a value of `7` for user would be: `read + write + execute`

You could in theory leave this as 0644 (u=read+write, go=read) and make sure to call it with:
`/bin/sh /usr/bin/aktualizr-toml-update`
But I don't think we want that as some handlers may not be shell scripts.

ricardosalveti: It is fine for the script to have +x for every user.

mike-scott:
Changes:
- moved helper to /usr/share/fioconfig/handlers
- bumped fioconfig to include last commit from @doanac 
- added a path monitor service to make sure fioconfig service is started once /var/sota/sota.toml exists

doanac:
I've updated fioctl now as well: https://github.com/foundriesio/fioctl/commit/97f9a683985ee4638bbc770e3456998a6375448a

mike-scott:
Change:
- changed the file addition to not be so specific:
```
FILES_${PN} += " \
	${systemd_unitdir}/system/fioconfig.service \
	${systemd_unitdir}/system/fioconfig.path \
	${systemd_unitdir}/system/fioconfig-extract.service \
-	${datadir}/fioconfig/handlers/aktualizr-toml-update \
+	${datadir}/fioconfig/handlers \
"
```

mike-scott:
Rebased on current master

PR #165, opened 2020-04-29
base: lmp-auto-hostname: support model and serial via dmi

ricardosalveti:
Script was only checking for device-tree when looking for model and
serial, so extend it to also check for sysfs files available on systems
with dmi support.


ricardosalveti:
Tested on my intel nuc: intel-corei7-64-G6BE92900N85

ricardosalveti:
On my minnow: intel-corei7-64-8A20BC382

PR #168, opened 2020-05-04
Improve default features settings in LMP

ricardosalveti:
* Remove support for nfs (nfs client, not used/required by us)
* Manually added a default set for distro features based on was already available by OE. This will allow our user to add features that were previously removed via "_remove", such as x11 and wayland.


doanac:
LGTM - but you know my OE skillz :)

PR #171, opened 2020-05-07
base: kernel-lmp-fitimage: generate config even without dtb

liuming50:
Generate a default fit configuration even when no dtb is provided by the
kernel/board, as it gets used to verify all the other components (hash
and signature).


liuming50:
Without this commit, how could the fitimage boot on qemu machines?

PR #174, opened 2020-05-18
docker-compose: Add support for compose apps

doanac:
This adds a small out-of-tree patch to docker-compose that allows us
to "download" and Compose App. Some context about apps can be found
here:

 https://foundries.io/insights/2020/04/17/container-orchestration-v2/

We decided that "Bundles" was a bit confusing in the product since we
were using "Apps" everywhere. So its now called "compose apps" instead
of "compose bundles".


doanac:
all - not sure if we want to pull this into the LMP yet. As you can see, its isolated so it *should* be safe. We can carry it our own factories a little longer if you prefer though. This patch is needed so that we can use this:

 https://github.com/foundriesio/aktualizr-lite/pull/8

doanac:
gonna close this for now. We think doing this "download" logic in aklite is going to make more sense.

PR #183, opened 2020-06-16
Extending support for the lmp-base distribution (no ostree/sota)

ricardosalveti:
This series extends support for the lmp-base distribution, allowing machines to be built using DISTRO=lmp-base, which results in a yocto image that is based on lmp, but without ostree/sota by default.

The main idea for this type of builds is to allow an easier development flow when performing hardware bringup. Once the build is known to be functional, the user can simply just change the DISTRO variable to lmp and build the traditional image instead.


ricardosalveti:
https://ci.foundries.io/projects/lmp/builds/963/ has these changes in as a test.

ricardosalveti:
Thanks for the review!

PR #184, opened 2020-06-17
Add support for running e2fsck at the initrd

ricardosalveti:
This is basically 2 minor changes:
 * Add e2fsck to the initrd (increases ~0.5mb)
 * Run e2fsck before mounting the rootfs

Systemd also provides a job for running fsck on the rootfs, but e2fsck doesn't work correctly on mounted filesystems, which is why it is better and safer to just run at the initrd level instead.

### DIFF
--- a/meta-lmp-base/recipes-core/images/initramfs-ostree-lmp-image.bb
+++ b/meta-lmp-base/recipes-core/images/initramfs-ostree-lmp-image.bb
@@ -1,6 +1,12 @@
 DESCRIPTION = "Linux microPlatform OSTree initramfs image"
 
-PACKAGE_INSTALL = "initramfs-framework-base initramfs-module-udev initramfs-module-rootfs initramfs-module-ostree ${VIRTUAL-RUNTIME_base-utils} udev base-passwd ${ROOTFS_BOOTSTRAP_INSTALL}"
+PACKAGE_INSTALL = "initramfs-framework-base \
+	initramfs-module-udev \
+	initramfs-module-rootfs \
+	initramfs-module-ostree \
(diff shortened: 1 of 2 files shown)


ricardosalveti:
v2:
 - using /dev/kmsg instead of console as it can then be consumed by journalctl
 - also print retcode in case of errors

```
diff --git a/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/rootfs b/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/rootfs
index f326ad4..ee6ab24 100644
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/rootfs
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/rootfs
@@ -8,12 +8,15 @@ rootfs_enabled() {

 e2fsck_check() {
        if [ -n "`which e2fsck`" ]; then
-               e2fsck -p -v ${1} >/dev/console
-               # Return code higher than 4 means uncorrected / operational error
+               fsckout=`e2fsck -p -v ${1}`
+               fsckret=$?
+               # Avoid empty newline after summary
+               echo "e2fsck: ${fsckout}" >/dev/kmsg
+               # Return code >= 4 means uncorrected / operational error
                ## TODO: force boot into a recovery mode or similar, as there is really not
                ## much we can do in case the fs is corrupted in a bad way
-               if [ "$?" -ge "4" ]; then
-                       msg "WARNING: file system errors left uncorrected"
+               if [ "${fsckret}" -ge "4" ]; then
+                       echo "e2fsck: WARNING: file system errors left uncorrected: ret ${fsckret}" >/dev/kmsg
                fi
        fi
 }
```

PR #191, opened 2020-06-25
base: lmp-image-common: only increase overhead factor for QEMU targets

mike-scott:
As a part of the LmP common image changes, we run a resize-helper which
expands the rootfs to fill the available space.  This usually takes care
of needing enough space on the rootfs to download a complete OTA image.
(about double the size at a minimum)

QEMU target images are not expanded past their original size during this
process and so we increased the overhead factor for all targets to 2.3.

This has 2 downsides:
1. Subscribers have a hard time overriding this setting
2. non-QEMU target wic images are needlessly larger than they should be
   due to the resize-helper.

By selectively setting an overridable value of 2.3 for QEMU targets,
we solve both of these issues.

### DIFF
--- a/meta-lmp-base/recipes-samples/images/lmp-image-common.inc
+++ b/meta-lmp-base/recipes-samples/images/lmp-image-common.inc
@@ -13,8 +13,13 @@ SRC_URI = "\
     file://sudoers \
 "
 
-# Enough free space for a full image update
-IMAGE_OVERHEAD_FACTOR = "2.3"
+# In most cases the rootfs will be expanded to fit available storage
+# However, for QEMU targets we need to pre-allocate enough free space
+# for a full image update
+IMAGE_OVERHEAD_FACTOR_qemuriscv64 ?= "2.3"
+IMAGE_OVERHEAD_FACTOR_qemuarm64 ?= "2.3"
+IMAGE_OVERHEAD_FACTOR_qemuarm ?= "2.3"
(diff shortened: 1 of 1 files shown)


meta-lmp-base/recipes-samples/images/lmp-image-common.inc
```diff
@@ -13,8 +13,13 @@ SRC_URI = "\
...
 
-# Enough free space for a full image update
-IMAGE_OVERHEAD_FACTOR = "2.3"
+# In most cases the rootfs will be expanded to fit available storage
+# However, for QEMU targets we need to pre-allocate enough free space
+# for a full image update
+IMAGE_OVERHEAD_FACTOR_qemuriscv64 ?= "2.3"
+IMAGE_OVERHEAD_FACTOR_qemuarm64 ?= "2.3"
+IMAGE_OVERHEAD_FACTOR_qemuarm ?= "2.3"
+IMAGE_OVERHEAD_FACTOR_intel-corei7-64 ?= "2.3"
```
EmbeddedAndroid, line 22: I understand the QEMU changes here, but what is the rationale behind this target? It should expand properly like our other hardware based MACHINES?

mike-scott: We spoke on Slack but for the purposes of documenting:
This is here because we can use the intel-corei7-64 WIC image with QEMU x86 and in that use-case you still need the larger space (similar to the other QEMU targets).

PR #192, opened 2020-06-26
Configurable docker image download

mike-sul:


mike-scott:
Reviewing this now.  I love that you set this up.  Thanks!

meta-lmp-base/recipes-containers/docker/docker-ce_%.bbappend
```diff
@@ -3,9 +3,19 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 SRC_URI_append = " \
```
mike-scott, line 3: Can we add a comment above here with the tracking location of the docker patch for max download retries?

mike-sul: done

PR #199, opened 2020-07-29
base: kernel-lmp-fitimage: add support for FIT_LOADABLES

mike-scott:
Extend this class to handle FIT_LOADABLES as a list of deployed files
to be added to the kernel fitimage.  This allows subscribers to
retreive these files in u-boot where the contents are verified prior
to using.

Example would be to add the following line to machine.conf:
FIT_LOADABLES = "core0_m4_image.bin core1_m4_image.bin"

- A recipe would place those files in ${DEPLOYDIR} in do_deploy
- Then an additional do_compile[depends] is added to the kernel
  recipe such as:
  do_compile[depends] += "<recipe>:do_deploy"

### DIFF
--- a/meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
+++ b/meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
@@ -4,6 +4,7 @@ inherit kernel-fitimage
 
 # Default value for deployment filenames
 FPGA_BINARY ?= "fpga.bin"
+FIT_LOADABLES ?= ""
 
 #
 # Emit the fitImage ITS configuration section
@@ -14,7 +15,8 @@ FPGA_BINARY ?= "fpga.bin"
 # $4 ... ramdisk ID
 # $5 ... setup ID
 # $6 ... fpga ID
-# $7 ... default flag
+# $7 ... loadable ID
+# $8 ... default flag
 fitimage_emit_section_config() {
 
 	conf_csum="${FIT_HASH_ALG}"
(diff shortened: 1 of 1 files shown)


meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
```diff
@@ -259,14 +308,24 @@ fitimage_assemble() {
...
+				fitimage_emit_section_config ${1} "${DTB}" "${kernelcount}" "${DTB}" "${ramdiskcount}" "${setupcount}" "${fpgacount}" "" "`expr ${i} = ${dtbcount}`"
 			fi
 			i=`expr ${i} + 1`
 		done
 	else
-		fitimage_emit_section_config ${1} "${kernelcount}" "" "${ramdiskcount}" "${setupcount}" "${fpgacount}" ""
+		fitimage_emit_section_config ${1} "" "${kernelcount}" "" "${ramdiskcount}" "${setupcount}" "${fpgacount}" "" ""
+	fi
+
+	i=0
```
mike-scott, line 321: Just realized, I don't need `i` anymore.  I changed this last night.

mike-scott:
Split this into 2 patches showing the adjustments to this code in a better way

mike-scott:
Adjusted per discussion with @ricardosalveti  (back to 1 commit)

mike-scott:
NOTE: This needs to be re-tested: DO NOT MERGE

mike-scott:
@ricardosalveti This is ready for review.  I've tested it.

PR #200, opened 2020-07-30
base: docker-ce: Untar images before running docker

mike-sul:
Systemd service that extracts container images from a tarball into /var/lib/docker so they become 'preloaded'
before running dockerd service.
The tarball is generated and injected into a system image/rootfs (WIC) either by lmp or [container build](https://ci.foundries.io/projects/msul-dev01/lmp/builds/306/assemble-system-image/) or by fioctl (TBD).
This should go with https://github.com/foundriesio/ci-scripts/pull/73 and https://git.foundries.io/development/cloudplatforms/infrastructure/repoz/merge_requests/15

### DIFF
--- a/meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
+++ b/meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
@@ -1,0 +1,17 @@
+[Unit]
+Description=Docker container images preloading
+After=local-fs.target
+Before=docker.service
+ConditionPathExists=/etc/systemd/system/multi-user.target.wants/aktualizr-lite.service
+ConditionPathExists=/var/sota/import/installed_versions
+ConditionPathExistsGlob=/var/sota/import/*.tar
+
+[Service]
+Type=oneshot
+ExecStartPre=mkdir -p /var/lib/docker
+# pre-load container images into the docker data root directory (only works for overlay2 graph driver)
+ExecStart=/bin/sh -c 'for f in `find /var/sota/import -name "*.tar"`; do echo "Compose App: preloading $f" && tar -xf $f -C /var/lib/docker/ --warning=no-timestamp && rm $f; done'
+RemainAfterExit=yes
+
+[Install]
+WantedBy=docker.service
--- a/meta-lmp-base/recipes-samples/images/lmp-feature-docker.inc
+++ b/meta-lmp-base/recipes-samples/images/lmp-feature-docker.inc
@@ -5,6 +5,7 @@ CORE_IMAGE_BASE_INSTALL += " \
     docker-credential-helper-fio \
     python3-docker \
     python3-docker-compose \
+    docker-image-preload \
 "
 
 EXTRA_USERS_PARAMS += "\
--- a/meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload_0.1.bb
+++ b/meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload_0.1.bb
@@ -1,0 +1,19 @@
+SUMMARY = "Systemd service for preloading of container images"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit allarch systemd
+
+SRC_URI = "file://docker-image-preload.service"
+
+PACKAGE_ARCH = "${MACHINE_ARCH}"
+
+SYSTEMD_SERVICE_${PN} = "docker-image-preload.service"
+SYSTEMD_AUTO_ENABLE_${PN} = "enable"
+
+do_install() {
+  install -d ${D}${systemd_system_unitdir}
+  install -m 0644 ${WORKDIR}/docker-image-preload.service ${D}${systemd_system_unitdir}
+}
+
+FILES_${PN} += "${systemd_system_unitdir}/docker-image-preload.service"


meta-lmp-base/recipes-containers/docker/docker-ce/docker.service
```diff
@@ -34,6 +34,9 @@ StartLimitInterval=60s
 # after three consecutive failures.
 FailureAction=reboot
 
+# pre-load container images into the docker data root directory (only works for overlay2 graph driver)
+ExecStartPre=-/bin/sh -c 'mkdir -p /var/lib/docker && for f in `find /var/sota/import -name "*.tar"`; do echo "compose app: preloading $f" && tar -xf $f -C /var/lib/docker/ && rm $f; done'
+
```
doanac, line 39: clever - so it only runs once, unless you add to that folder on subsequent boots? I'm guessing you may have thoughts on (ab)using this later on?

ricardosalveti: I was wondering if we should just create a separated service file just for importing the containers, as that way we can probably control the process better and avoid changing the main docker service file. That way the user can add or remove the logic by deciding what to include at the lmp image.

If we create another service we can make it run before docker and only when there are known tar files at var/sota/import (e.g. `ConditionPathExistsGlob=/var/sota/import/*.tar`).

mike-sul: > That way the user can add or remove the logic by deciding what to include at the lmp image.

This is actually for the container build. An assumption is that a customer makes lmp builds rarely and normally works with container build and if at some point he/she decides to get a system image pre-loaded with his/her latest apps' container images then these ci-scripts params should be turned on https://source.foundries.io/factories/msul-dev01/ci-scripts.git/tree/factory-config.yml#n38. Moreover, @doanac thinks that we should add an additional command to `fioctl` that would trigger a job (container build run) that produces a system image with pre-loaded container images taking Target as an input. Such 'interface/API' assumes that each Target's lmp build/image should include the given logic unconditionally, so users could assemble a "preloaded" image for any Target they choose.

Having said that, I think it's not a big deal to add a new a systemd service. I agree that it would "probably control the process better and avoid changing the main docker service file". I just think that we might not need it at all in the future once we move to ostree.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Docker container images preloading
+
```
ricardosalveti, line 3: Please remove this empty line.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
```diff
@@ -0,0 +1,17 @@
...
+After=local-fs.target
+Before=docker.service
+
```
ricardosalveti, line 6: Please remove this empty line.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
```diff
@@ -0,0 +1,17 @@
...
+ConditionPathExists=/var/sota/import/installed_versions
+ConditionPathExistsGlob=/var/sota/import/*.tar
+
+[Service]
+Type=oneshot
```
ricardosalveti, line 11: I believe it is also recommended to use 'RemainAfterExit=yes' here, otherwise the service will show up as dead after being executed. This is not particularly clear to, but it seems the standard.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
```diff
@@ -0,0 +1,17 @@
...
+
+# pre-load container images into the docker data root directory (only works for overlay2 graph driver)
+ExecStart=/bin/sh -c 'mkdir -p /var/lib/docker && for f in `find /var/sota/import -name "*.tar"`; do echo "Compose App: preloading $f" && tar -xf $f -C /var/lib/docker/ && rm $f; done'
```
ricardosalveti, line 14: You can move 'mkdir -p /var/lib/docker' to ExecStartPre.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Docker container images preloading
+
+After=local-fs.target
+Before=docker.service
+
+ConditionPathExists=/var/sota/import/installed_versions
```
ricardosalveti, line 7: IIRC this file is removed by aktualizr once it starts for the first time, so probably also good to add a dependency on aktualizr-lite (Before).

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
mike-sul: Why? I put it there deliberately, IMHO  it makes 'code' more readable.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
mike-sul: I read about it and tried it, but I don't see any reason it's needed in this particular case.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
mike-sul: +

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
mike-sul: It doesn't bring any value.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
mike-sul: I just double-checked it once more, I don't see any difference in the service behavior with or without  RemainAfterExit=yes.

mike-sul: Actually, there is some difference. With `RemainAfterExit=yes` and if the conditions are met and the ExecStart* scripts return(s) 0 then the service state will be at `Active: active (exited)` state while without `RemainAfterExit=yes` or if the conditions are not met or one of ExecStart* scripts fail then it will be at `Active: inactive (dead)`. Also, the service will be at this state `Active: inactive (dead)` after the second boot regardless of `RemainAfterExit=yes`.
If the service is at `Active: inactive (dead)` then the docker service restart causes a restart of the given service. It's not quite what we need but it doesn't cause any issue anyway.

mike-sul:
@rsalveti Updated according to your comments and verified on qemu and rpi4 under different conditions, with available an image tarball, without it, behavior after rebooting, etc.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
ricardosalveti: Cool, thanks for investigating that one.

meta-lmp-base/recipes-support/docker-image-preload/docker-image-preload/docker-image-preload.service
ricardosalveti: Just so we can separate the blocks via an empty line.

PR #204, opened 2020-08-13
Adding linux-lmp-lts and bumping linux-lmp to 5.8.1

ricardosalveti:
Tested 5.8.1 kernel on qemuarm64, intel-corei7-64, beaglebone-yocto, qemuriscv64, freedom-u540, qemuarm and apalis-imx6.

Kernel meta was also updated with the latest config related changes.

### DIFF
--- a/meta-lmp-base/recipes-kernel/wireguard/wireguard-module_1.0.20200712.bb
+++ b/meta-lmp-base/recipes-kernel/wireguard/wireguard-module_1.0.20200712.bb
@@ -28,6 +28,14 @@ MODULE_NAME = "wireguard"
 # with kernel-module-.
 PKG_${PN} = "kernel-module-${MODULE_NAME}"
 
+# Only set default rprovides if kernel is different than linux-lmp, assuming
+# it is older than 5.6 (version in which the module is provided by the kernel)
+python __anonymous() {
+    pn = d.getVar("PN")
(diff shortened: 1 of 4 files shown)


meta-lmp-base/recipes-kernel/wireguard/wireguard-module_1.0.20200712.bb
```diff
@@ -28,6 +28,14 @@ MODULE_NAME = "wireguard"
 # with kernel-module-.
 PKG_${PN} = "kernel-module-${MODULE_NAME}"
 
+# Only set default rprovides if kernel is different than linux-lmp, assuming
+# it is older than 5.6 (version in which the module is provided by the kernel)
+python __anonymous() {
+    pn = d.getVar("PN")
+    if d.getVar("PREFERRED_PROVIDER_virtual/kernel") != "linux-lmp":
```
mike-scott, line 35: Is there any way to check the kernel version of the virtual/kernel instead of linux-lmp?

ricardosalveti: Unfortunately not from this recipe, as bitbake/OE only really fetches that level of information when executing steps on the linux-based recipe. At this point all we have is the preferred provider, and just because we have that in the machine configuration.

PR #207, opened 2020-08-22
Update apalis to reference platform level

ricardosalveti:
- add support for the spl/optee/u-boot flow
- switch to latest u-boot-fio tree, which is based on the latest nxp bsp
- add support for mfgtools

Changes mostly based from the work done by @mike-scott at some of our factories.


meta-lmp-bsp/conf/machine/include/lmp-mfgtool-machine-custom.inc
```diff
@@ -5,3 +5,17 @@ PREFERRED_PROVIDER_u-boot_mx6ull = "u-boot-fio-mfgtool"
 PREFERRED_PROVIDER_u-boot-default-script_mx6ull = "u-boot-ostree-scr-fit"
 IMAGE_BOOT_FILES_remove_mx6ull = "boot.scr uEnv.txt zImage"
 IMAGE_BOOT_FILES_append_mx6ull = " boot.itb SPL u-boot.itb"
+
+# Apalis-iMX6
+PREFERRED_PROVIDER_virtual/kernel_apalis-imx6 = "linux-lmp-dev-mfgtool"
+KERNEL_REPO_apalis-imx6 ?= "git://github.com/foundriesio/linux.git"
+LINUX_VERSION_apalis-imx6 ?= "5.4.y"
```
mike-scott, line 12: @ricardosalveti Sorry, this has been updated to 5.8.y in master.  We should probably do the same here.

mike-scott: Unless of course, we would rather tie apalis-imx6 to the linux-lmp-fslc-imx recipe?  In which case we probably need to make a bunch of *-mfgtool recipes for easy maintenance.

ricardosalveti: We are not yet using the kernel generated by mfgtools, which is why it wasn't much of a concern for me.

I was planning on switching to the imx recipe after this gets tested and merged, to avoid adding more variables to the game here (mostly due the possible cache related issues which might require a new optee-os as well).

PR #210, opened 2020-09-04
Switch iMX-based devices to latest NXP BSP

ricardosalveti:
* Switch remaining targets to u-boot-fio 2020.04
* imx8mmevk: switch to latest u-boot/optee/kernel
* apalis-imx8: import machine config and bump atf/u-boot to latest
* imx7ulpea-ucom: import machine config and bump u-boot to latest
* Drop compatibility with meta-toradex bsp layers


mike-scott (commented):
@ricardosalveti I haven't take the time to look through all of these commits.  But I'm inclined to get this merged into master so we can get real HW builds that can go hand in hand with review.

For some of this we need to see the builds.

PR #220, opened 2020-10-19
base: classes: kernel-fit-image: fix DTB iteration variable

mike-scott:
When we use variable "i" to count iteratation through DTBS in
fitimage_assemble.  However, during each loop, we also call
fitimage_emit_section_config which also uses a variable "i" to
count loops when processing LOADABLES.  This causes the value
of "i" in the fitimage_assemble loop to seemingly stay the same
or randomly change in odd ways.

Let's fix the fitimage_assemble iteration count by renaming the
variable to "n" which is not used elsewhere.


ricardosalveti (approved):
LGTM

PR #225, opened 2020-11-05
recipes-support: mfgtool-files: allow full_image flash filename override

mike-scott:
Let's add an OE var: MFGTOOL_FLASH_IMAGE to be used by the flash scripts for
building the .wic filename.

Default to: lmp-base-console-image

### DIFF
--- a/meta-lmp-base/recipes-support/mfgtool-files/mfgtool-files_0.1.bb
+++ b/meta-lmp-base/recipes-support/mfgtool-files/mfgtool-files_0.1.bb
@@ -9,6 +9,7 @@ INITRAMFS_IMAGE = "fsl-image-mfgtool-initramfs"
 DEPENDS = "${INITRAMFS_IMAGE}"
 
 UUU_RELEASE = "1.3.154"
+MFGTOOL_FLASH_IMAGE ?= "lmp-base-console-image"
 
 SRC_URI = " \
     https://github.com/NXPmicro/mfgtools/releases/download/uuu_${UUU_RELEASE}/uuu;name=Linux \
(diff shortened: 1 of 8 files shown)


meta-lmp-base/recipes-support/mfgtool-files/mfgtool-files_0.1.bb
```diff
@@ -9,6 +9,7 @@ INITRAMFS_IMAGE = "fsl-image-mfgtool-initramfs"
 DEPENDS = "${INITRAMFS_IMAGE}"
 
 UUU_RELEASE = "1.3.154"
+FULL_FLASH_IMAGE ?= "factory-image"
```
ricardosalveti, line 12: As we discussed on slack, I would prefer having this set to an image that is available in meta-lmp alone, such as lmp-base-console-image. We can change ci-scripts (bb-config) to set a default value for factory users, as we already do for image name.

Then it can be easily customized by our factory user as well (in case they decide to produce an image with a different name).

mike-scott: Agreed and done.

mike-scott:
Changes from v1:
- Rebased on today's master
- changed var name to: MFGTOOL_FLASH_IMAGE
- default value to: lmp-base-console-image
  NOTE: I removed the initial "lmp-" from the scripts so that this value doesn't need to follow any format

mike-scott:
@ricardosalveti Let me know if you see any other issues.

mike-scott:
Requires: https://github.com/foundriesio/ci-scripts/pull/100

meta-lmp-base/recipes-support/mfgtool-files/mfgtool-files_0.1.bb
```diff
@@ -28,7 +29,8 @@ PACKAGE_ARCH = "${MACHINE_ARCH}"
 
 do_compile() {
     sed -e 's/@@MACHINE@@/${MACHINE}/' ${S}/bootloader.uuu.in > bootloader.uuu
-    sed -e 's/@@MACHINE@@/${MACHINE}/' ${S}/full_image.uuu.in > full_image.uuu
+    sed -e 's/@@MACHINE@@/${MACHINE}/' ${S}/full_image.uuu.in > full_image.uuu.tmp
+    sed -e 's/@@MFGTOOL_FLASH_IMAGE@@/${MFGTOOL_FLASH_IMAGE}/' ${S}/full_image.uuu.tmp > full_image.uuu
```
ricardosalveti, line 33: No need for 2 sed lines here, you can add another command with another -e after the first, that way we don't need to generate a tmp file.

ricardosalveti (approved):
LGTM

PR #226, opened 2020-11-06
u-boot: drop u-boot.inc

liuming50:
The u-boot.inc file has been removed in meta-toradex-nxp dunfell then
we could remove it from here.

Now only the u-boot.inc file from OE would be used.

This fixes a issue that .cfg files were never applied when UBOOT_CONFIG
is set, which has been fixed in OE by commit 4fc38627:
[ u-boot: fix condition to allow use of *.cfg ]

Signed-off-by: Ming Liu <ming.liu@toradex.com>


ricardosalveti (approved):
LGTM, thanks!

PR #228, opened 2020-11-10
docker-credential-helper: Support variable server names

doanac:
This allows the script to use the configured tls.server to grab
credentials from. This will become important when customers have their
own device gateway urls.

### DIFF
--- a/meta-lmp-base/recipes-support/docker-credential-helper-fio/files/docker-credential-fio-helper
+++ b/meta-lmp-base/recipes-support/docker-credential-helper-fio/files/docker-credential-fio-helper
@@ -2,9 +2,22 @@
 
 # Use stderr for logging err output in libaktualizr
 export LOG_STDERR=1
+SOTA_DIR="${SOTA_DIR-/var/sota}"
 
 LOGLEVEL="${CREDS_LOGLEVEL-4}"
 
 if [ "$1" = "get" ] ; then
-	exec /usr/bin/aktualizr-get --loglevel $LOGLEVEL -u https://ota-lite.foundries.io:8443/hub-creds/
+	if [ $(id -u) != "0" ] ; then
+		echo "ERROR: $0 must be run as root to access $SOTA_DIR"
+		exit 1
+	fi
+	if [ ! -f ${SOTA_DIR}/sota.toml ] ; then
+		echo "ERROR: Device does not appear to be registered under $SOTA_DIR"
+		exit 1
+	fi
+	server=$(grep -m1 ^server ${SOTA_DIR}/sota.toml | cut -d\" -f2)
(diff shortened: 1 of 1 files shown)


meta-lmp-base/recipes-support/docker-credential-helper-fio/files/docker-credential-fio-helper
```diff
@@ -6,5 +6,9 @@ export LOG_STDERR=1
 LOGLEVEL="${CREDS_LOGLEVEL-4}"
 
 if [ "$1" = "get" ] ; then
-	exec /usr/bin/aktualizr-get --loglevel $LOGLEVEL -u https://ota-lite.foundries.io:8443/hub-creds/
+	server=$(cat /var/sota/sota.toml | grep ^server | head -n1 | cut -d\" -f2)
```
ricardosalveti, line 9: One problem is that this helper will be called with all users, and only root will have access to sota.toml, so that will let to a permission denied issue when called as fio (or any normal user).

So I would probably move the default server settings above this line (server="https://ota-lite.foundries.io:8443") and check if enough permission to parse sota.toml, and then update the variable if another server is available.

You can also drop cat and head by having one single grep call instead, such as:

`grep -m1 ^server /var/sota/sota.toml | cut -d\" -f2
`

doanac: i've been thinking the same thing. its really confusing right now. this would help usability.

doanac: force push fix.

meta-lmp-base/recipes-support/docker-credential-helper-fio/files/docker-credential-fio-helper
```diff
@@ -2,9 +2,22 @@
...
-	exec /usr/bin/aktualizr-get --loglevel $LOGLEVEL -u https://ota-lite.foundries.io:8443/hub-creds/
+	if [ $(id -u) != "0" ] ; then
+		echo "ERROR: $0 must be run as root to access $SOTA_DIR"
+		exit 1
+	fi
+	if [ ! -f ${SOTA_DIR}/sota.toml ] ; then
+		echo "ERROR: Device does not appear to be registered under $SOTA_DIR"
+		exit 1
+	fi
+	server=$(cat ${SOTA_DIR}/sota.toml | grep -m1 ^server | cut -d\" -f2)
```
ricardosalveti, line 18: no need for cat here, you can just grep the file directly

ricardosalveti (approved):
LGTM

PR #238, opened 2020-11-25
u-boot-fitImage: pass the machine name in the SPL fit

ldts:
SPL will then be able to string match properly during
board_fit_config_name_match (CONFIG_SPL_LOAD_FIT)


ldts:
once that is merged, we can fix the string match in some of the currently supported platforms SPL (ie, datasignals)

ricardosalveti (approved):
Change is fine, unclear how that would help SPL, but good to have the machine description anyway.

PR #241, opened 2020-11-30
WIP: base: docker: turn on 'live restore' feature

mike-sul:
The 'live restore' feature makes a docker daemon not to stop running
containers during its shutdown or restart. The given feature can be
useful for an ostree based image update, i.e. after an image update is
checked out, a docker daemon restart is required, to make it 'see' the
updated images and we don't want to stop/restart running containers,
images of which haven't been updated.


ricardosalveti:
Is there any implication to the normal system shutdown/restart process?

mike-sul:
> Is there any implication to the normal system shutdown/restart process?

Haven't noticed any so far, but it's WIP, I mean verification and testing. While this configuration doesn't stop running containers it actually 'turns off' port mapping during the daemon restart, not sure if it's exactly what we need.

PR #250, opened 2021-01-06
base: fioconfig: Add environment file to configure interval

munoz0raul:
fioconfig runs in a default interval of 300 seconds. This change will allow interval modifications by changing the environment file.

### DIFF
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig/fioconfig.service
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig/fioconfig.service
@@ -4,6 +4,7 @@ After=network.target
 ConditionPathExists=/var/sota/sota.toml
 
 [Service]
+EnvironmentFile=-/etc/default/fioconfig
 RestartSec=10
 Restart=always
 ExecStartPre=mkdir -p /var/run/secrets


doanac:
I think the cli arg parsing tool I use for fioconfig can read environment variables for options. We should probably do that to simplify this. I'll look tomorrow

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
```diff
@@ -25,6 +26,8 @@ SYSTEMD_SERVICE_${PN} = "fioconfig.service fioconfig-extract.service fioconfig.p
 
 do_install_append() {
 	install -d ${D}${systemd_system_unitdir}
+    install -d ${D}${sysconfdir}/default/
+    install -m 0644 ${WORKDIR}/fioconfig.conf ${D}${sysconfdir}/default/fioconfig
```
doanac, line 30: tabs vs spaces?

meta-lmp-base/recipes-support/fioconfig/fioconfig/fioconfig.service
```diff
@@ -4,10 +4,11 @@ After=network.target
 ConditionPathExists=/var/sota/sota.toml
 
 [Service]
+EnvironmentFile=/etc/default/fioconfig
 RestartSec=10
 Restart=always
 ExecStartPre=mkdir -p /var/run/secrets
-ExecStart=/usr/bin/fioconfig daemon
+ExecStart=/usr/bin/fioconfig daemon -i ${FIOCONFIG_INTERVAL}
```
doanac, line 11: You can just set the variable `DAEMON_INTERVAL` and not need this:

https://github.com/foundriesio/fioconfig/blob/481fe3ee2b595d7e1c3e10b98169ab5bb7befa46/main.go#L119

ricardosalveti: With DAEMON_INTERVAL you won't need to change ExecStart.

Also, don't ship with a default configuration, this should instead be provided by a factory only (for cases where the default value is not enough).

Change EnvironmentFile to 

`EnvironmentFile=-/etc/default/fioconfig
`

ricardosalveti (approved):
LGTM, thanks!

PR #255, opened 2021-01-25
meta-lmp: xilinx-tools: bitstream-extraction: Add bitstream support

Tim-Anderson:

### DIFF
--- a/meta-lmp-bsp/conf/machine/uz3eg-iocc.conf
+++ b/meta-lmp-bsp/conf/machine/uz3eg-iocc.conf
@@ -12,3 +12,7 @@ HDF_MACHINE ?= "zcu102-zynqmp"
 HDF_BASE = "file://"
 HDF_PATH = "system.dsa"
 HDF_EXT = "dsa"
+
+# bitstream support
+MACHINE_FIRMWARE_append = " bitstream-extraction fpga-manager-script"
+MACHINE_EXTRA_RRECOMMENDS += "${MACHINE_FIRMWARE}"
--- a/meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/bitstream/bitstream-extraction_git.bb
+++ b/meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/bitstream/bitstream-extraction_git.bb
@@ -1,0 +1,33 @@
+DESCRIPTION = "Recipe to extract bitstream"
+
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+DEPENDS += "virtual/hdf"
+
+PROVIDES = "virtual/bitstream"
+
+PACKAGE_ARCH ?= "${MACHINE_ARCH}"
+
+inherit xsctbit deploy
+
+XSCTH_MISC = "-hwpname ${XSCTH_PROJ}_hwproj -hdf_type ${HDF_EXT}"
+
+do_compile[noexec] = "1"
+
+SYSROOT_DIRS += "${nonarch_base_libdir}/firmware"
+
+do_install() {
+
+    if [ -e ${XSCTH_WS}/${XSCTH_PROJ}_hwproj/*.bit ]; then
+        install -d ${D}/${nonarch_base_libdir}/firmware/
+        ln -s /var/lib/firmware/bitstream ${D}/${nonarch_base_libdir}/firmware/bitstream
+        install -Dm 0644 ${XSCTH_WS}/${XSCTH_PROJ}_hwproj/*.bit ${D}/${nonarch_base_libdir}/firmware/
+    fi
+}
+
+do_deploy() {
+}
+addtask do_deploy after do_install
+
+FILES_${PN} += "${nonarch_base_libdir}/firmware/bitstream ${nonarch_base_libdir}/firmware/*.bit"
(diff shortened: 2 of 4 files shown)


meta-lmp-bsp/conf/machine/include/ultrazed-lmp.inc
```diff
@@ -55,3 +55,7 @@ PMU_FIRMWARE_DEPLOY_DIR ?= "${DEPLOY_DIR_IMAGE}"
 
 # Not building for QEMU standard Xinix build
 EXTRA_IMAGEDEPENDS_remove = "qemu-devicetrees"
+
+# bitstream support
+MACHINE_FIRMWARE_append_uz3eg-iocc = " bitstream-extraction fpga-manager-script"
```
ricardosalveti, line 60: Since this is machine specific, move this line to conf/machine/uz3eg-iocc.conf instead.

Also, please split the machine changes into a different commit.

meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/bitstream/bitstream-extraction_git.bb
```diff
@@ -0,0 +1,30 @@
...
+
+do_compile[noexec] = "1"
+
+SYSROOT_DIRS += "${nonarch_base_libdir}/firmware"
+
+do_install() {
+
+    if [ -e ${XSCTH_WS}/${XSCTH_PROJ}_hwproj/*.bit ]; then
+        install -d ${D}/${nonarch_base_libdir}/firmware/
+        install -d ${D}/var/lib/firmware/bitstream/
```
ricardosalveti, line 24: Var is a special directory in our setup because it is RW, so not really managed by ostree.

As all you need is a link, it is better to create a systemd tmpfile for this recipe, like https://github.com/foundriesio/meta-lmp/blob/master/meta-lmp-base/recipes-security/softhsm/softhsm/tmpfiles.conf (but using link instead, check https://www.freedesktop.org/software/systemd/man/tmpfiles.d.html for more info). This allows the link to be created by systemd during boot automatically, and it will keep working even if the user erases /var completely (as it will be restored on a following boot).

meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/bitstream/bitstream-extraction_git.bb
```diff
@@ -0,0 +1,30 @@
...
+        ln -s -r ${D}/var/lib/firmware/bitstream ${D}/${nonarch_base_libdir}/firmware/bitstream
+        install -Dm 0644 ${XSCTH_WS}/${XSCTH_PROJ}_hwproj/*.bit ${D}/${nonarch_base_libdir}/firmware/
+    fi
+}
+
```
ricardosalveti, line 29: For loading from fit you will also need a do_deploy (no needed for now, but just a heads up).

meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/bitstream/bitstream-extraction_git.bb
ricardosalveti: Actually, your recipe will have to install the link and the systemd tmpfiles configuration will have to create the directory (under var).

meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/fpga-manager-script/files/0001-Correct-used-paths.patch
```diff
@@ -0,0 +1,160 @@
+From 84394ab52b5eedc5c1220b7988a44db617b98691 Mon Sep 17 00:00:00 2001
+From: Tim Anderson <tim.anderson@foundries.io>
+Date: Fri, 22 Jan 2021 22:56:30 +0000
+Subject: [PATCH] Correct used paths
+
+This updates the utility to use paths that are correct for this
+distribution.
+
```
ricardosalveti, line 8: While the configpath change can easily land upstream, please add the following line (to track the patch upstream status):

Upstream-Status: Inappropriate [lmp specific]

meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/fpga-manager-script/fpga-manager-script_%.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI_append += " \
```
ricardosalveti, line 3: No need to use append with += (either append or +=).

ricardosalveti (changes requested):
Comments inline.

meta-lmp-bsp/conf/machine/include/ultrazed-lmp.inc
Tim-Anderson: will do

meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/bitstream/bitstream-extraction_git.bb
Tim-Anderson: the kernel driver will only load from /lib/firmware/... so the symlink has to point to someplace that is writable. and since the utility creates the /var/lib/firmware/bitstream automatically I do not think a systemd operation is strictly neccessary.

meta-lmp-bsp/dynamic-layers/xilinx-tools/recipes-bsp/bitstream/bitstream-extraction_git.bb
Tim-Anderson: understood, I will add that when I look at the u-boot pieces

meta-lmp-bsp/conf/machine/uz3eg-iocc.conf
```diff
@@ -12,3 +12,7 @@ HDF_MACHINE ?= "zcu102-zynqmp"
 HDF_BASE = "file://"
 HDF_PATH = "system.dsa"
 HDF_EXT = "dsa"
+
+# bitstream support
+MACHINE_FIRMWARE_append_uz3eg-iocc = " bitstream-extraction fpga-manager-script"
```
ricardosalveti, line 17: No need for append_uz3eg-iocc as this is already the board configuration file, just append works.

ricardosalveti (approved):
LGTM

PR #272, opened 2021-03-04
bsp: Add mfgtool files for create fuse and close script for imx6ullevk

Tim-Anderson:

### DIFF
--- a/meta-lmp-bsp/recipes-support/mfgtool-files/mfgtool-files/imx6ullevk/gen_close.bat
+++ b/meta-lmp-bsp/recipes-support/mfgtool-files/mfgtool-files/imx6ullevk/gen_close.bat
@@ -1,0 +1,97 @@
+@echo off
+
+:: Copyright (C) 2021 Foundries.IO
+::
+:: SPDX-License-Identifier: MIT
+::
+
+:: this will generate the nxp uuu.exe closing script based on the srktool *fuse.bin file
+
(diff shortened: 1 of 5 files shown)


mike-scott:
@Tim-Anderson @ricardosalveti I wonder if these scripts wouldn't be better landed in `lmp-tools`

mike-scott:
Ah.. NM, I see they are kind of hard-coded to imx6ullevk:
```
SDP: boot -f SPL-mfgtool.signed -dcdaddr 0x0907000 -cleardcd
```

ricardosalveti:
> @Tim-Anderson @ricardosalveti I wonder if these scripts wouldn't be better landed in `lmp-tools`

I think it is probably better to land at lmp-tools because the flow is a bit confusing, as you use the script to generate another script that should then be also provided via mfgtools by default.

We can have a soc family specific folder at https://github.com/foundriesio/lmp-tools/tree/master/security/imx_hab4.

Tim-Anderson:
I understand what you are saying. and I am okay with that as long as we do
not put actual fuse and close scripts in the mfgtool build then.

ricardosalveti:
> I understand what you are saying. and I am okay with that as long as we do not put actual fuse and close scripts in the mfgtool build then.

Yup, I think the expected flow would be for our customer to use the scripts from lmp-tools and then include as part of meta-subscriber-overrides.

Tim-Anderson:
Moving to lmp-tools

PR #274, opened 2021-03-05
bsp: recipes-kernel: wireguard: add option to opt-out of ext module

mike-scott:
Currently, only linux-lmp and linux-lmp-rt (both 5.10.y atm) opt out of
building the external wireguard module which applies to kernels < 5.6.y.

Let's provide a way for edge case kernels to also opt-out of the module:
KERNEL_BUILTIN_WIREGUARD = "1"


ricardosalveti (approved):
LGTM

PR #275, opened 2021-03-05
add wayland lmp distro (v2)

mike-scott:
This series adds a new distro configuration to support Wayland/Weston with Foundries.io branding.

- allow user to build with `DISTRO=lmp-wayland` as an easy way of adding Wayland / Weston to the platform build
- added a feature include file for subscribers to include needed components
- added some default FIO branding such as background and pslash (these can be overridden in the Factory)
- handle imx build break where [shell]  is uncommented in weston.ini (it's already uncommented in our default)


mike-scott:
I notice when testing this:
```
WARNING: lmp-factory-image-1.0-r0 do_image_ostree: Data in /opt directory is not preserved by OSTree. Consider moving it under /usr
```
Seems that vim82 package is now creating something in /opt

mike-scott:
One other note: I moved the weston.ini under lmp-wayland so that it would take priority over the other default weston.ini files.  There's another option for creating specific versions of the weston.ini for our reference HW:
imx/*
zynqmp/*
etc.

EmbeddedAndroid:
Looks good, just tested this for Intel, working fine there.

mike-scott:
Let me add some commit descriptions and I'll repush before merging.

munoz0raul:
Just tested it on iMX8mm-EVK and it is working fine as well.

Just want to remember that Wayland + Weston launched as root:

```
ls /run/user/0/
bus  systemd  use-g2d-renderer	wayland-0  wayland-0.lock
ps aux | grep weston
root         526  0.1  1.0 274324 21592 tty7     Ssl+ 18:18   0:00 /usr/bin/weston --modules=systemd-notify.so
```

So we might have a problem when launching chromium application as root.

munoz0raul:
@mike-scott Are you planning to update the default lmp-factory-image.bb with the line below:

`require ${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'recipes-samples/images/lmp-feature-wayland.inc', '', d)}
`

https://source.foundries.io/factories/unique-name/meta-subscriber-overrides.git/tree/recipes-samples/images/lmp-factory-image.bb#n11

Or each customer will need to do this change by themself?

mike-scott:
> @mike-scott Are you planning to update the default lmp-factory-image.bb with the line below:
> 
> `require ${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'recipes-samples/images/lmp-feature-wayland.inc', '', d)} `
> 
> https://source.foundries.io/factories/unique-name/meta-subscriber-overrides.git/tree/recipes-samples/images/lmp-factory-image.bb#n11
> 
> Or each customer will need to do this change by themself?

@munoz0raul Once this gets merged, I'm going to open a PR for the Factory template to include it there as well.

munoz0raul:
@mike-scott my weston display is turning off after some time.

In the demo Factory I fixed that by adding idle-time=0 in the weston.ini
`idle-time=0`
Have you seen the behavior?

mike-scott:
- Rebased
- Commit descriptions added

mike-scott:
> @mike-scott my weston display is turning off after some time.
> 
> In the demo Factory I fixed that by adding idle-time=0 in the weston.ini
> `idle-time=0`
> Have you seen the same behavior?

@munoz0raul Yes, this is intended behavior as a default.  (To avoid screen burn).

meta-lmp-base/conf/distro/lmp-wayland.conf
```diff
@@ -0,0 +1,9 @@
+require conf/distro/lmp.conf
+
+DISTRO = "lmp-wayland"
+DISTROOVERRIDES = "lmp:lmp-wayland"
+DISTRO_NAME = "Linux-microPlatform Wayland"
+
+DISTRO_FEATURES_append = " wayland opengl"
+
+IMAGE_LINGUAS ?= "en-us"
```
ricardosalveti, line 9: This is already set in lmp.config and with the same value, probably not needed.

meta-lmp-base/recipes-graphics/wayland/weston_%.bbappend
```diff
@@ -0,0 +1,6 @@
+# Weston with Xwayland support (requires X11 and Wayland)
+PACKAGECONFIG[xwayland] = "-Dxwayland=true,-Dxwayland=false,libxcursor"
```
ricardosalveti, line 2: OE already has the following:

`PACKAGECONFIG[xwayland] = "-Dxwayland=true,-Dxwayland=false"`

Why are we replicating this, for libxcursor?

meta-lmp-base/recipes-graphics/wayland/weston_%.bbappend
```diff
@@ -0,0 +1,6 @@
...
+
+# Clients support
+SIMPLE_CLIENTS = "all"
+PACKAGECONFIG[clients] = "-Dsimple-clients=${SIMPLE_CLIENTS} -Ddemo-clients=true,-Dsimple-clients= -Ddemo-clients=false"
```
ricardosalveti, line 6: From OE-core:

`PACKAGECONFIG[clients] = "-Dsimple-clients=all -Ddemo-clients=true,-Dsimple-clients= -Ddemo-clients=false"`

Why are we replicating this as well?

ricardosalveti:
Also, please split "base: graphics: wayland: add FIO branding" into 3 separated commits, as psplash can be used without wayland.

meta-lmp-base/conf/distro/lmp-wayland.conf
mike-scott: Good catch!  Fixing.

meta-lmp-base/recipes-graphics/wayland/weston_%.bbappend
mike-scott: Not sure.  And TBH, we aren't enabling xwayland in this PR.  I can drop this for now and revisit later.

meta-lmp-base/recipes-graphics/wayland/weston_%.bbappend
mike-scott: This was a left over change from when `apalis-imx6` setup SIMPLE_CLIENTS differently (for some reason).  I'll drop this change as well.

mike-scott:
> Also, please split "base: graphics: wayland: add FIO branding" into 3 separated commits, as psplash can be used without wayland.

@ricardosalveti How about 2 commits: core change for `psplash` and graphics change for wayland (not that `weston_%.bbappend` file will be dropped?

mike-scott:
- Rebased
- Re-ordered commits so that recipes were added prior to feature include / image additions
- Split `base: graphics: wayland: add FIO branding` into 2 commits:
  base: core: add default Foundries.io psplash logo
  base: graphics: wayland: add default FIO settings
- dropped weston_%.bbappend

mike-scott:
- Added my "Signed-off-by" to these commits as I did the rework.

ricardosalveti (approved):
LGTM

PR #276, opened 2021-03-06
WIP: bsp: u-boot-fio: add lmp-spl.cfg defaults and reduce lmp.cfg

mike-scott:
WIP PATCHSET DO NOT MERGE

Let's establish LmP defaults for SPL with a config fragment: lmp-spl.cfg.
Then reduce the lmp.cfg fragments to only the basic needed changes per machine.


mike-scott:
Just getting this started.  I'll split this up into commits in a bit and provide the diff vs. current config

igoropaniuk:
@mike-scott could you please also add https://github.com/foundriesio/meta-lmp/commit/d4986f5a6e2b4257c50ced7c84aedeec14350989 to your patch (it's not set to y by default in your changes)

angolini:
Do we have an internal jira ticket to track this work? The PR has no update since Mar 22, 2023 so I turned it into draft, but I would like to close and track it as internal jira instead

PR #279, opened 2021-03-09
Erase boot0 prior to flashing bootloader images

igoropaniuk:


ricardosalveti (approved):
I think this is a safe option to have, so LGTM

PR #284, opened 2021-03-12
bsp: mfgtool-files: imx6ullevk: fix erase offset

ricardosalveti:
There is no eMMC device available by default on imx6ullevk, so taking
into account the partition layout and correct offset when erasing the
bootloader area.


ricardosalveti:
Will merge to fix the regression, we can improve as we go.

PR #290, opened 2021-03-16
base: lmp-feature-wayland: add xset when xwayland is enabled

mike-scott:
Subscribers may choose to use xset to configure x options, so let's
add that to the build for xwayland users.


ricardosalveti (approved):
LGTM

mike-scott:
Dropping this PR.  It's not working as expected.

PR #291, opened 2021-03-19
Drop unit address from FIT

ricardosalveti:
As an effort to move away from using unit address when loading the FIT
components (u-boot CVE-2021-27138), allow the fit node separator to be
set during build time via the FIT_NODE_SEPARATOR variable, and have it
set to "-" by default.

Targets that are already running based on the older boot.cmd will require 
FIT_NODE_SEPARATOR to be set to "@", otherwise it won't boot the new
kernel fitimage during an update (and will rollback to previous).

If flashing from scratch, nothing needs to be done and it will already use
the correct syntax (which will be a requirement on u-boot 2021.04), but know
that this will only work if the entire fleet is being flashed with the new format,
because we currently don't support updating u-boot and boot.cmd (will later
once @igoropaniuk finishes the bootloader update scheme).


meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
```diff
@@ -72,26 +206,26 @@ fitimage_emit_section_config() {
...
+                conf${FIT_NODE_SEPARATOR}${3} {
                         description = "${8} ${conf_desc}";
                         ${kernel_line}
                         ${fdt_line}
                         ${ramdisk_line}
                         ${setup_line}
                         ${fpga_line}
                         ${loadable_line}
-                        hash@1 {
+                        hash-1 {
```
Tim-Anderson, line 228: could this be ${FIT_NODE_SEPARATOR} here instead of the hard coded - ?

This would be he same question for the rest of these occurances.

ricardosalveti: There is no need to use that here because we're not manually referring to this node, so we can simply just move to - instead.

Now for dtb, kernel, initrd, etc this is a bit more complicated because we're referring to them manually in our boot scripts.

angolini (approved):

Tim-Anderson (approved):
LGTM

meta-lmp-base/recipes-bsp/u-boot/u-boot-ostree-scr-fit.bb
```diff
@@ -16,12 +16,16 @@ B = "${WORKDIR}/build"
 # fitImage Hash Algo
 FIT_HASH_ALG ?= "sha256"
 
+# Allow transition to cover CVE-2021-27097 and CVE-2021-27138
+FIT_NODE_SEPARATOR ?= "-"
```
mike-scott, line 20: If the `kernel-lmp-fitimage.bbclass` already sets the optional default value, do we need it here again?  Or was this to make it obvious that this recipe uses an optional default value here?  And it can be overridden?

meta-lmp-base/recipes-bsp/u-boot/u-boot-ostree-scr.bb
```diff
@@ -17,12 +17,16 @@ KERNEL_BOOTCMD_aarch64 ?= "booti"
 S = "${WORKDIR}"
 B = "${WORKDIR}/build"
 
+# Allow transition to cover CVE-2021-27097 and CVE-2021-27138
+FIT_NODE_SEPARATOR ?= "-"
```
mike-scott, line 21: Same here about having this defined twice

mike-scott:
@ricardosalveti I know we use the "@@" template notation elsewhere, but this makes it really hard to parse replacing a single "@" with the template "@@FIT_NODE_SEPARATOR@@" value.  We might consider using a slightly different template notation here.

As the very least, it would make easier use of grep looking for where this needs to be fixed.

meta-lmp-base/recipes-bsp/u-boot/u-boot-ostree-scr-fit.bb
ricardosalveti: To avoid setting it twice we would need to make it part of the lmp distro configuration, as it gets parsed before the rest.

Both cases would work, I just didn't want to change the distro default configuration for something that should, hopefully, be temporary :-)

ricardosalveti:
> @ricardosalveti I know we use the "@@" template notation elsewhere, but this makes it really hard to parse replacing a single "@" with the template "@@FIT_NODE_SEPARATOR@@" value. We might consider using a slightly different template notation here.
> 
> As the very least, it would make easier use of grep looking for where this needs to be fixed.

Sure, can change to another separator.

mike-scott:
> > @ricardosalveti I know we use the "@@" template notation elsewhere, but this makes it really hard to parse replacing a single "@" with the template "@@FIT_NODE_SEPARATOR@@" value. We might consider using a slightly different template notation here.
> > As the very least, it would make easier use of grep looking for where this needs to be fixed.
> 
> Sure, can change to another separator.

Ugh, now I'm kind of waffling.  We're going to create a unique "thing" here that will be hard to remember everytime we implement a new boot.cmd script.

ricardosalveti:
> > > @ricardosalveti I know we use the "@@" template notation elsewhere, but this makes it really hard to parse replacing a single "@" with the template "@@FIT_NODE_SEPARATOR@@" value. We might consider using a slightly different template notation here.
> > > As the very least, it would make easier use of grep looking for where this needs to be fixed.
> > 
> > 
> > Sure, can change to another separator.
> 
> Ugh, now I'm kind of waffling. We're going to create a unique "thing" here that will be hard to remember everytime we implement a new boot.cmd script.

Right, at least this format is known (that was my initial thinking). And factory users wouldn't necessarily have to update when sticking with the older format (until they decide to switch to the new format), all that would be required is to override FIT_NODE_SEPARATOR at lmp-factory-custom.inc.

mike-scott (approved):
LGTM

PR #294, opened 2021-03-23
recipes-support: libfyaml for aktualizr-lite

ldts:
This tool (binary, not libraries) is used by aklite to convert yaml to json

### DIFF
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
@@ -16,9 +16,10 @@ SRC_URI_append_libc-musl = " \
     file://utils.c-disable-tilde-as-it-is-not-supported-by-musl.patch \
 "
 
-PACKAGECONFIG += "${@bb.utils.filter('SOTA_EXTRA_CLIENT_FEATURES', 'fiovb', d)}"
+PACKAGECONFIG += "${@bb.utils.filter('SOTA_EXTRA_CLIENT_FEATURES', 'fiovb', d)} libfyaml"
 PACKAGECONFIG[fiovb] = ",,,optee-fiovb aktualizr-fiovb-env-rollback"
 PACKAGECONFIG[ubootenv] = ",,u-boot-fw-utils,u-boot-fw-utils u-boot-default-env aktualizr-uboot-env-rollback"
+PACKAGECONFIG[libfyaml] = ",,,libfyaml"
 
 SYSTEMD_PACKAGES += "${PN}-lite"
 SYSTEMD_SERVICE_${PN}-lite = "aktualizr-lite.service"
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 BRANCH_lmp = "master"
-SRCREV_lmp = "be4532fc1076cca4f198c322e9c0676e3111dd19"
+SRCREV_lmp = "4be89e29494d3d7e1ff912e50fa7ac3cd15624c5"
 
 SRC_URI_lmp = "gitsm://github.com/foundriesio/aktualizr-lite;branch=${BRANCH};name=aktualizr \
     file://aktualizr.service \
--- a/meta-lmp-base/recipes-support/libfyaml/libfyaml_0.6.3.bb
+++ b/meta-lmp-base/recipes-support/libfyaml/libfyaml_0.6.3.bb
@@ -1,0 +1,11 @@
+SUMMARY = "A fancy 1.3 YAML and JSON parser/writer."
+HOMEPAGE = "https://github.com/pantoniou/libfyaml"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=6399094fbc639a289cfca2d660c010aa"
+
+SRC_URI = "https://github.com/pantoniou/libfyaml/releases/download/v${PV}/libfyaml-${PV}.tar.gz"
+SRC_URI[sha256sum] = "aba6e5b1667bb5a05318f0ad70c617345f2a9e5ce79b37ff1e5322162c9a033e"
+
+S = "${WORKDIR}/libfyaml-${PV}"
+
+inherit autotools pkgconfig


meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
```diff
@@ -0,0 +1,16 @@
+SUMMARY = "A fancy 1.2 YAML and JSON parser/writer."
+HOMEPAGE = "https://github.com/pantoniou/libfyaml"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=6399094fbc639a289cfca2d660c010aa"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}-${PV}:"
+
+SRC_URI = "git://github.com/pantoniou/libfyaml.git"
+SRCREV = "de89df9041b3a72b258680d6c8f6df31c5cee28a"
```
ricardosalveti, line 9: Do we need to use git or using the latest known release is enough? Released snapshots are preferred (then renaming the recipe from _git.bb to _0.6.bb).

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
```diff
@@ -0,0 +1,16 @@
+SUMMARY = "A fancy 1.2 YAML and JSON parser/writer."
+HOMEPAGE = "https://github.com/pantoniou/libfyaml"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=6399094fbc639a289cfca2d660c010aa"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}-${PV}:"
```
ricardosalveti, line 6: No need for this line.

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
```diff
@@ -0,0 +1,16 @@
...
+
+SRC_URI = "git://github.com/pantoniou/libfyaml.git"
+SRCREV = "de89df9041b3a72b258680d6c8f6df31c5cee28a"
+
+DEPENDS = " pkgconf "
```
ricardosalveti, line 11: Is this needed?

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
```diff
@@ -0,0 +1,16 @@
...
+
+S = "${WORKDIR}/git"
+
+inherit autotools
+PARALLEL_MAKE = ""
```
ricardosalveti, line 16: Why setting parallel_make to ""?

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
ldts: yes

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
ldts: someone somewhere mentioned that parallel builds might mess autoconf (I didnt validate it though, but seemed possible)

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
ldts: ok

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
ldts: I havent tried, will test and update then

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
ldts: will remove and check the build

ldts:
@rsalveti these autotools dont seem to be installing in the final image?

jramirez@trex build-lmp$ find . -name "fy-tool"
./tmp-lmp/work/cortexa53-crypto-lmp-linux/libfyaml/0.6-r0/image/usr/bin/fy-tool
./tmp-lmp/work/cortexa53-crypto-lmp-linux/libfyaml/0.6-r0/package/usr/bin/.debug/fy-tool
./tmp-lmp/work/cortexa53-crypto-lmp-linux/libfyaml/0.6-r0/package/usr/bin/fy-tool
./tmp-lmp/work/cortexa53-crypto-lmp-linux/libfyaml/0.6-r0/packages-split/libfyaml/usr/bin/fy-tool
./tmp-lmp/work/cortexa53-crypto-lmp-linux/libfyaml/0.6-r0/packages-split/libfyaml-dbg/usr/bin/.debug/fy-tool
./tmp-lmp/work/cortexa53-crypto-lmp-linux/libfyaml/0.6-r0/build/src/fy-tool
./tmp-lmp/work/cortexa53-crypto-lmp-linux/libfyaml/0.6-r0/build/src/.libs/fy-tool


is this a bug or am I missing something in the recipe??

ldts:
ok I'll cancel the PR until I figure out the install part. will try again later :)

ldts:
added it to the factory.inc....although since it is only used by aktualizr-lite perhaps they should go together.
or with the ota-utils...dunno, so many choices...

ricardosalveti:
> added it to the factory.inc....although since it is only used by aktualizr-lite perhaps they should go together.
> or with the ota-utils...dunno, so many choices...

If used/required by aktualizr-lite, just make it a dependency of aktualizr.

ricardosalveti:
```
PACKAGECONFIG += "${@bb.utils.filter('SOTA_EXTRA_CLIENT_FEATURES', 'fiovb', d)}"
PACKAGECONFIG[fiovb] = ",,,optee-fiovb aktualizr-fiovb-env-rollback"
PACKAGECONFIG[ubootenv] = ",,u-boot-fw-utils,u-boot-fw-utils u-boot-default-env aktualizr-uboot-env-rollback"
```

You could extend that to cover libfyaml, like:
`PACKAGECONFIG[fyaml] = ",,,libfyaml"
`

And adding fyaml to PACKAGECONFIG by default.

ldts:
right and I was thinking about messing with that but what would happen if other clients decide to use this tool? 
it is not much different than aktualizr-lite depending on docker-cli (actually is pretty much the same sort of dependency)
so I opted for just adding it to the factory inc.

"And adding fyaml to PACKAGECONFIG by default."

what do you mean? in lmp.inc? I am not too familiar with this bit...so is the typical use case of this construct for binaries calling other binaries? seems strange capturing dependencies between binaries in this way...

meta-lmp-base/recipes-support/libfyaml/libfyaml_git.bb
ricardosalveti: yeah, that is preferred (inherit).

ricardosalveti:
> right and I was thinking about messing with that but what would happen if other clients decide to use this tool?
> it is not much different than aktualizr-lite depending on docker-cli (actually is pretty much the same sort of dependency)
> so I opted for just adding it to the factory inc.
> 
> "And adding fyaml to PACKAGECONFIG by default."
> 
> ```
> PACKAGECONFIG += "libfyaml"
> PACKAGECONFIG[libfyaml] = ",,,libfyaml"
> ```

Right, this change would need to be part of the aktualizr_%.bbappend we have in meta-lmp-base.

> what do you mean something like the above? I am not too familiar with this bit...so is the typical use case of this construct for binaries calling other binaries? seems strange capturing dependencies between binaries in this way...

IIRC the syntax here is:

`PACKAGECONFIG[foobar] = "optiontoenableduringbuildtime,optiontodisableduringbuildtime,builddependency,runtimedependency"
`

So you could have this as a build option in aktualizr as well if needed. Since it seems you need during runtime, you have to add it at the last block.

meta-lmp-base/recipes-support/libfyaml/libfyaml_0.6.bb
```diff
@@ -0,0 +1,12 @@
...
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${S}/LICENSE;md5=6399094fbc639a289cfca2d660c010aa"
+
+SRC_URI = "https://github.com/pantoniou/libfyaml/archive/refs/tags/v${PV}.tar.gz"
+SRC_URI[sha256sum] = "f28113cf3a3a62ccb6c73b0f24072cc8cee3af87531029718511419d38bc2621"
+
+S = "${WORKDIR}/libfyaml-${PV}"
+
+inherit autotools pkgconfig
+
```
ricardosalveti, line 12: Just drop this last empty line.

ricardosalveti:
`WARNING: libfyaml-0.6-r0 do_package_qa: QA Issue: libfyaml: SRC_URI uses unstable GitHub archives [src-uri-bad]`

I think we need to use the release link instead.

ricardosalveti:
https://github.com/pantoniou/libfyaml/archive/refs/tags/v0.6.tar.gz

> `WARNING: libfyaml-0.6-r0 do_package_qa: QA Issue: libfyaml: SRC_URI uses unstable GitHub archives [src-uri-bad]`
> 
> I think we need to use the release link instead.

Seems release is not available for 0.6, sigh.

In this case, use git then and use the hash from the 0.6 tag.

ricardosalveti:
> right and I was thinking about messing with that but what would happen if other clients decide to use this tool?

Then they will need to rdepend on it as well.

In our case, as we know which tool is expecting to use it, it is better to just make it a dependency the tool we're already installing, as then we don't need to change the image.

ldts:
the issue is that .tarball-version is missing in the tarball
should contain something like: 
`0.6.0`

meta-lmp-base/recipes-support/libfyaml/libfyaml_0.6.bb
ldts: ok. do you know some magic to add a file to the tarball before the autotools recipe starts processing it?

ldts:
I believe this should get merged rather than waiting for v.07 (or whatever version comes along). we can always bump it later

ldts:
> I believe this should get merged rather than waiting for v.07 (or whatever version comes along). we can always bump it later

it doesnt matter anymore...just got an update from Pantelis

angolini (approved):

mike-sul:
@ricardosalveti Can we merge it so I can test the latest aklite dev version (deps on fy-tool) against the latest lmp.

ricardosalveti:
> @ricardosalveti Can we merge it so I can test the latest aklite dev version (deps on fy-tool) against the latest lmp.

Merging just the recipe itself is fine, as that won't influence anything for now, let me see if I can just cherry-pick that.

ricardosalveti:
I pushed just the recipe at https://github.com/foundriesio/meta-lmp/commit/1dd1d1f487af829a7e477bd1b35e36ff693c3d11, @ldts please update this PR with just the aktualizr-lite addition once there is an updated rev that requires it.

meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
```diff
@@ -19,6 +19,8 @@ SRC_URI_append_libc-musl = " \
 PACKAGECONFIG += "${@bb.utils.filter('SOTA_EXTRA_CLIENT_FEATURES', 'fiovb', d)}"
 PACKAGECONFIG[fiovb] = ",,,optee-fiovb aktualizr-fiovb-env-rollback"
 PACKAGECONFIG[ubootenv] = ",,u-boot-fw-utils,u-boot-fw-utils u-boot-default-env aktualizr-uboot-env-rollback"
+PACKAGECONFIG += "libfyaml"
```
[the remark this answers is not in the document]
ldts: ok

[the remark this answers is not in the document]
ldts: @ricardosalveti done

ricardosalveti:
Manually merged after resolving a minor conflict and removing one patch which was already added in v80.

mike-sul:
@ricardosalveti Thanks, so the libfyaml stuff is in the master/HEAD of lmp-manifest but is not in v80, which is the right thing to do.

PR #296, opened 2021-03-24
initramfs-framework: create /var/lock later

liuming50:
In run-tmpfs.patch, we had changed /run to be a tmpfs mountpoint, and
/var/lock is a symbolic link to ../run/lock, so we shall move the
'mkdir' after /run created, or else we get a error as follows:
| mkdir: can't create directory '/var/lock': No such file or directory

### DIFF
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/run-tmpfs.patch
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/run-tmpfs.patch
@@ -24,11 +24,11 @@ index c71ce0c..79cf5c1 100755
  
  # initialize /proc, /sys, /run/lock and /var/lock
 -mkdir -p /proc /sys /run/lock /var/lock
-+mkdir -p /proc /sys /run /var/lock
++mkdir -p /proc /sys /run
  mount -t proc proc /proc
  mount -t sysfs sysfs /sys
 +mount -t tmpfs tmpfs /run
-+mkdir -p /run/lock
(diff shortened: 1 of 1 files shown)


ricardosalveti:
@liuming50 in your case, where is /var being created and how? At least from what I see here, /var/lock (at the initrd level) is not a link to /run/lock by default, so trying to understand what in your env makes you to face this error.

ricardosalveti:
And if /var/lock is indeed a sym link, we don't actually need to call mkdir for it.

liuming50:
> @liuming50 in your case, where is /var being created and how? At least from what I see here, /var/lock (at the initrd level) is not a link to /run/lock by default, so trying to understand what in your env makes you to face this error.

@ricardosalveti This error was observed when we added meta-security/meta-integrity layer with meta-lmp, I spent some time but did not figure out how it's created '/var/lock > ../run/lock', but I verified this fix should be able to let it work with both cases.

ricardosalveti:
> > @liuming50 in your case, where is /var being created and how? At least from what I see here, /var/lock (at the initrd level) is not a link to /run/lock by default, so trying to understand what in your env makes you to face this error.
> 
> @ricardosalveti This error was observed when we added meta-security/meta-integrity layer with meta-lmp, I spent some time but did not figure out how it's created '/var/lock > ../run/lock', but I verified this fix should be able to let it work with both cases.

Sure, will merge then as it won't really have any bad side effect and should indeed work fine for both.

ricardosalveti (approved):

PR #302, opened 2021-04-05
base: go: limit RISC-V cgo/pie support patches to riscv64

mike-scott:
These patches seem to be causing a regression when applied for all
arches.  Let's limit them to only riscv64.


ricardosalveti (approved):
LGTM

PR #303, opened 2021-04-06
base: aktualizr-lite: bump to fe447e0

mike-sul:
Relevant changes:
- fe447e0 fixup! get currently installed apps correctly

### DIFF
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 BRANCH_lmp = "master"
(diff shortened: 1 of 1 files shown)


doanac:
Looks good. This makes me think we should probably have some test-plan type thing associated with this type of merge that @mwasilew can help do verification with.

mwasilew:
If there is LmP build on creation of PR, this can be automatically tested before it's merged. At this point I can only see PRs to manifest built automatically. As for test plan part, there already is an entry in the doc, but no tests running yet.

ricardosalveti:
> If there is LmP build on creation of PR, this can be automatically tested before it's merged. At this point I can only see PRs to manifest built automatically. As for test plan part, there already is an entry in the doc, but no tests running yet.

That would be ideal, but for us being able to test once we merge lmp-manifest is already a big plus at this stage.

ricardosalveti (approved):
LGTM

PR #304, opened 2021-04-06
apalis-imx6 pl310 fix

mike-scott:
- Linux patches to fix Linux init of PL310 caches.
- Bump optee-os SHAs to include PL310 tuning changes

These changes fix performance issues for imx6qdl HW when running OP-TEE.


Tim-Anderson (approved):
LGTM

ricardosalveti:
Should we make the append override based on mx6/mx6q instead?

mike-scott:
> Should we make the append override based on mx6/mx6q instead?

I suppose we could.  I didn't want to cause an issue with `mx6ul`  where this isn't needed.

ricardosalveti (approved):

PR #305, opened 2021-04-07
Add apps-early-start recipe

doanac:
Its often desirable to start compose-apps (containers) before a device
is registered and aklite is running. This recipe creates a one shot
service that will start apps if aklite is not yet setup.

### DIFF
--- a/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start.bb
+++ b/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start.bb
@@ -1,0 +1,19 @@
+SUMMARY = "A systemd oneshot helper to start compose apps as early as possible"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+SRC_URI = " \
+	file://compose-apps-early-start.service \
+	file://compose-apps-early-start \
+"
+
+inherit systemd
+
+SYSTEMD_SERVICE_${PN} = "compose-apps-early-start.service"
+
+do_install() {
+	install -d ${D}${systemd_system_unitdir}
+	install -m 0644 ${WORKDIR}/compose-apps-early-start.service ${D}${systemd_system_unitdir}/
+	install -d ${D}${bindir}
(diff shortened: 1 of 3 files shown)


ricardosalveti:
Would you mind renaming to compose-apps-eary-start? Just to avoid confusion later on, as this is specific to our compose apps format.

meta-lmp-base/recipes-support/apps-early-start/apps-early-start/apps-early-start.service
```diff
@@ -0,0 +1,14 @@
...
+Before=lmp-device-auto-register.service
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/bin/lmp-apps-early-start
+
+[Install]
+WantedBy=multi-user.target
+
```
ricardosalveti, line 14: Empty line, can be removed.

meta-lmp-base/recipes-support/apps-early-start/apps-early-start/lmp-apps-early-start
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh -e
+
+if [ -f /var/sota/sql.db ] ; then
```
ricardosalveti, line 3: lmp-device-auto-register has `ConditionPathExists=!/var/sota/sql.db`, so this condition will never really be false when started from systemd.

Do we expect people to manually call this script? Otherwise we can probably drop this and reduce one if block.

ricardosalveti: Actually, since it is before we might have it executed, but would need to check (on the dependency chain lmp-device-auto-register.service is never executed when /var/sota/sql.db exists, so this job might not be executed as consequence).

To be safe we could also have the same condition and remove this if block from the script.

ricardosalveti:
Also, mind prepending "base: " to your commit subject? Just because this is for the meta-lmp-base layer.

meta-lmp-base/recipes-support/apps-early-start/apps-early-start/lmp-apps-early-start
doanac: k, i think i've got this fixed in the force push

ricardosalveti (approved):
LGTM, thanks for the changes.

PR #307, opened 2021-04-09
initramfs-framework: create /var/lock later

liuming50:
In run-tmpfs.patch, we had changed /run to be a tmpfs mountpoint, and
/var/lock is a symbolic link to ../run/lock, so we shall move the
'mkdir' after /run created, or else we get a error as follows:
| mkdir: can't create directory '/var/lock': No such file or directory


liuming50:
Hi, maintainers, can we merge this to the LTS dunfell?

ricardosalveti (approved):

PR #309, opened 2021-04-09
Pre-load boot.itb from U-Boot FIT image

igoropaniuk:


meta-lmp-base/classes/uboot-fitimage.bbclass
```diff
@@ -25,6 +25,7 @@ ATF_BINARY ?= "arm-trusted-firmware.bin"
 ATF_SUPPORT = "${@bb.utils.contains('EXTRA_IMAGEDEPENDS', 'virtual/trusted-firmware-a', 'true', 'false', d)}"
 
 do_compile[depends] += " virtual/optee-os:do_deploy"
+do_compile[depends] += " u-boot-default-script:do_deploy"
```
ricardosalveti, line 28: From the build perspective I think we should make this optional, kind of as done for ATF, since then we can enable/disable this support based on how the board is configured (e.g. allow u-boot.itb with normal boot script).

Like:

```
BOOTSCR_SUPPORT = "${@bb.utils.contains('PREFERRED_PROVIDER_u-boot-default-script', 'u-boot-ostree-scr-fit', 'true', 'false', d)}"
do_compile[depends] += " ${@'u-boot-default-script:do_deploy' if d.getVar('BOOTSCR_SUPPORT') == 'true' else ''}"

```
And then make the fit block conditional.

meta-lmp-base/classes/uboot-fitimage.bbclass
```diff
@@ -56,11 +58,11 @@ uboot_fitimage_assemble() {
...
 		optee_type="standalone"
 		config_firmware="atf"
-		config_loadables='"uboot", "optee"';
+		config_loadables='"uboot", "optee", "bootscr"';
 	else
 		optee_type="firmware"
 		config_firmware="optee"
-		config_loadables='"uboot"';
+		config_loadables='"uboot", "bootscr"';
 	fi
```
ricardosalveti, line 66: Then you could add another if block based on BOOTSCR_SUPPORT and extend config_loadables as needed.

meta-lmp-base/classes/uboot-fitimage.bbclass
```diff
@@ -208,7 +221,7 @@ do_deploy_prepend() {
 	else
 		cd ${B}
 		UBOOT_LOAD_ADDR=`grep 'define CONFIG_SYS_TEXT_BASE' u-boot.cfg | cut -d' ' -f 3`
-		uboot_fitimage_assemble ${UBOOT_ITB_BINARY} ${UBOOT_LOAD_ADDR} ${OPTEE_LOAD_ADDR} ${ATF_LOAD_ADDR}
+		uboot_fitimage_assemble ${UBOOT_ITB_BINARY} ${UBOOT_LOAD_ADDR} ${OPTEE_LOAD_ADDR} ${ATF_LOAD_ADDR} ${BOOTSCR_PRELOAD_ADDR}
```
ricardosalveti, line 224: Follow the other vars, use BOOTSCR_LOAD_ADDR.

meta-lmp-base/classes/uboot-fitimage.bbclass
igoropaniuk: I think we can store pre-load address of boot.scr instead of just true/false

ricardosalveti: Sure, as long it can also be enabled/disabled based on the default u-boot-default-script provider, we should be fine.

meta-lmp-bsp/recipes-bsp/u-boot/u-boot-fio/imx8mmevk/lmp.cfg
```diff
@@ -47,7 +47,7 @@ CONFIG_BOOTCOUNT_LIMIT=y
 CONFIG_BOOTCOUNT_ENV=y
 CONFIG_BOOTDELAY=-2
 CONFIG_USE_BOOTCOMMAND=y
-CONFIG_BOOTCOMMAND="fatload mmc 2:1 ${loadaddr} /boot.itb; setenv verify 1; source ${loadaddr}; reset"
+CONFIG_BOOTCOMMAND="setenv verify 1; source 0x43800000; reset"
```
igoropaniuk, line 50: @ricardosalveti @mike-scott could you please confirm that we're ok about using this address (that works on my side, but just to be sure that we are not going to use for something else)

meta-lmp-base/classes/uboot-fitimage.bbclass
mike-scott: @ricardosalveti comment here didn't get addressed.  Was this change rejected?

meta-lmp-bsp/recipes-bsp/u-boot/u-boot-fio/imx8mmevk/lmp.cfg
mike-scott: address LGTM

mike-scott (approved):
LGTM

igoropaniuk:
@ricardosalveti Do you have any additional comments/suggestions?

meta-lmp-base/classes/uboot-fitimage.bbclass
```diff
@@ -208,7 +231,7 @@ do_deploy_prepend() {
 	else
 		cd ${B}
 		UBOOT_LOAD_ADDR=`grep 'define CONFIG_SYS_TEXT_BASE' u-boot.cfg | cut -d' ' -f 3`
-		uboot_fitimage_assemble ${UBOOT_ITB_BINARY} ${UBOOT_LOAD_ADDR} ${OPTEE_LOAD_ADDR} ${ATF_LOAD_ADDR}
+		uboot_fitimage_assemble ${UBOOT_ITB_BINARY} ${UBOOT_LOAD_ADDR} ${OPTEE_LOAD_ADDR} ${ATF_LOAD_ADDR} ${BOOTSCR_LOAD_ADDR}
```
ricardosalveti, line 234: I think we're also missing to add this at the first uboot_fitimage_assemble call (part of the if block).

meta-lmp-base/classes/uboot-fitimage.bbclass
```diff
@@ -45,6 +47,7 @@ uboot_fitimage_assemble() {
 	ubootloadaddr=${2}
 	opteeloadaddr=${3}
 	atfloadaddr=${4}
+	bootscrloadaddr=${5}
```
ricardosalveti, line 50: We should also make a condition to see if this variable is indeed valid (set by the board). Otherwise BOOTSCR_SUPPORT can end up being set to true (based on u-boot-ostree-scr-fit support), but missing BOOTSCR_LOAD_ADDR, resulting in an invalid fit.

ricardosalveti: Or we could extend the BOOTSCR_SUPPORT variable set to check for 2 conditions, PREFERRED_PROVIDER_u-boot-default-script and BOOTSCR_LOAD_ADDR not being empty.

igoropaniuk: fixed

meta-lmp-base/classes/uboot-fitimage.bbclass
igoropaniuk: fixed

ricardosalveti:
LGTM, just testing this locally now.

ricardosalveti (approved):

PR #310, opened 2021-04-13
base: classes: allow 64-bit addresses in fit image

Tim-Anderson:
I found that some platforms support 64-bit addressing with 4GB
or more of memory. This will check if the address is greater
than a 32-bit address and reformat the address so the dtc will
process it correctly.


mike-scott (approved):
LGTM

ricardosalveti (approved):
LGTM, thanks

PR #311, opened 2021-04-14
base: networkmanager: make bbappend specific to the 1.22 series

ricardosalveti:
Local patch and modification might not necessarily apply to any NM
version, so restrict to the base version available in meta-oe.


mike-scott (approved):
LGTM

PR #314, opened 2021-04-15
Switch u-boot-fio default revision to 2021.04+fio.

ricardosalveti:
Switch default (for !imx targets) revision to 2021.04+fio.

It was already set and used by rpi, just making it default for others (e.g. qemu, freedom, beagle)


mike-scott (approved):
LGTM

PR #317, opened 2021-04-19
base: dropbear: prefer ecdsa instead of rsa

ricardosalveti:
ECDSA is a lot faster to generate during first use, so prefer it instead
of RSA 2048 (default from OE).


angolini (approved):

Tim-Anderson (approved):
LGTM

PR #323, opened 2021-04-22
base: fiotools: add a recipe for fiotools

mike-sul:
- A recipe to build and install fiotools.
  It includes fiopush and fiocheck aimed to push an ostree repo to and check if it's synced with OSTreeHub.

Signed-off-by: Mike Sul <mike.sul@foundries.io>

PR #324, opened 2021-04-22
Use fiotools instead of garage

mike-sul:


mike-sul:
This https://github.com/foundriesio/meta-lmp/pull/323 should go first.

@ricardosalveti I couldn't figure out how to do inheritance properly and re-use the  "base" class implementation, so I just copy-paste. Your help is welcomed.

mike-sul:
@ricardosalveti I've checked the change for a different use-case looks fine. This CI job is for the case when it's merged and ci-scripts are not updated with `USE_FIOTOOLS` flag. https://github.com/foundriesio/ci-scripts/pull/125

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -56,3 +56,66 @@ IMAGE_CMD_ostree_append () {
...
+	if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
+                if [ -e "${SOTA_PACKED_CREDENTIALS}" ]; then
+                        "${1}" -repo "${OSTREE_REPO}" -creds "${SOTA_PACKED_CREDENTIALS}"
+                else
+                        bbwarn "SOTA_PACKED_CREDENTIALS file does not exist."
+                fi
+        else
+                bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
+        fi
+
```
ricardosalveti, line 70: No need for this extra empty line.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -56,3 +56,66 @@ IMAGE_CMD_ostree_append () {
...
+			unzip -p ${SOTA_PACKED_CREDENTIALS} root.json targets.pub targets.sec tufrepo.url 2>&1 >/dev/null || exit 0
+
+			ostree_target_hash=$(cat ${WORKDIR}/ostree_manifest)
+
+			garage-check --ref=${ostree_target_hash} \
+				--credentials=${SOTA_PACKED_CREDENTIALS} \
+				--cacert=${STAGING_ETCDIR_NATIVE}/ssl/certs/ca-certificates.crt
+		fi
+	fi
+}
```
ricardosalveti, line 121: We might be able to remove the duplication by prepending and returning if FIOTOOLS, let me do a quick check.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -56,3 +56,66 @@ IMAGE_CMD_ostree_append () {
...
+			fi
+		else
+			bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
+		fi
+	fi
+}
+
+do_image_garagecheck[depends] += "fiotools-native:do_populate_sysroot ca-certificates-native:do_populate_sysroot"
+IMAGE_CMD_garagecheck_lmp () {
+	if [ -n "${USE_FIOTOOLS}" ]; then
```
ricardosalveti, line 105: Same here.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -56,3 +56,66 @@ IMAGE_CMD_ostree_append () {
...
+        else
+                bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
+        fi
+
+}
+
+do_image_ostreepush[depends] += "fiotools-native:do_populate_sysroot ca-certificates-native:do_populate_sysroot"
+IMAGE_CMD_ostreepush_lmp () {
+	if [ -n "${USE_FIOTOOLS}" ]; then
+		bbwarn "Pushing repo by using fiopush"
```
ricardosalveti, line 76: Better to remove as this will generate a build warning, which is not really what is going on (useful for debugging, but we can safely remove).

You can change to bbnote if needed, but need a run to see if that shows up at the build log.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -56,3 +56,66 @@ IMAGE_CMD_ostree_append () {
...
+                else
+                        bbwarn "SOTA_PACKED_CREDENTIALS file does not exist."
+                fi
+        else
+                bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
+        fi
+
+}
+
+do_image_ostreepush[depends] += "fiotools-native:do_populate_sysroot ca-certificates-native:do_populate_sysroot"
```
ricardosalveti, line 73: Looking at the final generated output, we just need to add fiotools-native really, as ca-certificates will be set by meta-updater.

ricardosalveti:
I believe we can simplify this PR with the following diff:

```
diff --git a/meta-lmp-base/classes/lmp.bbclass b/meta-lmp-base/classes/lmp.bbclass
index 615caf8..7c97d90 100644
--- a/meta-lmp-base/classes/lmp.bbclass
+++ b/meta-lmp-base/classes/lmp.bbclass
@@ -56,3 +56,33 @@ IMAGE_CMD_ostree_append () {
        # Update default home path
        sed -i -e 's,:/home,:/var/rootdirs/home,g' usr/etc/passwd
 }
+
+run_fiotool_cmd () {
+       if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
+                if [ -e "${SOTA_PACKED_CREDENTIALS}" ]; then
+                        "${1}" -repo "${OSTREE_REPO}" -creds "${SOTA_PACKED_CREDENTIALS}"
+                else
+                        bbwarn "SOTA_PACKED_CREDENTIALS file does not exist."
+                fi
+        else
+                bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
+        fi
+}
+
+do_image_ostreepush[depends] += "fiotools-native:do_populate_sysroot"
+IMAGE_CMD_ostreepush_prepend_lmp () {
+       if [ -n "${USE_FIOTOOLS}" ]; then
+               run_fiotool_cmd "fiopush"
+               # manually return to avoid calling garage-push (set by meta-updater)
+               return
+       fi
+}
+
+do_image_garagecheck[depends] += "fiotools-native:do_populate_sysroot"
+IMAGE_CMD_garagecheck_prepend_lmp () {
+       if [ -n "${USE_FIOTOOLS}" ]; then
+               run_fiotool_cmd "fiocheck"
+               # manually return to avoid calling garage-check (set by meta-updater)
+               return
+       fi
+}
```

Let me know what you think.

meta-lmp-base/classes/lmp.bbclass
mike-sul: I tried `bbnote` and didn't see anything in the logs so decided to put `bbwarn` as these log messages are useful, we can see which type of `push` we use.

meta-lmp-base/classes/lmp.bbclass
mike-sul: ok, I'll check it out.

mike-sul: It looks like it's working. This is the run for USE_FIOTOOLS=0 https://ci.foundries.io/projects/msul-dev01/lmp/builds/793/ and this one for USE_FIOTOOLS=1 https://ci.foundries.io/projects/msul-dev01/lmp/builds/794/. It looks fine, the first run pushed to Treehub the second one to OSTreeHub (our solution) in both cases device was updated successfully.

ricardosalveti: Nice.

meta-lmp-base/classes/lmp.bbclass
ricardosalveti: Only problem is that we're trying to push the build to stop producing warnings and flag when a warning happens (we even produce a special file with all warnings, such as https://ci.foundries.io/projects/msul-dev01/lmp/builds/793/raspberrypi4-64/artifacts/other/bitbake_warning.log), and with this change every build will generate such warning (ones using fiotools).

Maybe just drop this line as we now got it to work and also tested the output?

ricardosalveti:
Due the warning issue I commented inline, I'm happy to merge if we drop or change it to bbnote.

meta-lmp-base/classes/lmp.bbclass
mike-sul: Yep, I think it makes sense to drop it now and just rely on the value `USE_FIOTOOLS` (which is printed in the logs) to see which of the tools was used by a specific CI run.

mike-sul: @ricardosalveti I removed `bbwarn`.

ricardosalveti (approved):
LGTM, thanks!

PR #327, opened 2021-04-26
uz3eg-iocc: enable UMS on u-boot

ldts:


ricardosalveti:
@ldts since this causes a bad side effect on the linux side, should we just close the pr?

PR #331, opened 2021-05-10
base: fiotools: switch to the fiotools git repo

mike-sul:
Since it was decided to make the server implementation private we need to move the client tools to their own dedicated git repo, hence we need to update the given recipe correspondingly.


doanac:
i don't want to bikeshed, but I think "fiotools" is *really* generic. This new repo is only going to be an ostree uploader or something like that. We should probably use a more specific name for now in case we some day really need a "fiotools" repo for something else

mike-sul:
> i don't want to bikeshed, but I think "fiotools" is _really_ generic. This new repo is only going to be an ostree uploader or something like that. We should probably use a more specific name for now in case we some day really need a "fiotools" repo for something else

I am open to suggestions. The initial version was called `fiopush` but then I realized that includes fiocheck too and potentially it will incorporate some other useful tools...

doanac:
> I am open to suggestions. The initial version was called fiopush but then I realized that includes fiocheck too and potentially it will incorporate some other useful tools...

We have an "ostreehub", maybe "ostree-uploader" ? I

mike-sul:
> > I am open to suggestions. The initial version was called fiopush but then I realized that includes fiocheck too and potentially it will incorporate some other useful tools...
> 
> We have an "ostreehub", maybe "ostree-uploader" ? I

I am fine with it, @ricardosalveti what do you think?

meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
```diff
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 BRANCH_lmp = "master"
-SRCREV_lmp = "fe447e0c56d05d9d003722a2bbdc9a00205f6a55"
+SRCREV_lmp = "1cfc6e6dbc7aa358effc7334c6f861dc30c2332e"
```
ricardosalveti, line 4: No need to update aktualizr, latest is already in meta-lmp now.

mike-sul: fixed.

PR #336, opened 2021-05-19
base: class: lmp-disable-gplv3: Enable the LMP_DISABLE_GPLV3

angolini:
Turn LMP_DISABLE_GPLV3 on when this class is used.

This way, we can better control along the meta-layer which package or packagegroup is being installed


vanmaegima (approved):

ricardosalveti (approved):
LGTM

PR #347, opened 2021-05-31
Enable support for useradd-staticids and nss-altfiles

ricardosalveti:
The current way we're handling user/group id and their respective /etc/{passwd,group} files is not ideal, as they both can change depending on the order that the packages are installed, causing later a potential mismatch as images are updated via OTA (e.g. ids changed locally will stay the same, even if they are updated on a fresh image).

Support for staticids allows us to better control all the user and group ids that are used by LMP, as the database (used as reference) is maintained in meta-lmp-base.

NSS altfiles allow us to maintain the system user and groups outside /etc, and keep them updated as we push new images via OTA. This is basically what the upstream ostree project suggests to use with /etc/passwd at https://ostreedev.github.io/ostree/adapting-existing/#usrlibpasswd.


Tim-Anderson (approved):
LGTM

mike-scott (approved):
LGTM.  Nice changes

PR #350, opened 2021-06-02
base: init-install-efi: fix compose-apps preloading

ricardosalveti:
Fix compose-apps preloading logic by making sure the older preloaded
content gets removed before copying over the ones available as part of
the EFI installer.


doanac (approved):
i was expecting a mention about *me* causing the problem.

PR #356, opened 2021-06-09
u-boot-ostree-scr-fit: boot-common: force primary path before reset

igoropaniuk:
```
If we are validating secondary boot path, Linux should force reboot back to
primary path after running reboot cmd (this is needed for boards without
software POR reset support, like Apalis iMX6).

```


ricardosalveti (approved):
LGTM, please merge once you are able to test on every supported target (for boot firmware update).

Tim-Anderson (approved):
LGTM

PR #358, opened 2021-06-10
u-boot-ostree-scr-fit: boot-common: import only bootfirmware_version

igoropaniuk:
```
Provide explicitly bootfirmware_version as the only
variable that should be imported from version.txt.

For env import help:
env import [-d] [-t [-r] | -b | -c] addr [size] [var ...]
...

var... List of the names of the only variables that get
       imported from the environment at address 'addr'.
       Without arguments, the whole environment gets imported.

```


Tim-Anderson (approved):
LGTM

ricardosalveti (approved):
LGTM

PR #368, opened 2021-06-15
base: recipes-samples: lmp-image-common: dont crush sysconf_file

mike-scott:
Let's not crush the original ${sysconf_file} and instead append.

Noted on new container builds which use network_mode="host" that
some errored out with localhost connection errors: bad host.

Reproduced with and tested on intel-corei7-64 MACHINE.

Turns:

fio@intel-corei7-64:~$ cat /usr/etc/hosts
127.0.1.1 intel-corei7-64

Into:

fio@intel-corei7-64:~$ cat /etc/hosts
127.0.0.1       localhost.localdomain           localhost

::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
127.0.1.1 intel-corei7-64
127.0.1.1 intel-corei7-64


Tim-Anderson (approved):
LGTM

PR #370, opened 2021-06-15
base: lmp: nss-altfiles: keep standard hosts

ricardosalveti:
Allow hosts to be managed via nss-altfiles, but keep the original
content in place as it is common for containers to bind/reuse /etc/hosts
directly, which is a problem when that gets provided by nss-altfiles.


ricardosalveti:
Replaces https://github.com/foundriesio/meta-lmp/pull/368

Tim-Anderson (approved):
LGTM

PR #381, opened 2021-06-21
xilinx: switch to load FPGA from SPL

ricardosalveti:


meta-lmp-base/classes/uboot-fitimage.bbclass
```diff
@@ -62,12 +67,18 @@ fmt_address() {
 # $2 ... U-Boot load address
 # $3 ... OP-TEE load address
 # $4 ... ATF load address
-# $5 ... bootscr load address
+# $5 ... FPGA load address
```
Tim-Anderson, line 70: for compatibility would it make more sense to have FPGA load address as $6?

Tim-Anderson (commented):
I think in general this is okay. Just had one question in-line

ricardosalveti:
Will merge to get more testing on this.

PR #382, opened 2021-06-23
base: ostree: patch to setup a curl request timeout

mike-sul:
Set a timeout for an overall request processing in order to avoid an
ostree pull/fetch API call or CLI command hang.
The timeout value can be specified via an env variable OSTREE_CURL_TIMEOUT,
if not defined, or equal to zero or it cannot be converted to a long type then
the default value is set.

### DIFF
--- a/meta-lmp-base/recipes-extended/ostree/ostree/0001-ostree-pull-set-request-timeout.patch
+++ b/meta-lmp-base/recipes-extended/ostree/ostree/0001-ostree-pull-set-request-timeout.patch
@@ -1,0 +1,21 @@
+diff --git a/src/libostree/ostree-fetcher-curl.c b/src/libostree/ostree-fetcher-curl.c
+index 0ce3ff00..d50d62e7 100644
+--- a/src/libostree/ostree-fetcher-curl.c
++++ b/src/libostree/ostree-fetcher-curl.c
+@@ -817,6 +817,16 @@ initiate_next_curl_request (FetcherRequest *req,
+   curl_easy_setopt (req->easy, CURLOPT_WRITEDATA, task);
+   curl_easy_setopt (req->easy, CURLOPT_PROGRESSDATA, task);
+ 
++
++  /* set a request timeout, make sure it's not 0, otherwise an overall ostree pull session might hang */
++  long curl_timeout = 0L;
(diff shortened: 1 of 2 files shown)


mike-sul:
@doanac This patch actually fixes the ostree hang issue. The default timeout value is something we can discuss. This is a timeout for a single request within an overall commit fetch/pull operation, which could be lower I suppose, on the other side, we can always override it in the aklite systemd service by setting the env. variable.

meta-lmp-base/recipes-extended/ostree/ostree/0001-ostree-pull-set-request-timeout.patch
```diff
@@ -0,0 +1,21 @@
...
+--- a/src/libostree/ostree-fetcher-curl.c
++++ b/src/libostree/ostree-fetcher-curl.c
+@@ -817,6 +817,16 @@ initiate_next_curl_request (FetcherRequest *req,
+   curl_easy_setopt (req->easy, CURLOPT_WRITEDATA, task);
+   curl_easy_setopt (req->easy, CURLOPT_PROGRESSDATA, task);
+ 
++
++  /* set a request timeout, make sure it's not 0, otherwise an overall ostree pull session might hang */
++  long curl_timeout = 0L;
++  const char* curl_timeout_str = g_getenv ("OSTREE_CURL_TIMEOUT");
```
ricardosalveti, line 12: Do we want to allow this value to be customized via an environment variable or should we just try to force a sane default?

mike-sul: IMO, it might be useful for debugging as well as customization/tuning for a specific customer, we might wanna have a different value for devices with different connectivity types, e.g. LTE vs 1Gb Ethernet.

doanac (commented):
seems fine to me. i'll you two decide the best way for it to be integrated.

meta-lmp-base/recipes-extended/ostree/ostree/0001-ostree-pull-set-request-timeout.patch
ricardosalveti: Sure, seems reasonable.

ricardosalveti (approved):
LGTM

Tim-Anderson (approved):
LGTM

PR #389, opened 2021-07-02
base: recipes-support resize helper needs to be more forceful

Tim-Anderson:
The -p option on e2fsck is not enough to make sure the partition is clean for resizing
so change it to a -y so it will fix.

Then resize may still want e2fsck but since we just did it, we can force resize2fs to
do the resize anyway.


ricardosalveti (approved):
LGTM

PR #390, opened 2021-07-02
Initial implementation for ostree factory reset

ricardosalveti:
On ostree-based systems there are basically 2 locations for rw content, /etc and /var.

As the original /etc (when creating the rootfs) is also stored under /usr/etc on the same deployment (and used during the 3-way merge for updating /etc when applying an update), it is easy to handle as we can simply just erase /etc and restore from what is available at /usr/etc.

As the system is able to manage an empty /var (which is the case during first boot) and recreate all the require directories and files (via systemd tmpfiles.d), we don't need to do much here besides just removing it entirely. While /var can be completely removed, we also have use cases for requiring /var/sota to stay (e.g. not wiping the device registration), so we need to allow the user to decide what to wipe.

As a way to trigger the factory reset, 3 stamp files where created:
 - /var/.factory_reset: sync /etc and wipes /var
 - /var/.factory_reset_keep_sota: sync /etc and wipes everything on /var but /var/sota
 - /var/.factory_reset_keep_sota_docker: sync /etc and wipes everything on /var but /var/sota and /var/lib/docker

TODO/Unknowns:
 - Decide what to do about u-boot env/fiovb
 - What to do when the original image had preloaded containers?

### DIFF
--- a/meta-lmp-base/classes/lmp.bbclass
+++ b/meta-lmp-base/classes/lmp.bbclass
@@ -32,6 +32,8 @@ IMAGE_CMD_ostree_append () {
 
 	# Prefer /usr/lib/tmpfiles.d instead of /etc
 	mv usr/etc/tmpfiles.d/00ostree-tmpfiles.conf usr/lib/tmpfiles.d
+	# Cover missing /var/rootdirs/home (pending in meta-updater)
+	echo "d /var/rootdirs/home 0755 root root -" >> usr/lib/tmpfiles.d/00ostree-tmpfiles.conf
 
 	# Update default home path (compatible with nss alt files)
 	sed -i -e 's,:/home,:/var/rootdirs/home,g' usr/etc/passwd
--- a/meta-lmp-base/recipes-core/images/initramfs-ostree-lmp-image.bb
+++ b/meta-lmp-base/recipes-core/images/initramfs-ostree-lmp-image.bb
@@ -4,6 +4,7 @@ PACKAGE_INSTALL = "initramfs-framework-base \
 	initramfs-module-udev \
 	initramfs-module-rootfs \
 	initramfs-module-ostree \
+	initramfs-module-ostree-factory-reset \
 	${VIRTUAL-RUNTIME_base-utils} \
 	${@bb.utils.contains('DISTRO_FEATURES', 'ima', 'initramfs-framework-ima', '', d)} \
 	udev base-passwd e2fsprogs-e2fsck \
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/ostree_factory_reset
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/ostree_factory_reset
@@ -1,0 +1,74 @@
+#!/bin/sh
+# Copyright (C) 2021 Foundries.io Ltd
+#
+# SPDX-License-Identifier: MIT
+
+OSTREE_DISTRO="lmp"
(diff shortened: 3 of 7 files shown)


mike-sul:
> * What to do when the original image had preloaded containers?

I think it's possible to implement resetting of Apps to the original state that they were at before the first boot.

Of course, the "brute force" approach is to put preloaded Apps and images under `/usr/etc` location (or similar) and then copy them to /var/lib/docker and /var/sota/compose-apps, so resetting means removing these dirs and then doing the copy again.

Another idea is to preload/inject Apps' images into some other location, e.g. /usr/etc or /var/sota/preloaded-images, and then hardlink all files under it to /var/lib/docker. Layers are represented as CAS/by unique folder so the original content under the "preloaded-images" shouldn't be modified during Apps lifetime, except a few files, e.g. `image/overlay2/repositories that can be just copied. Also, the content under `compose-apps` should be copied to /var/sota/compose-apps. It usually doesn't occupy much space thus should be OK to do so.

mike-sul:
> * Should we have an option to keep the docker images in place (and remove volumes and the rest)?

My understanding is that the "reset" is just removing user/device data generated during its lifetime, it doesn't mean rollback to the original version of LmP that the device was originally flashed with? If it is so (no rollback) then it shouldn't be an issue with keeping the latest Target images in place (just remove containers, volumes, etc).
If we actually need to rollback to the original version then it's not clear how to do it, since ostree keeps just last two versions on a device.

doanac:
> My understanding is that the "reset" is just removing user/device data generated during its lifetime, it doesn't mean rollback to the original version of LmP that the device was originally flashed with? If it is so (no rollback) then it shouldn't be an issue with keeping the latest Target images in place (just remove containers, volumes, etc).

We might want to clear docker volumes? 
Might also want to clear files in `/var/sota/compose-apps/<app>/` not in the app tarball. i.e. sometimes people do things in compose like:
~~~
  volumes:
    - ./data:/data
~~~

Tim-Anderson (approved):
This LGTM

ricardosalveti:
> > * Should we have an option to keep the docker images in place (and remove volumes and the rest)?
> 
> My understanding is that the "reset" is just removing user/device data generated during its lifetime, it doesn't mean rollback to the original version of LmP that the device was originally flashed with? If it is so (no rollback) then it shouldn't be an issue with keeping the latest Target images in place (just remove containers, volumes, etc).
> If we actually need to rollback to the original version then it's not clear how to do it, since ostree keeps just last two versions on a device.

Yeah, no need to rollback to a previous version, just making the device fresh from the installed target perspective is fine (kind of a factory state if installed from the same target).

ricardosalveti:
Updated to cover 3 different use cases, which should be enough for the initial implementation:

```
FACTORY_RESET_TAG=".factory_reset"
FACTORY_RESET_KEEP_SOTA_TAG=".factory_reset_keep_sota"
FACTORY_RESET_KEEP_SOTA_DOCKER_TAG=".factory_reset_keep_sota_docker"
```

factory_reset_keep_sota_docker can be used to keep the docker content (assuming the user will know what to clean up when creating the factory reset tag). Nothing really special is done at this stage as the complexity of identifying what needs to be saved or not will get too complicated.

For users wanting to restore a preloaded container that is later removed once registered and running, the recommended way would probably be by delivering the container as part of the rootfs (as then it can be updated together with the rootfs and won't really affect the management of the other containers).

angolini:
I like the new variables! Much better @ricardosalveti

angolini (approved):

PR #391, opened 2021-07-13
base: ostree: increase default curl request timeout

mike-sul:
Increase the default curl request timeout up to 13 minutes. We cannot set higher than ~ 15m because of this issue https://github.com/ostreedev/ostree/issues/2383.
It can be critical in the case of big files pulling through the network with limited bandwidth. libostree doesn't do a resumable download of a single file, so if a timeout occurs then the subsequent file download retry starts from the beginning, which might lead to endless retries...
This change slightly improves chances to succeed. In addition to this, I am going to limit the maximum allowed file size in the rootfs that LmP build generates (currently it's set to 4GB).


mike-sul:
@mwasilew FYI.

ricardosalveti (approved):
LGTM

doanac (approved):

PR #396, opened 2021-07-14
base: u-boot-tools: fix a mkimage signature issue

igoropaniuk:
```
Backport patch from [1], which fixes FIT signing issues. Details:
A following error was observed:
| Can't write signature for 'signature@1' signature node in
| 'conf@imx6ull-colibri-wifi-eval-v3.dtb' conf node: <unknown error>
| uboot-mkimage Can't add hashes to FIT blob: -1

This is caused by a wrong return value being used in uboot source.

The return value '-ENOSPC' of fit_set_timestamp function does not match
the caller fit_image_write_sig's expection which is '-FDT_ERR_NOSPACE'.

Fix it by not calling fit_set_timestamp, but call fdt_setprop instead.

[1] http://git.yoctoproject.org/cgit.cgi/poky/commit/?id=e9fa8f9001471bb240e997cb4537d2f617f22957
```

### DIFF
--- a/meta-lmp-base/recipes-bsp/u-boot/u-boot-tools_%.bbappend
+++ b/meta-lmp-base/recipes-bsp/u-boot/u-boot-tools_%.bbappend
@@ -1,0 +1,3 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+SRC_URI += "file://0001-tools-image-host-fix-wrong-return-value.patch"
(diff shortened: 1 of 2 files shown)


igoropaniuk:
@mike-scott @ricardosalveti could you please take a look?

meta-lmp-base/recipes-bsp/u-boot/u-boot-tools_%.bbappend
```diff
@@ -0,0 +1,3 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+SRC_URI_append = "file://0001-tools-image-host-fix-wrong-return-value.patch"
```
ricardosalveti, line 3: Either add an extra space before file (when using append) or simply just use SRC_URI += "file..." here.

Append does a direct string append, and it is safer to use it with an extra space.

igoropaniuk: fixed, thanks!

ricardosalveti (approved):
LGTM

PR #400, opened 2021-07-15
base: wayland: weston-init: Add $OPTARGS and enable xwayland

munoz0raul:
1 - Some devices do not launch weston without the input device connected.
    Looks like the configuration on weston.ini [core] require-input=bool
    is not the same as --continue-without-input.
    According to this issue, it will be fixed in the next version:
    https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/524

2 - meta-freescale implements a function to uncomment the #xwayland=true on weston.ini.
    To trigger that function, we need INI_UNCOMMENT_ASSIGNMENTS for non-imx devices.
    https://github.com/Freescale/meta-freescale/blob/master/recipes-graphics/wayland/weston-init.bbappend#L11

### DIFF
--- a/meta-lmp-base/recipes-graphics/wayland/weston-init.bbappend
+++ b/meta-lmp-base/recipes-graphics/wayland/weston-init.bbappend
@@ -3,10 +3,16 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 SRC_URI_append_lmp-wayland = " \
     file://utilities-terminal.png \
     file://background.jpg \
+    file://weston.env \
+    file://weston.service.patch \
 "
 
 FILES_${PN}_append_lmp-wayland = " \
     ${datadir}/weston \
+"
+
+INI_UNCOMMENT_ASSIGNMENTS = " \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'x11 wayland', 'xwayland=true', '', d)} \
 "
 
 do_install_append_lmp-wayland() {
@@ -15,4 +21,8 @@ do_install_append_lmp-wayland() {
 
     install -m 0644 ${WORKDIR}/utilities-terminal.png ${D}${datadir}/weston/icon/utilities-terminal.png
     install -m 0644 ${WORKDIR}/background.jpg ${D}${datadir}/weston/backgrounds/background.jpg
+
+    for assignment in ${INI_UNCOMMENT_ASSIGNMENTS}; do
+        uncomment "$assignment" ${D}${sysconfdir}/xdg/weston/weston.ini
+    done
 }
--- a/meta-lmp-base/recipes-graphics/wayland/weston-init/lmp-wayland/weston.env
+++ b/meta-lmp-base/recipes-graphics/wayland/weston-init/lmp-wayland/weston.env
@@ -1,0 +1,1 @@
+OPTARGS=--continue-without-input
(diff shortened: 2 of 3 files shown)


meta-lmp-base/recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -15,4 +17,9 @@ do_install_append_lmp-wayland() {
 
     install -m 0644 ${WORKDIR}/utilities-terminal.png ${D}${datadir}/weston/icon/utilities-terminal.png
     install -m 0644 ${WORKDIR}/background.jpg ${D}${datadir}/weston/backgrounds/background.jpg
+    install -Dm 0644 ${WORKDIR}/weston.conf ${D}${sysconfdir}/default/weston
```
ricardosalveti, line 20: There is already a weston.env file doing exactly what you need here, so instead of creating a new file and installing it here, just create a custom weston.env instead (which will then be used by the main recipe).

```
meta/recipes-graphics/wayland/weston-init.bb:           file://weston.env \
meta/recipes-graphics/wayland/weston-init.bb:   install -Dm644 ${WORKDIR}/weston.env ${D}${sysconfdir}/default/weston
```

meta-lmp-base/recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -15,4 +17,9 @@ do_install_append_lmp-wayland() {
...
 }
+
+INI_UNCOMMENT_ASSIGNMENTS = " \
```
ricardosalveti, line 23: This is a problem as this function is only available in meta-freescale, which can be removed by our customers (not every bsp layer is forced to be included, they are just by default).

```
meta-freescale/recipes-graphics/wayland/weston-init.bbappend

do_install_append() {
    for assignment in ${INI_UNCOMMENT_ASSIGNMENTS}; do
        uncomment "$assignment" ${D}${sysconfdir}/xdg/weston/weston.ini
    done
}
```

For having this generic to LmP we would need to have a similar function declared at meta-lmp-base, just do it in a way that still works correctly when meta-freescale is enabled and used.

meta-lmp-base/recipes-graphics/wayland/weston-init.bbappend
```diff
@@ -3,16 +3,26 @@ FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
...
 do_install_append_lmp-wayland() {
     install -d ${D}${datadir}/weston/backgrounds
     install -d ${D}${datadir}/weston/icon
 
     install -m 0644 ${WORKDIR}/utilities-terminal.png ${D}${datadir}/weston/icon/utilities-terminal.png
     install -m 0644 ${WORKDIR}/background.jpg ${D}${datadir}/weston/backgrounds/background.jpg
-}
+
+    for assignment in ${INI_UNCOMMENT_ASSIGNMENTS}; do
+        uncomment "$assignment" ${D}${sysconfdir}/xdg/weston/weston.ini
```
ricardosalveti, line 26: The function uncomment is only defined in the bbappend from meta-freescale.

To test this, try disabling the meta-freescale layer when generating the image.

As we need the OPTARGS changes for getting rpi (and other machines to work), mind splitting this into 2 patches, one for OPTARGS and another for INI_UNCOMMENT_ASSIGNMENTS?

PR #401, opened 2021-07-15
base: wayland: weston: Installing weston-screenshooter

munoz0raul:
To add the possibility of capture a screenshot with (super + s),
    weston-screenshooter needed to be installed.

### DIFF
--- a/meta-lmp-base/recipes-graphics/wayland/weston_%.bbappend
+++ b/meta-lmp-base/recipes-graphics/wayland/weston_%.bbappend
@@ -1,0 +1,1 @@
+FILES_${PN} += " ${bindir}/weston-screenshooter "


ricardosalveti:
This one is adding a commit from a previous PR, need to clean it up.

munoz0raul:
Could you check if now it is correct?

ricardosalveti:
This one might not be needed as it is already packaged at weston-examples, so just make sure you're installing that package.

`./packages-split/weston-examples/usr/bin/weston-screenshooter`

munoz0raul:
Should it be added as default in the lmp Wayland distro? Or should we allow users to choose if it is really necessary?

ricardosalveti:
> Should it be added as default in the lmp Wayland distro? Or should we allow users to choose if it is really necessary?

I would prefer the user to add when needed, since this tool is quite specific to a particular use case.

PR #404, opened 2021-07-27
base: wayland: weston-init: Add $OPTARGS

munoz0raul:
- Some devices do not launch weston without the input device connected.
  Looks like the configuration on weston.ini [core] require-input=bool
  is not the same as --continue-without-input.
  According to this issue, it will be fixed in the next version:
  https://gitlab.freedesktop.org/wayland/weston/-/merge_requests/524

Signed-off-by: Raul Muñoz <raul@foundries.io>


ricardosalveti (approved):
LGTM

PR #405, opened 2021-07-27
base: wayland: weston-init: Uncomment Assignments

munoz0raul:
meta-freescale implements a function to uncomment assignments like

This function is just available if meta-freescale is included.
The commit will duplicate the function and add the set
INI_UNCOMMENT_ASSIGNMENTS for x11 wayland distro regardless of
the board.

Tested with and without meta-freescale.

Signed-off-by: Raul Muñoz <raul@foundries.io>


ricardosalveti (approved):
LGTM

PR #408, opened 2021-07-30
base: lmp: add support for ETC_GROUP_MEMBERS

ricardosalveti:
Currently a normal user is unable to change the group list for a user
based on normal tooling (e.g. useradd/usermod), as the main group file
is provided by nss-altfiles (ro - /usr/lib/group) and not by /etc/group.

Follow the same solution as done in rpm-ostree and define a list of
groups that should be mirrored from /usr/lib/group to /etc/group when
creating the image. This should be used for groups that are commonly
changed by the user at runtime (e.g. when adding a new user) such as
audio, sudo and video. Build time groups and users can stay at
nss-altfiles (getgrent will do a merge based on what is configured in
nsswitch.conf).

Reference change from rpm-ostree:
https://github.com/coreos/rpm-ostree/commit/9a2007389337c61c152ea6e89575436e4fe5b1c0


angolini (approved):

Tim-Anderson (approved):
LGTM

PR #423, opened 2021-09-03
Migrating am64xx evm to scr-fit

ricardosalveti:


Tim-Anderson (approved):
After finally getting to the bottom of the failure, i.e. changing SD cards, this patch set works correctly. LGTM

PR #425, opened 2021-09-13
bsp: u-boot-xlnx: deploy u-boot config used by the build

igoropaniuk:
```
Deploy the u-boot configuration that was used during the build
process as it's done by default now for u-boot-fio,
as it might be useful as reference later.

```

### DIFF
--- a/meta-lmp-base/recipes-bsp/u-boot/u-boot-fio-common.inc
+++ b/meta-lmp-base/recipes-bsp/u-boot/u-boot-fio-common.inc
@@ -21,6 +21,7 @@ B = "${WORKDIR}/build"
 do_configure[cleandirs] = "${B}"
 
 require recipes-bsp/u-boot/u-boot.inc
+require recipes-bsp/u-boot/u-boot-lmp-common.inc
 
 # Support additional u-boot classes such as u-boot-fitimage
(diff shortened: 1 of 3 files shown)


igoropaniuk:
@ricardosalveti this is just the same as:
```
ommit 37b99ff14d2f6073a40b053d2421b0fa2ec8bded
Author: Ricardo Salveti <ricardo@foundries.io>
Date:   Wed Oct 28 22:44:35 2020 -0300

    base: u-boot-fio-common: deploy u-boot config used by the build
    
    Also deploy the u-boot configuration that was used during the build
    process, as it might be useful as reference later on (as it gets stored
    by our CI service).
    
```

igoropaniuk:
as an option we can move that to the common include (and inlude it in both u-boot-fio and u-boot-xilinx), to avoid code duplication

ricardosalveti:
> as an option we can move that to the common include (and inlude it in both u-boot-fio and u-boot-xilinx), to avoid code duplication

Yeah, that would be better, because we also got other u-boot recipes that could also use such include file.

igoropaniuk:
@ricardosalveti done

ricardosalveti (approved):
LGTM

PR #426, opened 2021-09-14
Update main kernel to 5.14, and update LTS to track 5.10 instead of 5.4.

ricardosalveti:


ricardosalveti:
@MrCry0 I updated our main kernel branch (and respective tracking branches), removing most patches we had on 5.4, simplifying our upstream tracking process for it.

We now have https://github.com/foundriesio/linux/commits/tracking-fio-arm-5.14, https://github.com/foundriesio/linux/commits/tracking-fio-misc-5.14 and https://github.com/foundriesio/linux/commits/tracking-fio-imx-5.14:
 * tracking-fio-arm-5.14: backports for isp1760 (all already in 5.15) + quirk for n1sdp that upstream already refused, so nothing to be done from our side
 * tracking-fio-misc-5.14: we need to align with @ldts here, but the one patch we have is to leverage optee (the TA we also track in our fork) for providing RNG. I know there was an effort in optee to make a generic TA for RNG, we just need to do a follow up to understand if something we can drop as we update our optee as well.
 * tracking-fio-imx-5.14: please also align with @ldts on the one patch we have, I don't remember exactly why we needed the caam clocks to be stay enabled.

vanmaegima (approved):

MrCry0 (approved):
LGTM

Tim-Anderson (approved):
LGTM

PR #430, opened 2021-09-23
recipes-sota: Bump aktualizr-lite

doanac:
This brings in a new version of aktualizr-lite that can be built as
an api and shared library for custom clients to extend.

Other changes include:
 * 68325ced - Fix hwinfo reporting on x86

### DIFF
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
 BRANCH_lmp = "master"
-SRCREV_lmp = "a40a69b8dae0b4fcf7360c6d132ed6d294016864"
+SRCREV_lmp = "5e2f6da2c1ad02ea87fa32564d9b37d9771a4ec3"
 
 SRC_URI_lmp = "gitsm://github.com/foundriesio/aktualizr-lite;branch=${BRANCH};name=aktualizr \
     file://aktualizr.service \
@@ -53,7 +53,7 @@ do_install_append_lmp() {
     install -m 0644 ${WORKDIR}/tmpfiles.conf ${D}${nonarch_libdir}/tmpfiles.d/aktualizr-lite.conf
 }
 
-PACKAGES += "${PN}-get ${PN}-lite"
+PACKAGES += "${PN}-get ${PN}-lite ${PN}-lite-lib"
 FILES_${PN}-get = "${bindir}/aktualizr-get"
 FILES_${PN}-lite = "${bindir}/aktualizr-lite"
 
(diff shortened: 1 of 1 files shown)


meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
```diff
@@ -53,11 +53,14 @@ do_install_append_lmp() {
...
 FILES_${PN}-get = "${bindir}/aktualizr-get"
 FILES_${PN}-lite = "${bindir}/aktualizr-lite"
 
 # Force same RDEPENDS, packageconfig rdepends common to both
 RDEPENDS_${PN}-lite = "${RDEPENDS_aktualizr}"
 
 FILES_${PN}-lite += "${nonarch_libdir}/tmpfiles.d/aktualizr-lite.conf"
+
+FILES_${PN}-lite-lib = "${libdir}/libaktualizr_lite.so"
+FILES_${PN}-dev += "${includedir}/${PN}-lite ${includedir}/json"
```
ricardosalveti, line 66: What exactly are we installing under ${includedir}/json? Just curious because this is a generic path that has a potential to conflict with other packages.

doanac: i was worried this might be a problem. aktualizr pulls in a third-party dependency as a git submodule:

 https://github.com/open-source-parsers/jsoncpp/tree/6aba23f4a8628d599a9ef7fa4811c4ff6e4070e2/include/json

ricardosalveti: Hm, there is a recipe for that in OE already, do we want to use the version available as a submodule or should we just handle it as an external dependency and build using the jsoncpp recipe instead?

ricardosalveti: Seems the upstream project is not doing that many releases lately, so not sure if we should be concerned with version mismatch, but will let that one for you as you are consuming it.

From the OE perspective I would just add jsoncpp as a build dependency in aktualizr-lite if used by the project, otherwise I would not install anything here and let the client have jsoncpp as a build dependency, as this is just another generic third-party dependency.

mike-sul: To some extend this case is similar to aklite dependency on dockerd, docker and docker-compose. Test and dev flow uses the utilities installed in the test container and a dev local/container host and a version of these utilities doesn't always match the version defined in LmP. Thus, the final testing happens on actual device/VM, so it could be the same for jsoncpp dependency.

doanac: i've force pushed a change that drops the HEAD commit in aklite: https://github.com/foundriesio/aktualizr-lite/commit/87fde29d993f457fa304c4467132d698d6dac523 

I think we should just revert that change. I've built my custom client with a DEPENDS on jsoncpp and things work fine

ricardosalveti: Cool, then better to just drop the jsoncpp piece.

doanac: oops - i forgot to git-push yesterday. fixed now.

mike-sul (approved):
I guess this commit should be removed https://github.com/foundriesio/aktualizr-lite/commit/87fde29d993f457fa304c4467132d698d6dac523

doanac:
> I guess this commit should be removed foundriesio/aktualizr-lite@87fde29

Yep, i'm going to do a git-revert PR for that in a bit.

ricardosalveti (approved):
LGTM

PR #439, opened 2021-10-05
cleanups for imx-boot and optee-os-fio

mike-scott:
- Change missed TEMP reference to TODO
- Consolidate SoC entries for mx8qm


ricardosalveti (approved):
LGTM

PR #445, opened 2021-10-21
base: kernel-lmp-fitimage: fix loadable and dtb counters

mike-scott:
In fitimage_assemble, we use a shell variable "i" to keep track
of the number of DTBs we've added.  This variable is compared
to "dtbcount" in order to see if we're on the first or "default"
dtb.

However, we also use the "i" variable in fitimage_emit_section_config
to keep a running count of loadables so that we can name them
"loadable_${i}" for uniqueness.

This means anytime "FIT_LOADABLES" is set, we are messing with the DTB
counter and it causes more than 1 "default" node to possibly be set.

Fix this by not using the "i" variable in fitimage_emit_section_config.


Tim-Anderson (approved):
LGTM

ricardosalveti (approved):
LGTM

PR #446, opened 2021-10-21
fix single char vars

MrCry0:
Rename single-char variables to make code easy to understand and avoid possible name conflicts.


Tim-Anderson (approved):
LGTM

PR #456, opened 2021-11-04
base: fioconfig: reuse OE ldflags and enable shared

ricardosalveti:
Let OE handle the default ldflags used by the build and also allow
shared support as an effort to reduce the binary size.

### DIFF
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
@@ -17,8 +17,8 @@ UPSTREAM_CHECK_COMMITS = "1"
 
 inherit go-mod systemd
 
-GOBUILDFLAGS += '-tags vpn -ldflags="-X ${GO_IMPORT}/internal.Commit=${SRCREV}"'
-GO_LINKSHARED = ""
+# Extend from go.bbclass to include internal.Commit and tags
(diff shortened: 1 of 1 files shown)


meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
```diff
@@ -17,8 +17,7 @@ UPSTREAM_CHECK_COMMITS = "1"
 
 inherit go-mod systemd
 
-GOBUILDFLAGS += '-tags vpn -ldflags="-X ${GO_IMPORT}/internal.Commit=${SRCREV}"'
-GO_LINKSHARED = ""
+GO_LDFLAGS += '-tags vpn -ldflags="-X ${GO_IMPORT}/internal.Commit=${SRCREV}"'
```
doanac, line 20: nice.

doanac (approved):

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
ricardosalveti: Had to change in the end as that was resulting in 2 ldflags at the same build line, and which didn't work right.

Same thing now, but we're setting the variable based on the original value, instead of just appending (there is no common build flag yet that can be used for recipes to customize just one small piece).

PR #458, opened 2021-11-11
base: containerd: move back to the 1.4.x series

ricardosalveti:
Docker upstream is still aligned with the 1.4.x series, so prefer that
instead of the 1.5.x one to avoid issues caused by the mismatch.

This change moves it back to the current latest stable release of the
1.4 branch, 1.4.11.


mike-scott (approved):
LGTM

PR #472, opened 2021-11-29
ebbr-nxp: bsp: u-boot: apalis-imx6-sec: enable misc uclass device

MrCry0:
The u-boot class device "misc" now disabled by default but it is
required by crypto device. Enable it explicitly to avoid fatal
error [1].

[1]
  Cannot find uclass for id 54: please add the UCLASS_DRIVER() declaration for this UCLASS_... id
  Missing uclass for driver caam_jr
  Error binding driver 'caam_jr': -96


vanmaegima (approved):
LGTM

PR #482, opened 2021-12-03
Use freescale 5.10-2.1.x-imx linux kernel instead of nxp lf-5.10.y

MrCry0:
The 5.10-2.1.x-imx includes fixes from the mainline linux-stable and atm seems compatible with our patches and configs for lf-5.10.y.


vanmaegima (approved):
LGTM

PR #483, opened 2021-12-06
master: add wi-fi firmware for imx8mm evkb

MrCry0:
Can be merged separately. Wi-Fi starts working after merging https://github.com/foundriesio/lmp-kernel-cache/pull/25

PR #484, opened 2021-12-06
ebbr-nxp: add wi-fi firmware for imx8mm evkb #483

MrCry0:
Can be merged separately. Wi-Fi starts working after merging https://github.com/foundriesio/lmp-kernel-cache/pull/24

### DIFF
--- a/meta-lmp-base/recipes-samples/images/lmp-feature-wifi.inc
+++ b/meta-lmp-base/recipes-samples/images/lmp-feature-wifi.inc
@@ -1,4 +1,6 @@
 # Wifi packages
+MACHINE_FIRMWARE_append = " ${@bb.utils.contains('MACHINE_FEATURES', 'nxp8987', 'linux-firmware-nxp89xx', '', d)}"
+
 CORE_IMAGE_BASE_INSTALL += " \
     hostapd \
     linux-firmware-ar3k \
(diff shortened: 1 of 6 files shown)


vanmaegima (approved):
LGTM

ricardosalveti:
I wonder if we should instead just import https://source.codeaurora.org/external/imx/meta-imx/tree/meta-bsp/recipes-kernel/linux-firmware/linux-firmware_%25.bbappend?h=hardknott-5.10.52-2.1.0, which seems to be adding all the relevant missing firmware.

Since this might not necessarily apply to every machine, we can make the appends be specific to imx if needed.

MrCry0:
> I wonder if we should instead just import https://source.codeaurora.org/external/imx/meta-imx/tree/meta-bsp/recipes-kernel/linux-firmware/linux-firmware_%25.bbappend?h=hardknott-5.10.52-2.1.0

@ricardosalveti 
But HOW? :) I've been searching for hours for that firmware's recipe through all imx repos and the whole Google :)
Well, so now it is better to use that recipe, keep an original firmware name, and add a kernel patch to alternate a firmware name in a mwifiex driver.
Do you mind?

ricardosalveti:
We can copy paste that recipe and just make the append specific to imx (imx is a generic override for all imx machines), simply to avoid replacing firmware files that could break other machines.

But if the files are not replacing anything, we should be ok to have it as done at the current recipe.

Tim-Anderson (approved):
LGTM

PR #487, opened 2021-12-09
lmp_base: force NM to ignore docker interfaces

mwasilew:
Running NetworkManager CLI and attempting to turn networking off affects
docker bridge interfaces. NM turns them off and is not able to turn them
back on. This patch tells NM to ignore all docker* and br-* interfaces.
It might affect other bridges and should be used with caution.

Signed-off-by: Milosz Wasilewski <milosz.wasilewski@foundries.io>

### DIFF
--- a/meta-lmp-base/recipes-connectivity/networkmanager/networkmanager/20-docker0.conf
+++ b/meta-lmp-base/recipes-connectivity/networkmanager/networkmanager/20-docker0.conf
@@ -1,0 +1,2 @@
+[keyfile]
+unmanaged-devices+=interface-name:docker*,interface-name:br-*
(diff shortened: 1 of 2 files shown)


ricardosalveti:
@mwasilew can you try https://github.com/ricardosalveti/meta-lmp/commit/c88e68e5ca0229201843b78452574301ddb38cba instead? That only covers the bridges created by NM.

With this patch the bridge devices created by docker are now unmanaged.

mwasilew:
@rsalveti your patch works well. Should I drop this PR?

ricardosalveti:
> @rsalveti your patch works well. Should I drop this PR?

Thanks for confirming, yeah, let's use mine instead as it is specific only to the bridges created and managed by docker.

Will push to master.

PR #492, opened 2021-12-14
base: plug-and-trust-seteec: bump revision

ldts:
The updated release restores the P&T check for the SE applet version to match the one
selected at build time.


ricardosalveti (approved):
LGTM

PR #496, opened 2021-12-22
base: recipes-core: dump os-release to CI artifact

mike-sul:
Copy the os-release data to a CI build artifact/file.

### DIFF
--- a/meta-lmp-base/recipes-core/os-release/os-release.bbappend
+++ b/meta-lmp-base/recipes-core/os-release/os-release.bbappend
@@ -9,3 +9,12 @@ SUPPORT_URL = "https://support.foundries.io/"
 LMP_MACHINE = "${MACHINE}"
 LMP_FACTORY = "${LMP_DEVICE_FACTORY}"
 LMP_FACTORY_TAG = "${LMP_DEVICE_REGISTER_TAG}"
+
+inherit deploy
+
+do_deploy () {
+    install -d ${DEPLOYDIR}
+    install -m 0644 os-release ${DEPLOYDIR}
+}
+
+addtask do_deploy after do_install


doanac (commented):
i'll defer to ricardo.

meta-lmp-base/recipes-core/os-release/os-release.bbappend
```diff
@@ -9,3 +9,9 @@ SUPPORT_URL = "https://support.foundries.io/"
 LMP_MACHINE = "${MACHINE}"
 LMP_FACTORY = "${LMP_DEVICE_FACTORY}"
 LMP_FACTORY_TAG = "${LMP_DEVICE_REGISTER_TAG}"
+
+do_install_append () {
```
ricardosalveti, line 13: Here we should use do_deploy instead, as the main idea is to deploy the same file that gets installed (which is the main task used for building the final recipe-based package).

meta-lmp-base/recipes-core/os-release/os-release.bbappend
```diff
@@ -9,3 +9,9 @@ SUPPORT_URL = "https://support.foundries.io/"
...
+    if [ -n "${ARCHIVE_DIR}" ] && [ -d "${ARCHIVE_DIR}" ]; then
+        cat os-release > "${ARCHIVE_DIR}"/os-release
```
ricardosalveti, line 15: I find a bit weird that we're using ARCHIVE_DIR directly, which is only known to our CI, and I see this same logic as something that can be useful on normal builds.

Would it work if we just deploy at the normal OE deploy folder (${DEPLOYDIR}) and change ci-scripts to move it to archive_dir as done for the normal build artifacts?

mike-sul: I guess you mean this https://github.com/foundriesio/ci-scripts/blob/3a09a1c3e248d89cd4c9e9586bfbe30b584c12cb/lmp/build.sh#L50?

I'll try, I agree that it makes more sense.

meta-lmp-base/recipes-core/os-release/os-release.bbappend
mike-sul: `do_deploy` or `do_deploy_append` since we are inheriting the original recipe?

ricardosalveti: There is no do_deploy on the original recipe, so either would work.

But to be safe I think doing append is better, in case later on the original recipe starts having its own function.

meta-lmp-base/recipes-core/os-release/os-release.bbappend
ricardosalveti: Inside OE the variable that should be used for deployed content is DEPLOYDIR.

ricardosalveti: e.g.

```
do_deploy() {
    install -d ${DEPLOYDIR}
    install -m 0644 boot.scr ${DEPLOYDIR}/boot.scr-${MACHINE}-${PV}-${PR}
    ln -sf boot.scr-${MACHINE}-${PV}-${PR} ${DEPLOYDIR}/boot.scr-${MACHINE}
    ln -sf boot.scr-${MACHINE}-${PV}-${PR} ${DEPLOYDIR}/boot.scr
    install -m 0644 uEnv.txt ${DEPLOYDIR}
}
```

mike-sul:
Corresponding changes to ci-scripts https://github.com/foundriesio/ci-scripts/pull/168.
Test CI build https://ci.foundries.io/projects/msul-dev01/lmp/builds/1180/raspberrypi3-64/.
@ricardosalveti just `do_deploy_append` doesn't work, `do_deploy` is required.

PR #498, opened 2021-12-30
bsp: linux-lmp-fslc-imx: don't enable GPIO_VF610

MrCry0:
This driver has enabled for all MXC SoCs, whereas it is required by
VF610 and i.MX7ULP only. Apply the patch to enable the driver for
proper SoCs only.

It fixes building a kernel 5.10 after configs optimization.


ldts (commented):
cant we just use the same Kconfig as used in mainline?
https://github.com/torvalds/linux/blob/master/drivers/gpio/Kconfig#L665

MrCry0:
@ldts this bug is introduced in the NXP branch only. (5bb96ccd5bab "LF-2590 gpio: vf610: remove the dependency on SOC_VF610")
So we need to upstream the fix to NXP, not to mainline.

Supporting imx7ulp in mainline is out of the scope of this task.

ldts:
> @ldts this bug is introduced in the NXP branch only. (5bb96ccd5bab "LF-2590 gpio: vf610: remove the dependency on SOC_VF610") So we need to upstream the fix to NXP, not to mainline.
> 
> Supporting imx7ulp in mainline is out of the scope of this task.

um, AFAICS imx7ulp is supported in mainline.
so why not just add the missing code to the config in mainline and then carry a more generic patch (aligned with mainline). 

My point here anyway is to have the review upstream first (whether mainline or the NXP branch...). but perhaps you posted to NXP already?

MrCry0:
@ldts sure, it makes sense.

MrCry0:
@ldts https://github.com/Freescale/linux-fslc/pull/511
I'm marking the current PR as a draft so far.

ricardosalveti:
@MrCry0 thanks for pushing this directly to the community maintained tree.

We need to find the best way to get this to NXP, but we can follow that up separately.

MrCry0:
@ricardosalveti I remember George talking about such way but can't find details. Do you remember that?

MrCry0:
@rsalveti @ldts anyway, the patch is already merged to linux-fslc so I'm dropping this PR and bumping the linux kernel version instead.

PR #499, opened 2022-01-04
linux-yocto: fixup virtualization include path

jeromebrunet:
This append is copied from meta-virtualization but now it is out of
the path is incomplete and the linux-yocto kernel virtualization are
silently dropped.

Correct the path and get the virtualization working again

Signed-off-by: Jerome Brunet <jbrunet@baylibre.com>


ricardosalveti (approved):
LGTM, thanks for the fix!

PR #505, opened 2022-01-08
Master: fix imx-atf

MrCry0:
since imx-atf is updated in meta-freescale, these commits are needed to make master good again.


ricardosalveti (approved):
LGTM

PR #518, opened 2022-01-20
u-boot-xlnx: update to 2022.01 

ricardosalveti:


igoropaniuk (approved):
Run boot smoke test + also tested `reset` cmd, all ok,
Thanks for fixing that boot issue!

PR #525, opened 2022-01-26
base: systemd: define default packageconfig via bbappend

ricardosalveti:
To allow the user to enable support for networkd, we can't really use
the _removal syntax because that always gets parsed last, so instead
define a default set of packageconfig entries used and required by the
default LmP configuration via bbappends.

This is heavily based on the oe-core systemd recipe, and will have to be
maintained and updated as the main recipe adds/removes new options.

### DIFF
--- a/meta-lmp-base/recipes-core/systemd/systemd_%.bbappend
+++ b/meta-lmp-base/recipes-core/systemd/systemd_%.bbappend
@@ -1,5 +1,43 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
+# Based on the original recipe but changed for LmP (done to avoid changing via removal)
+## NOTE: This list will have to be reviewed / updated on every systemd recipe update from OE-Core
+PACKAGECONFIG ?= " \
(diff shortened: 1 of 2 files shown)


angolini (approved):

Tim-Anderson (approved):
LGTM

vanmaegima (approved):

meta-lmp-base/recipes-core/systemd/systemd_%.bbappend
```diff
@@ -1,5 +1,42 @@
 FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
 
+# Based on the original recipe but changed for LmP (done to avoid changing via removal)
```
MrCry0, line 3: I'd add there a reminder for this variable having to be maintained and updated from the main recipe.

ricardosalveti: Done.

MrCry0 (approved):
LGTM

PR #532, opened 2022-02-08
bsp: linux-lmp-dev-mfgtool: drop custom patch for apalis-imx6

ricardosalveti:
We are now using the NXP BSP by default on mfgtool, so drop patch as it
is already part of the BSP tree.


MrCry0 (approved):
LGTM

PR #533, opened 2022-02-09
base: lmp-passwd/group: add systemd journal-gateway and remote user/group

ricardosalveti:
Both can be required depending on the packageconfig used by the systemd
recipe.


angolini (approved):

mike-scott (approved):
LGTM

PR #534, opened 2022-02-10
docker-py: Fix configuration import logic

doanac:
The proper logic as done by our golang /usr/bin/docker is to load the
system config if present and then *extend* that with any user defined
config files. The docker-py logic did not allow for that.


ricardosalveti (approved):
LGTM, that was a wild one.

PR #538, opened 2022-02-14
bsp: imx-boot: change DDR DMEM firmware padding

vanmaegima:
Change padding of DMEM DDR firmware to save space in internal RAM.
DMEM is padded to 16KB but the actual size of this fw is less than
4KB, so padding to 4KB is enough.

imx-mkimage and U-Boot should be aligned in this change.


vanmaegima:
Still need the u-boot bump after https://github.com/foundriesio/u-boot/pull/136

Tim-Anderson (approved):
LGTM

PR #542, opened 2022-02-15
bsp: u-boot-fio-mfgtool: imx8qm-mek: add support for TEE/OP-TEE

igoropaniuk:
This is needed to force HAS_CAAM/FSL_CAAM to be unset in U-Boot,
as in current setup CAAM is handled by OP-TEE OS core.


ricardosalveti (approved):
LGTM

PR #546, opened 2022-02-16
bsp: optee-os-fio: apalis-imx6: fix system lockup

MrCry0:
Due to lockups when using HWRNG via OP-TEE, disable CFG_PTA_RNG for
now.


MrCry0:
Not a solution actually.

PR #548, opened 2022-02-17
V85 master backports to hardknott

ricardosalveti:


ricardosalveti:
With these commits master (v85) is in sync with the hardknott branch (no diff).

PR #549, opened 2022-02-18
base: u-boot-ostree-scr-fit: add missing 'echo' command

vanmaegima:
If an image is not found during 'load_image', the board reports:
Unknown command 'FIO: error occured while loading ...'

Fix this by adding the 'echo' command before the string.


ricardosalveti:
Patch subject should use 'base' instead of 'bsp'.

angolini (approved):

mike-scott:
Good with the fix once patch title is fixed.

PR #554, opened 2022-03-08
Add recovery service for compose-apps-early-start

ricardosalveti:


doanac (approved):
i think this should help. i'll keep poking some at dockerd, but we'll always need this as a last-result fail-safe.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
```diff
@@ -1,5 +1,22 @@
 #!/bin/sh -e
 
+function start_compose_app() {
+	app=$1
+	cd /var/sota/compose-apps/${app}
+
+	# Temporarily disable e to catch possible issues with the initial up
+	set +e
```
MrCry0, line 8: I'd recommend using the following trick to handle RC under `set -e`

```
echo "Starting app ${app}"
ret=0
docker-compose up -d || ret=$?
```
This method doesn't need to dis/en e

mike-sul (commented):
LGTM.

Is the `compose-apps-early-start` is executed at each boot? If so, shouldn't it be deactivated once it's run once, so `dockerd` starts containers by itself at each boot after that?

angolini (approved):

ricardosalveti:
> Is the `compose-apps-early-start` is executed at each boot? If so, shouldn't it be deactivated once it's run once, so `dockerd` starts containers by itself at each boot after that?

In this case it is actually a good thing it gets executed at each boot, because otherwise we would not be able to restore the system to a working stage.

And it is not a problem since docker will see that the container is already up and move forward with the boot.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
ricardosalveti: Nice, I knew there was a better way but forgot when doing this.

MrCry0 (approved):
LGTM

PR #562, opened 2022-03-21
lmp: default image format should really be a default

jeromebrunet:
Forcing the IMAGE_FSTYPES is bad because it overrides the BSP actual
default values, which is more likely to be appropriate for the device than
the one provides by lmp.

Doing so forces the BSP adaptation to IMAGE_FSTYPES:append to restore a
sensible default. This brings another problem with initramfs which tries to
reset the IMAGE_FSTYPES for a good reason. Because of the append, another
image type is added to the initramfs, when all we added is probably a cpio.

A good example of that in meta-lmp-bsp is tegra which needs "tegraflash"
for the actual image and because it is added through an append, it needs to
removed from the initramfs. Such approach, while kinda working, really
don't scale well. It would be better to let the BSP (and its adaptation to
LmP) deal with the default fstype.

Signed-off-by: Jerome Brunet <jbrunet@baylibre.com>

### DIFF
--- a/meta-lmp-base/conf/distro/include/lmp.inc
+++ b/meta-lmp-base/conf/distro/include/lmp.inc
@@ -56,8 +56,7 @@ USERADD_GID_TABLES = "files/lmp-group-table"
 USERADD_ERROR_DYNAMIC = "error"
 
 # Default image formats
-IMAGE_FSTYPES = "wic wic.gz wic.bmap"
-IMAGE_FSTYPES_remove = "tar.gz tar.bz2"
+IMAGE_FSTYPES ?= "wic wic.gz wic.bmap"
 
 # No Ostree tarball by default (prefer ota instead)
 BUILD_OSTREE_TARBALL ?= "0"
--- a/meta-lmp-base/conf/distro/include/lmp.inc
+++ b/meta-lmp-base/conf/distro/include/lmp.inc
@@ -13,19 +13,19 @@ TCLIBCAPPEND = ""
 require conf/distro/include/arm-defaults.inc
 require conf/distro/include/security_flags.inc
 
-PREFERRED_PROVIDER_virtual/kernel ?= "linux-lmp"
-PREFERRED_PROVIDER_virtual/optee-os ?= "optee-os-fio"
-PREFERRED_PROVIDER_u-boot-fw-utils ?= "libubootenv"
-PREFERRED_RPROVIDER_u-boot-fw-utils ?= "libubootenv"
-PREFERRED_PROVIDER_zlib ?= "zlib"
-PREFERRED_PROVIDER_zlib-native ?= "zlib-native"
-PREFERRED_PROVIDER_nativesdk-zlib ?= "nativesdk-zlib"
-PREFERRED_PROVIDER_qemu-native ?= "qemu-native"
-PREFERRED_PROVIDER_nativesdk-qemu ?= "nativesdk-qemu"
+PREFERRED_PROVIDER_virtual/kernel ??= "linux-lmp"
+PREFERRED_PROVIDER_virtual/optee-os ??= "optee-os-fio"
+PREFERRED_PROVIDER_u-boot-fw-utils ??= "libubootenv"
+PREFERRED_RPROVIDER_u-boot-fw-utils ??= "libubootenv"
+PREFERRED_PROVIDER_zlib ??= "zlib"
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Thanks for your patch, this is a clean-up that is indeed needed, I just need to verify in every layer and make sure they are all behaving as expected.

IIRC we had that as we had BSP layers doing append, but this should indeed be fixed per layer and not globally.

jeromebrunet:
Once you are ready to merge it, could please apply this to the hardknott and honister support branches as well please ?

jeromebrunet:
Adding another distro providing defaults, letting other layers (such as BSP) make the assignements if necessary

meta-lmp-base/conf/distro/include/lmp.inc
```diff
@@ -56,8 +56,7 @@ USERADD_GID_TABLES = "files/lmp-group-table"
 USERADD_ERROR_DYNAMIC = "error"
 
 # Default image formats
-IMAGE_FSTYPES = "wic wic.gz wic.bmap"
-IMAGE_FSTYPES_remove = "tar.gz tar.bz2"
```
ricardosalveti, line 60: Remove itself should be fine, this was more to avoid generating artifacts that are not required at the base level, but we can easily move to lmp-machine-custom and have it be bsp specific.

meta-lmp-base/conf/distro/include/lmp.inc
```diff
@@ -56,8 +56,7 @@ USERADD_GID_TABLES = "files/lmp-group-table"
...
+IMAGE_FSTYPES ?= "wic wic.gz wic.bmap"
```
ricardosalveti, line 59: This one is a bit more annoying as there is no really a standard based on the BSP used, and might require a bit more work.

One example with intel-corei7-64, with the current settings:
```
# $IMAGE_FSTYPES [14 operations]
#   set? /home/rsalveti/build/lmp/build-lmp/conf/../../layers/openembedded-core/meta/conf/machine/include/x86/x86-base.inc:11
#     "wic"
#   append /home/rsalveti/build/lmp/build-lmp/conf/../../layers/meta-intel/conf/machine/intel-corei7-64.conf:33
#     "wic"
#   set /home/rsalveti/build/lmp/build-lmp/conf/../../layers/meta-lmp/meta-lmp-base/conf/distro/include/lmp.inc:60
#     "wic wic.gz wic.bmap"
...
#   append /home/rsalveti/build/lmp/build-lmp/conf/../../layers/meta-updater/classes/sota.bbclass:12
#     "${@bb.utils.contains('DISTRO_FEATURES', 'sota', 'ostreepush garagesign garagecheck ota-ext4 wic', ' ', d)}"
#   append /home/rsalveti/build/lmp/build-lmp/conf/../../layers/meta-updater/classes/sota.bbclass:13
#     "${@bb.utils.contains('BUILD_OSTREE_TARBALL', '1', 'ostree.tar.bz2', ' ', d)}"
#   append /home/rsalveti/build/lmp/build-lmp/conf/../../layers/meta-updater/classes/sota.bbclass:14
#     "${@bb.utils.contains('BUILD_OTA_TARBALL', '1', 'ota.tar.xz', ' ', d)}"
# pre-expansion value:
#   "wic wic.gz wic.bmap ${@bb.utils.contains('DISTRO_FEATURES', 'sota', 'ostreepush garagesign garagecheck ota-ext4 wic', ' ', d)} ${@bb.utils.contains('BUILD_OSTREE_TARBALL', '1', 'ostree.tar.bz2', ' ', d)} ${@bb.utils.contains('BUILD_OTA_TARBALL', '1', 'ota.tar.xz', ' ', d)}"
IMAGE_FSTYPES="wic wic.gz wic.bmap ostreepush garagesign garagecheck ota-ext4 wic   ota.tar.xz"
```

Now if we want to switch and still generate `wic.gz` and `wic.bmap`, we can't easily do that because an append would later affect initrds, and if we set per machine (in lmp-machine-custom, for example), it won't include the ostree ones, and will result in ```IMAGE_FSTYPES="wic wic.gz wic.bmap wic.vmdk wic.vdi"```, since a settings that is done per machine has higher priority and won't work with generic appends.

Similar thing happens with other machines, unfortunately. So unless we fix all the bsp layers to provide the basic and expect value at the machine config level, it won't really work with ?=.

jeromebrunet:
solved by commit 5c34aab0ba36985c89650741134565a799860aba

PR #573, opened 2022-03-28
bsp: opensbi: drop custom fdt function

ricardosalveti:
Same issue as happened with riscv_get_extra_oemake_image (can't be set
to "" as that won't match None), so remove riscv_get_extra_oemake_fdt
since we expect the dtb to be provided via u-boot fit.


MrCry0 (approved):
LGTM

PR #574, opened 2022-03-28
base: docker-ce: restore riscv64 support

ricardosalveti:
Restore patch content from what was already used by docker-moby, which
is required for proper risc-v support.

This is needed as linkshared is not yet supported on this arch.


ricardosalveti:
Merging to unblock the risc-v-64 builds.

PR #578, opened 2022-03-29
imx-related updates for honister

ricardosalveti:


MrCry0:
I'd rather wait for the fix to be merged in `honister` branch of `meta-freescale` instead of removing the obsolete dtb in our include file.
The merges are prepared.

ricardosalveti:
> I'd rather wait for the fix to be merged in `honister` branch of `meta-freescale` than remove the obsolete dtb in our include file. The fix is merged: [Freescale/meta-freescale#1021](https://github.com/Freescale/meta-freescale/pull/1021)
> 
> And here is the fix for manifest: [foundriesio/lmp-manifest#174](https://github.com/foundriesio/lmp-manifest/pull/174)

Thanks for pushing upstream, that was fast, let me drop the patch.

PR #580, opened 2022-03-29
bsp: lmp-machine-custom: am64xx: set different tmpdir for k3r5

ricardosalveti:
Avoid tmp overlay with the main machine as that results in build
failures when running with multiple cpu threads (due competing tmp
folder access).


ricardosalveti:
@Tim-Anderson still testing, but in theory that should be enough.

Tim-Anderson (approved):
LGTM

ricardosalveti:
Seems it is not yet the final fix, but a good one to have anyway.

PR #581, opened 2022-03-30
optee: resolve include clash with meta-arm

jeromebrunet:
meta-arm already provides optee. Some BSP rely on this version of optee
rather than the one provided by lmp. Include files, such as
optee-client.inc, clash with meta-arm which the bb recipe from meta-arm use
the include from meta-lmp.

Append "-fio" to includes provided by meta-lmp so they don't clash with
meta-arm.


ricardosalveti (approved):
LGTM, thanks!

jeromebrunet:
Can you put this on the honister support branch too please ?

ricardosalveti:
> Thx, Can you put this on the honister support branch too please ?

Honister is now in sync with master, which includes your change. Manifest is not yet updated to reflect the latest meta-lmp but we should be doing that later today (just landing a few other minor fixes first).

PR #585, opened 2022-04-04
bsp: u-boot-xlnx: switch to xilinx-v2022.01-rebase

ricardosalveti:
Prefer the rebase-based branch now that it is available in the xilinx
upstream tree (easier to manage and identify the local changes).

Also update the signed boot.bin which includes the updated SPL.


ricardosalveti:
Merging as it was already validated by @igoropaniuk over slack.

PR #593, opened 2022-04-11
Docker compose v2

ricardosalveti:
V2 has a different set of commands, and it is not 100% compatible with V1, but this PR also includes compose-switch to facilitate testing V2 without changing any clients.

The main difference is that to call docker compose you now need to use `docker compose` instead of `docker-compose`.

The ideal scenario is to change all clients to use V2 by default, without requiring compose-switch to be installed, but we can do that migration later.

### DIFF
--- a/meta-lmp-base/recipes-containers/docker-compose/docker-compose-switch_1.0.4.bb
+++ b/meta-lmp-base/recipes-containers/docker-compose/docker-compose-switch_1.0.4.bb
@@ -1,0 +1,30 @@
+HOMEPAGE = "https://github.com/docker/compose-switch"
+SUMMARY = "Compose Switch is a replacement to the Compose V1 \
+ docker-compose (python) executable. It translates the command \
+ line into Compose V2 docker compose then run the latter."
+SECTION = "devel"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://src/${GO_IMPORT}/LICENSE;md5=3b83ef96387f14655fc854ddc3c6bd57"
+
+SRC_URI = "git://github.com/docker/compose-switch.git;branch=master;protocol=https"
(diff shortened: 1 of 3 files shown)


ricardosalveti:
Still building and validating on every arch we support.

ricardosalveti:
Tested on armhf, arm64, intel and risc-v, all good.

ricardosalveti:
@doanac @mike-sul you can test by replacing `python3-docker-compose` at `meta-lmp-base/recipes-samples/images/lmp-feature-docker.inc` with `docker-compose` (if testing the new syntax) or just extend lmp-factory-image to also install `docker-compose-switch`, which should work as described at https://github.com/docker/compose-switch

MrCry0 (approved):
LGTM

mike-sul:
> you can test by replacing `python3-docker-compose` at `meta-lmp-base/recipes-samples/images/lmp-feature-docker.inc` with `docker-compose` (if testing the new syntax) or just extend lmp-factory-image to also install `docker-compose-switch`, which should work as described at https://github.com/docker/compose-switch

Is it possible to have the both compose implementations installed?

ricardosalveti:
> Is it possible to have the both compose implementations installed?

Yes, they won't conflict with each other AFAIK (just don't install docker-compose-switch).

mike-sul:
So, this PR enables the docker compose v2 by default. Maybe we should have some configuration variable that would allow to switch between versions. The v2 version needs to be tested well at first and I wouldn't enable it by default for customers factories until we make sure that it works.

ricardosalveti:
> So, this PR enables the docker compose v2 by default. 

Not enabled, it is just available now. For it to be enabled by default we would have to change the dependencies to install docker-compose instead of python3-docker-compose (which is what we install).

PR #594, opened 2022-04-12
Add params to define fiopush/check value

mike-sul:
New params to specify fiopush/fiocheck to use, by default or if the specified one are not valid/exist then fiopush/fiocheck built by OE are used.

### DIFF
--- a/meta-lmp-base/classes/lmp.bbclass
+++ b/meta-lmp-base/classes/lmp.bbclass
@@ -47,18 +47,18 @@ IMAGE_CMD:ostree:append () {
 
 run_fiotool_cmd () {
 	if [ -n "${SOTA_PACKED_CREDENTIALS}" ]; then
-                if [ -e "${SOTA_PACKED_CREDENTIALS}" ]; then
+		if [ -e "${SOTA_PACKED_CREDENTIALS}" ]; then
 			if [ "${OSTREE_API_VERSION}" = "v2" ]; then
 				"${1}" -repo "${OSTREE_REPO}" -creds "${SOTA_PACKED_CREDENTIALS}" -api-version "${OSTREE_API_VERSION}" -cor-id "${GARAGE_TARGET_NAME}-${GARAGE_TARGET_VERSION}"
 			else
 				"${1}" -repo "${OSTREE_REPO}" -creds "${SOTA_PACKED_CREDENTIALS}"
 			fi
-                else
-                        bbwarn "SOTA_PACKED_CREDENTIALS file does not exist."
-                fi
-        else
-                bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
-        fi
+		else
+			bbwarn "SOTA_PACKED_CREDENTIALS file does not exist."
+		fi
+	else
+		bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
+	fi
 }
 
 do_image_ostreepush[depends] += "ostreeuploader-native:do_populate_sysroot"
--- a/meta-lmp-base/classes/lmp.bbclass
+++ b/meta-lmp-base/classes/lmp.bbclass
@@ -1,3 +1,7 @@
+USE_FIOTOOLS ?= "1"
+FIO_PUSH_CMD ?= "fiopush"
+FIO_CHECK_CMD ?= "fiocheck"
+
 # Provided by meta-lmp-bsp or any other compatible BSP layer
 include conf/machine/include/lmp-machine-custom.inc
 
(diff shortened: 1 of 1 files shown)


mike-sul:
https://github.com/foundriesio/ci-scripts/pull/175/files and https://github.com/foundriesio/lmp-manifest/pull/178 will make LmP builds use the container's fiotools (lmp-sdk).
I verified all use-cases.

doanac (commented):
i think there's some tabs vs spaces formatting that may be happening.

Maybe `run_fiotool_cmd` shouldn't take any arguments. Then you can have that bit of
```
# Fallback to the OE built fiocheck if FIO_CHECK_CMD is not defined or it refers to non-existing or non-executable
```
in a single place

mike-sul:
> i think there's some tabs vs spaces formatting that may be happening.

It looks like the whole file have had mixed style, I'll fix it.

mike-sul:
> Maybe `run_fiotool_cmd` shouldn't take any arguments. Then you can have that bit of

I am not sure I get it, there are two use-cases, push and check, without any arguments how the function will know which of the commands to run and what's the fallback command?

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -47,53 +51,61 @@ IMAGE_CMD:ostree:append () {
...
-        if not (dtb.endswith('dtb') or dtb.endswith('dtbo')):
-            bb.error("KERNEL_DEVICETREE entry %s is not a .dtb or .dtbo file." % (dtb) )
-        return os.path.basename(dtb) + ';dtb/' + os.path.basename(dtb)
+	def transform(dtb):
+		if not (dtb.endswith('dtb') or dtb.endswith('dtbo')):
+			bb.error("KERNEL_DEVICETREE entry %s is not a .dtb or .dtbo file." % (dtb) )
+		return os.path.basename(dtb) + ';dtb/' + os.path.basename(dtb)
 
-    return ' '.join([transform(dtb) for dtb in alldtbs.split() if dtb])
+	return ' '.join([transform(dtb) for dtb in alldtbs.split() if dtb])
```
ricardosalveti, line 111: Can you drop these changes (in make_efi_dtb_boot_files) as part of this patch? Fine to have it done isolated in another patch, but we also need to make sure this won't cause problems as the logic is done in python.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -47,53 +51,61 @@ IMAGE_CMD:ostree:append () {
...
-                fi
-        else
-                bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
-        fi
+		else
+			bbwarn "SOTA_PACKED_CREDENTIALS file does not exist."
+		fi
+	else
+		bbwarn "SOTA_PACKED_CREDENTIALS not set. Please add SOTA_PACKED_CREDENTIALS."
+	fi
```
ricardosalveti, line 71: Please also fix this indentation issues in a separated patch/commit.

meta-lmp-base/classes/lmp.bbclass
mike-sul: I've dropped them. Can make this fix in another commit or PR.

meta-lmp-base/classes/lmp.bbclass
mike-sul: Done.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -64,17 +75,17 @@ run_fiotool_cmd () {
 do_image_ostreepush[depends] += "ostreeuploader-native:do_populate_sysroot"
 IMAGE_CMD:ostreepush:prepend () {
 	if [ "${USE_FIOTOOLS}" = "1" ]; then
-		run_fiotool_cmd "fiopush"
-		# force return so garage-push called from meta-updater's IMAGE_CMD:ostreepush is not executed
+		run_fiotool_cmd "${FIO_PUSH_CMD}" "fiopush"
+		# force return so garage-push called from meta-updater's IMAGE_CMD_ostreepush is not executed
```
ricardosalveti, line 79: No need to change this comment line, the correct syntax now (>= honister) is indeed `IMAGE_CMD:ostreepush`.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -64,17 +75,17 @@ run_fiotool_cmd () {
...
 	fi
 }
 
 do_image_garagecheck[depends] += "ostreeuploader-native:do_populate_sysroot"
 IMAGE_CMD:garagecheck:prepend () {
 	if [ "${USE_FIOTOOLS}" = "1" ]; then
-		run_fiotool_cmd "fiocheck"
-		# force return so garage-check called from meta-updater's IMAGE_CMD:garagecheck is not executed
+		run_fiotool_cmd "${FIO_CHECK_CMD}" "fiocheck"
+		# force return so garage-check called from meta-updater's IMAGE_CMD_garagecheck is not executed
```
ricardosalveti, line 88: Same here.

meta-lmp-base/classes/lmp.bbclass
mike-sul: wtf, I don't recall changing it :).

mike-sul: done

ricardosalveti: Probably because you changed this in hardknott and moved back the file to honister? But all good now :-)

ricardosalveti (approved):
LGTM

PR #601, opened 2022-04-14
apalis-imx6-sec: fix flashing and booting

MrCry0:
There is a set of commits that fixes regression issues with flashing and booting an image.


vanmaegima (approved):
LGTM

Tim-Anderson (approved):
LGTM

PR #602, opened 2022-04-14
Enable SE050 support for imx8mp-lpddr4-evk

vanmaegima:


Tim-Anderson (approved):
LGTM

vanmaegima:
Builds ok for lmp and lmp-mfgtool.

meta-lmp-bsp/recipes-kernel/linux/linux-lmp-fslc-imx/0001-FIO-internal-arch-arm64-dts-imx8mp-enable-I2C5-bus.patch
```diff
@@ -0,0 +1,63 @@
...
++	clock-frequency = <100000>;
++	pinctrl-names = "default";
++	pinctrl-0 = <&pinctrl_i2c5>;
++	pinctrl-assert-gpios = <&pca6416 2 GPIO_ACTIVE_HIGH>; /* set high to select i2c5 instead of flexcan1 */
++	status = "okay";
++
++	se050@48 {
++		compatible = "nxp,se050";
++		reg = <0x48>;
++	};
```
ricardosalveti, line 41: Is this node (se050) used by anyone? Because the element is enabled/used in u-boot / op-tee, so unclear if we need this enabled at the kernel side.

We should probably also apply this patch only when MACHINE_FEATURES has se05x defined.

vanmaegima: right, the node is not needed

angolini (approved):
LGTM

PR #605, opened 2022-04-15
bsp: lmp-machine-custom: am64xx-evm: unset EFI_PROVIDER

ricardosalveti:
EFI_PROVIDER should only be set by ebbr targets to avoid having fstab
and other related configs to be set even when not really required.

This is only required because meta-ti sets this by default at k3.inc.


caiotpereira (approved):

Tim-Anderson (approved):
LGTM

MrCry0 (approved):
LGTM

brenwoo (approved):

PR #607, opened 2022-04-19
base: busybox: enable partprobe by default

ricardosalveti:
Enable partprobe as an option to the same utility provided by parted
(which is GPLv3 licensed).


ricardosalveti:
Might not be that useful because it won't work on mounted partitions (ebusy).

Tim-Anderson:
I agree that the limitations of` busybox partprobe` will not work for the resize-helper

ricardosalveti:
Not really needed, let me close it.

PR #610, opened 2022-04-25
bsp: lmp-machine-custom: versal: fix WKS_FILES settings

ricardosalveti:
meta-xilinx-core sets a default WKS_FILES (versal-generic.conf) which is
not compatible with sota-based images (ostree).


angolini (approved):
LGTM

PR #623, opened 2022-04-29
base: lmp: define ETC_PASSWD_MEMBERS for /etc/passwd mirroring

ricardosalveti:
Similar to ETC_GROUP_MEMBERS, but for users that should be mirrored in
/etc/passwd to allow customizations.

Initially set to LMP_USER (default fio), but can be extended by the user.


MrCry0 (approved):

Tim-Anderson (approved):
LGTM

PR #625, opened 2022-05-03
base: lmp-feature-optee: drop sks, examples and pkcs11test

ricardosalveti:
They are not strictly required, so remove in order to make them optional
based on the user needs.


MrCry0 (approved):
LGTM

vanmaegima (approved):
LGTM

angolini (approved):
LGTM

Tim-Anderson (approved):
Generally LGTM but does that remove the use of libsks ?

ricardosalveti:
> Generally LGTM but does that remove the use of libsks ?

Yup, users still requiring libsks will have to add it manually as part of the image. This is so we can transition to libcktee from upstream, which is the preferred option.

PR #644, opened 2022-05-12
Fix duplication in machine overrides

ricardosalveti:
Fix duplicated entries in machine overrides to avoid issues during the bitbake override logic.


vanmaegima (approved):
LGTM

MrCry0 (approved):
LGTM

PR #645, opened 2022-05-12
Move fiovb to MACHINE_FEATURES and enabled for optee-os-fio-mfgtool

ricardosalveti:
Migrate fiovb from SOTA_EXTRA_CLIENT_FEATURES to MACHINE_FEATURES (as it is specific to the hw platform) and fix fiovb/early ta support for optee-os-fio-mfgtool.


ricardosalveti:
@mwasilew this should fix the lack of fiovb support in the mfgtools build.

Tim-Anderson (approved):
LGTM

mwasilew:
Isn't imx8mp missing?

quaresmajose (approved):

quaresmajose:
> Isn't imx8mp missing?

With this patch the SOTA_EXTRA_CLIENT_FEATURES is not used anymore

mwasilew:
Yes, I understand that. It was replaced with MACHINE_FEATURES. However only imx8mm-lpddr4-evk-sec has fiovb added to it's MACHINE_FEATURES. We also support imx8mp-lpddr4-evk-sec and imx6ull-evk-sec. They don't seem to be present in this patch.

quaresmajose:
yes, it isn't but from what I've checked it wasn't before either. anyway @ricardosalveti or another with more context can give a better answer than mine

mwasilew:
IIUC it should be there. I only tested on a closed imx8mm but I think the same problem should come up on other NXP boards. Let's wait for @rsalveti to comment.

ricardosalveti:
Yes, it was already missing for a few other targets, let me update to include the missing ones.

ricardosalveti:
Added for 8mp as well.

PR #647, opened 2022-05-16
base: u-boot-fio: add recipe for 2022.04

ricardosalveti:
FIO-specific recipe based on the upstream u-boot 2022.04 release.


ricardosalveti:
This tree is not yet containing all the customizations we have for the other versions, it is supposed to be a starting point.

PR #655, opened 2022-05-24
bsp: lmp-boot-firmware: uz3eg-iocc-sec: update boot.bin

ricardosalveti:
Update signed boot.bin to include latest u-boot-xlnx aba57774e4.


ricardosalveti:
@quaresmajose this is a signed boot.bin (bootloader) binary that is used for the -sec target, which assumes secure boot by default (we don't want to do the signing as part of the build process as that involves the RoT key).

quaresmajose:
@ricardosalveti thanks for the overview, I had already asked myself why we use a binary when look the MR

quaresmajose (approved):

PR #657, opened 2022-05-25
bsp: mfgtool-files: sec: zero is_secondary_boot during full_image

ricardosalveti:
is_secondary_boot is another variable that gets used by the common boot
logic, so make sure it is also initialized with a valid (zero) value.


MrCry0 (approved):
LGTM

PR #658, opened 2022-05-26
base: lmp: Preload Apps if enabled

mike-sul:
If the OE builtin preload is turned ON, then invoke the preload script which is supposed to copy Apps and their images into the rootfs that is being bitbaked by the current OE build.

### DIFF
--- a/meta-lmp-base/classes/lmp.bbclass
+++ b/meta-lmp-base/classes/lmp.bbclass
@@ -11,6 +11,26 @@ include conf/machine/include/lmp-partner-custom.inc
 # Allow customizations per factory level
 include conf/machine/include/lmp-factory-custom.inc
 
+preload_apps () {
+	if [ "${COMPOSE_APP_TYPE}" = "restorable" ]; then
+		if [ -f "${APP_PRELOADER}" ] && [ -x "${APP_PRELOADER}" ]; then
+			bbplain "Preloading Apps of the given Target: ${GARAGE_TARGET_NAME}-${target_version}"
+
+			REGISTRY_SECRETS_FILE="${APP_PRELOAD_REGISTRY_SECRETS_FILE}" \
+			TARGET_JSON_FILE="${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/sota/import/installed_versions" \
+			APP_SHORTLIST="${APP_SHORTLIST}" \
+			OCI_STORE_PATH="${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/sota/reset-apps" \
+			TOKEN_FILE="${APP_PRELOAD_TOKEN_FILE}" \
+			LOG_FILE="${APP_PRELOAD_LOG_FILE}" \
+				${APP_PRELOADER}
+		else
+			bbwarn "Apps preloading is turned ON, but the specified preloader either doesn't exist or is not executable: ${APP_PRELOADER}"
+		fi
+	else
+		bbwarn "Apps preloading is skipped because the preloader doesn't support the given type of Apps: ${COMPOSE_APP_TYPE}"
+	fi
+}
+
 # Support initial customized target via GARAGE_CUSTOMIZE_TARGET
 # This is set by our CI scripts and allows the initial target to populated by
 # the build process so it can be incorporated at the first aktualizr-lite run
@@ -58,6 +78,10 @@ IMAGE_CMD:ota:append () {
 		rm -rf ${OTA_BOOT}
 		mv ${OTA_SYSROOT}/boot ${OTA_BOOT}
 		mkdir -p ${OTA_SYSROOT}/boot
+	fi
+
+	if [ "${APP_PRELOAD_WITHIN_OE_BUILD}" = "1" ]; then
+		preload_apps
 	fi
 }
 OTA_BOOT = "${WORKDIR}/ota-boot"


ricardosalveti:
Relevant ci-script changes: https://github.com/foundriesio/ci-scripts/pull/185

quaresmajose (approved):

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -20,6 +20,22 @@ sota_fstab_update() {
 
 ROOTFS_POSTPROCESS_COMMAND:append:sota = " sota_fstab_update; "
 
+preload_apps () {
+        if [ -f "${APP_PRELOADER}" ] && [ -r "${APP_PRELOADER}" ] && [ -x "${APP_PRELOADER}" ]; then
```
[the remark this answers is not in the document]
quaresmajose: probably yes, the `-x` has the  `-f` implied

```
-r
file has read permission (for the user running the test)
```
^ `bitbake` building the image inside the `pseudo` that is like a `fakeroot` so `-r` is not needed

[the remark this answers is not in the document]
mike-sul: `-x`  returns `true` for a folder, so to make sure that it is a file you have to use `-f`.
Also, `-r` is used to make sure that a file is readable by a given user, because both `-x` and `-f` can return `true` while the file cannot be executed because a current user doesn't have a read permission :). https://stackoverflow.com/questions/10319652/check-if-a-file-is-executable

ricardosalveti (approved):
LGTM

meta-lmp-base/classes/lmp.bbclass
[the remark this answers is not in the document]
MrCry0: The `exec` access is enough for binary files, but for bashed scripts, a `read` right is required.
```
$ ls -l z
---x--x--x 1 cryo cryo 24 тра 27 19:41 z
$ [ -f z ] && [ -x z ] && echo OK
OK
$ ./z 
/bin/bash: ./z: Permission denied
```$ ./z.sh 
/usr/bin/python: can't open file '/home/cryo/tmp/./z.sh': [Errno 13] Permission denied

The same for python:
```
$ ./z.sh 
/usr/bin/python: can't open file '/home/cryo/tmp/./z.sh': [Errno 13] Permission denied
```

Are we sure the APP_PRELOADER will always be binary?

File: meta-lmp-base/classes/lmp.bbclass [hunk 2]
```diff
@@ -30,6 +50,10 @@ IMAGE_CMD:ota:append () {
 		${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/sota/import/installed_versions \
 			${GARAGE_TARGET_NAME}-${target_version}
 	fi
+
+	if [ "${APP_PRELOAD_WITHIN_OE_BUILD}" = "1" ]; then
+		preload_apps
```
mike-sul, line 55: @ricardosalveti Do you think it's possible to run this command as a root user? `su root -c preload_apps`.

doanac (commented):
lgtm

mike-sul:
FYI: Corresponding PR to ci-scripts (https://github.com/foundriesio/ci-scripts/pull/185) has been merged.

meta-lmp-base/classes/lmp.bbclass
quaresmajose: This command runs in `IMAGE_CMD:ota:append` when building the image, for that bitbake makes use of the tool `pseudo`  that work like `fakeroot` with some additional functionality. This to say that the `preload_apps` is already running with root privileges.

mike-sul: For some reason `dockerd` doesn't think so. I just simply added  `dockerd &` to the preload-apps.sh and it complains that it doesn't have root permissions.

When I tried to run `dockerd-rootless` it complained that it cannot be started by a root user :).

I'll send a link to the error later on.

quaresmajose: I think this is another issue, running `dockerd &` is not possible at build time because because at build time the host architecture is the one in aws `amd64` and the target can be different as we use cross-compilation.

bitbake as the concept of native application (used during the build for run on the host and not on the target) but running a daemon is not supported afaik.

mike-sul: I understand it, but this is very unlikely the reason because I am not running containers by means of `dockerd`, I am just running an executable (dockerd) built for amd64 on amd64 host.

This is the error message

```
| Preloading Apps of the given Target: intel-corei7-64-lmp-1367
| time="2022-05-31T17:40:08.869842422Z" level=info msg="Starting up"
| dockerd needs to be started with root privileges. To run dockerd in rootless mode as an unprivileged user, see https://docs.docker.com/go/rootless/
```

mike-sul: https://ci.foundries.io/projects/msul-dev01/lmp/builds/1367/intel-corei7-64/artifacts/console.log

mike-sul: This is the error message for the rootless dockerd

```
| |--
| Preloading Apps of the given Target: intel-corei7-64-lmp-1361
| + [ -w /home/builder/.docker/run ]
| + [ -d /home/builder ]
| + rootlesskit=
| + command -v docker-rootlesskit
| + command -v rootlesskit
| + rootlesskit=rootlesskit
| + break
| + [ -z rootlesskit ]
| + :
| + :
| + : builtin
| + : auto
| + : auto
| + net=
| + mtu=
| + [ -z  ]
| + command -v slirp4netns
| + [ -z  ]
| + command -v vpnkit
| + net=vpnkit
| + [ -z  ]
| + mtu=1500
| + [ -z  ]
| + _DOCKERD_ROOTLESS_CHILD=1
| + export _DOCKERD_ROOTLESS_CHILD
| + id -u
| + [ 0 = 0 ]
| + echo This script must be executed as a non-privileged user
| + exit 1
| This script must be executed as a non-privileged user
```

mike-sul:
Since the other relevant PRs are merged (ci-scripts, repoz) we can merge this too and start dogfooding.

doanac (approved):
this looks like its ready to be merged, right?

meta-lmp-base/classes/lmp.bbclass
ricardosalveti: Right, I don't think we can run this script as root (and not sure if it would be recommended anyway), but would have to try to be sure.

I think the issue here is that dockerd knows it is not real root, and while called from pseudo it thinks it is root because of the user id (which is 0 due pseudo). Maybe what we can do here (without changing docker) is trying to run this command as a user that is not root?

mike-sul: The current implementation does not require running `dockerd` for preloading, it preloads just the skopeo store be means of `skopeo` utility. We just decided to go with non dockerd approach.

mike-sul: It's been ready since the last week :). I am going to rebase it on the latest main, check whether it's bitbakable, and then merge.

PR #669, opened 2022-06-01
Add generic config option for ostree-bootcsumdir-source

ricardosalveti:
Make the trigger to enable ostree deploy copying content from
/usr/lib/ostree-boot a generic configuration, as then the user can
enable in a more generic fashion.

### DIFF
--- a/meta-lmp-base/conf/distro/lmp.conf
+++ b/meta-lmp-base/conf/distro/lmp.conf
@@ -13,6 +13,7 @@ OSTREE_KERNEL ?= "${@oe.utils.conditional('KERNEL_IMAGETYPE', 'fitImage', '${KER
 OSTREE_KERNEL_ARGS_COMMON ?= "root=LABEL=otaroot rootfstype=ext4"
 OSTREE_KERNEL_ARGS ?= "${OSTREE_KERNEL_ARGS_COMMON}"
 OSTREE_SPLIT_BOOT ?= "0"
+OSTREE_DEPLOY_USR_OSTREE_BOOT ?= "0"
 DISTRO_FEATURES:append = " sota"
 DISTRO_FEATURES_NATIVE:append = " sota"
 INHERIT += "sota"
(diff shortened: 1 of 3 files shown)


MrCry0 (approved):
LGTM

meta-lmp-base/conf/distro/lmp.conf
```diff
@@ -13,6 +13,7 @@ OSTREE_KERNEL ?= "${@oe.utils.conditional('KERNEL_IMAGETYPE', 'fitImage', '${KER
 OSTREE_KERNEL_ARGS_COMMON ?= "root=LABEL=otaroot rootfstype=ext4"
 OSTREE_KERNEL_ARGS ?= "${OSTREE_KERNEL_ARGS_COMMON}"
 OSTREE_SPLIT_BOOT ?= "0"
+OSTREE_DEPLOY_USR_OSTREE_BOOT ?= "0"
```
quaresmajose, line 16: it seems to me that this new variable `OSTREE_DEPLOY_USR_OSTREE_BOOT` is implicit in the one `INITRAMFS_RECOVERY_IMAGE` introduced in the previous patch

ricardosalveti: Yes, I decided to move this to a variable instead (and have it set at the image level) to avoid having multiple recipes/packages shipping the same file, which would fail at the package install/do_rootfs process.

ricardosalveti: Oh sure, I decided to use a new flag because this is also going to be used for other binaries, such as firmware related ones, so not really specific to the recovery image, that was just the first user for this.

Having OSTREE_DEPLOY_USR_OSTREE_BOOT = 1 without OSTREE_DEPLOY_USR_OSTREE_BOOT won't cause any bad side effect, as there won't be any file to copy / deploy.

quaresmajose: > Oh sure, I decided to use a new flag because this is also going to be used for other binaries, such as firmware related ones, so not really specific to the recovery image, that was just the first user for this.

makes sense for me now

quaresmajose (approved):
LGTM

PR #671, opened 2022-06-07
Support for systemd-boot

ricardosalveti:
- Adds ostree support for using /boot/loader as a directory (required by systemd-boot due ESP/vfat partition)
  - Patch as an adaptation from an upstream proposal, which keeps the old link way as default still
- Adds minimal systemd-boot support in ostree, enough for copying boot/loader/loader.conf between deployments
- Update wks and installer to use systemd-boot by default as we want to drop support for grub-efi (due GPLv3)

Still to be stress tested and tested on multiple targets. Tested mainly on intel at this point.

This is the base work required for EFI Secure Boot support.


vanmaegima (approved):
LGTM

MrCry0 (approved):
LGTM

angolini (approved):
I don't have anything else to comment. For me it looks good

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -31,6 +31,24 @@ IMAGE_CMD:ota:append () {
...
+		if [ "${OSTREE_LOADER_LINK}" != "0" ]; then
+			bbfatal "Systemd-boot requires OSTREE_LOADER_LINK to be set to '0'"
+		fi
+		if [ "${OSTREE_SPLIT_BOOT}" != "1" ]; then
+			bbfatal "Systemd-boot requires OSTREE_SPLIT_BOOT to be set to '1'"
+		fi
+		if [ "${OSTREE_BOOTLOADER}" != "none" ]; then
+			bbfatal "Systemd-boot requires OSTREE_BOOTLOADER to be set to 'none'"
+		fi
+		# As upstream doesn't yet support systemd-boot, we have to undo none and change as needed
```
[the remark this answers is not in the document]
ricardosalveti: Ostree does not yet support systemd-boot, so the way to integrate this via our local patches was to force OSTREE_BOOTLOADER to none, which is handed by https://github.com/uptane/meta-updater/blob/master/classes/image_types_ota.bbclass#L28, and why we need to undo that part here.

Once we land the proper support upstream, this won't be needed and meta-updater will configure it correctly.

ricardosalveti:
@vanmaegima take your time.

meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
```diff
@@ -239,58 +239,52 @@ mkdir -p /boot
 
 # Handling of the target root partition
 mount $rootfs /tgt_root
-mount -o rw,loop,noatime,nodiratime /run/media/$1/$2 /src_root
+mount -o rw,loop,noatime,nodiratime /run/media/${live_dev_name}3/rootfs.img /src_root
```
quaresmajose, line 242: is the `rootfs.img` pretended? it seams a typo

quaresmajose (approved):
lgtm not counting the ostree content patches

meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
ricardosalveti: It is always rootfs.img, which is why I changed to be explicit here.

vanmaegima (approved):
LGTM

Tim-Anderson (approved):
LGTM

PR #672, opened 2022-06-07
versal: initial support for building with optee support

ricardosalveti:
At this stage simply adding the right build dependencies and basic configuration in order to get it to depend on optee-os.

XSA also updated to be aligned with the v2022.1 BSP, otherwise it fails to boot.


MrCry0 (commented):
LGTM

MrCry0 (approved):

PR #678, opened 2022-06-09
archiver: avoid race conditions when using multiconfig and sharing the same TMPDIR

quaresmajose:
Add temporarily the archiver.bbclass from git rev cb8647c089

When [1] is merged upstream we can revert this.
[1] https://lists.openembedded.org/g/openembedded-core/message/166784

```
4edc18f8 base: archiver: avoid race conditions when using multiconfig and sharing the same TMPDIR
a8ff21f4 base: add archiver.bbclass from yocto-4.0.1
```


angolini (approved):
Great PR!

ricardosalveti (approved):
LGTM

PR #683, opened 2022-06-13
Early startup for oe builtin preloaded apps

mike-sul:
Add early App startup for preloaded *Restorable* Apps.

1. One-shot systemd service aimed to populate the dockerd repository with Registry referencies of the preloaded Apps' images. It makes `dockerd` avoid making any call to Registries during Apps startup.

2. Adjust existing one-shot systemd service that does Apps early start in such a way that it can start both type of Apps, *restorable* and *compose*.

### DIFF
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
@@ -60,6 +60,7 @@ PACKAGES += "${PN}-get ${PN}-lite ${PN}-lite-lib ${PN}-lite-dev"
 FILES:${PN}-get = "${bindir}/${PN}-get"
 FILES:${PN}-lite = " \
                     ${bindir}/${PN}-lite \
+                    ${bindir}/aklite-apps \
                     ${nonarch_libdir}/tmpfiles.d/${PN}-lite.conf \
                     "
 FILES:${PN}-lite-lib = "${nonarch_libdir}/lib${PN}_lite.so"
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
@@ -1,7 +1,7 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 BRANCH:lmp = "master"
-SRCREV:lmp = "96640d24be97b597e2bfb545c22b255427888f60"
+SRCREV:lmp = "e05f7eb7a852491c2812b89b60af9979aca44501"
 
 SRC_URI:lmp = "gitsm://github.com/foundriesio/aktualizr-lite;protocol=https;branch=${BRANCH};name=aktualizr \
     file://aktualizr.service \
--- a/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start.bb
+++ b/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start.bb
@@ -7,16 +7,18 @@ SRC_URI = " \
 	file://compose-apps-early-start.service \
 	file://compose-apps-early-start-recovery \
 	file://compose-apps-early-start \
+	file://compose-apps-register-images.service \
 "
 
 inherit systemd
 
-SYSTEMD_SERVICE:${PN} = "compose-apps-early-start.service"
+SYSTEMD_SERVICE:${PN} = "compose-apps-early-start.service compose-apps-register-images.service"
 
 do_install() {
 	install -d ${D}${systemd_system_unitdir}
 	install -m 0644 ${WORKDIR}/compose-apps-early-start.service ${D}${systemd_system_unitdir}/
 	install -m 0644 ${WORKDIR}/compose-apps-early-start-recovery.service ${D}${systemd_system_unitdir}/
+	install -m 0644 ${WORKDIR}/compose-apps-register-images.service ${D}${systemd_system_unitdir}/
 	install -d ${D}${bindir}
 	install -m 0755 ${WORKDIR}/compose-apps-early-start ${D}${bindir}/
 	install -m 0755 ${WORKDIR}/compose-apps-early-start-recovery ${D}${bindir}/
--- a/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
+++ b/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 
-function start_compose_app() {
+start_compose_app() {
 	app=$1
 	cd /var/sota/compose-apps/${app}
 
@@ -14,21 +14,45 @@ function start_compose_app() {
 	fi
 }
 
-if [ -f /var/lmp/default-apps ] ; then
-	if [ -d /var/sota/compose-apps ] ; then
-		for x in $(ls /var/sota/compose-apps) ; do
-			if ! grep -q $x /var/lmp/default-apps 2>/dev/null ; then
-				echo "Disabling preloaded app: $x"
-				rm -rf /var/sota/compose-apps/$x
-			fi
-		done
+start_compose_apps() {
+	if [ -f /var/lmp/default-apps ] ; then
+		if [ -d /var/sota/compose-apps ] ; then
+			for x in $(ls /var/sota/compose-apps) ; do
+				if ! grep -q $x /var/lmp/default-apps 2>/dev/null ; then
+					echo "Disabling preloaded app: $x"
+					rm -rf /var/sota/compose-apps/$x
+				fi
+			done
+		fi
 	fi
-fi
 
-if [ -d /var/sota/compose-apps ] ; then
-	for app in `ls /var/sota/compose-apps` ; do
-		start_compose_app ${app}
-	done
+	if [ -d /var/sota/compose-apps ] ; then
+		for app in `ls /var/sota/compose-apps` ; do
+			start_compose_app ${app}
+		done
+	else
+		echo "No apps defined"
+	fi
+}
+
+start_restorable_apps() {
+	default_apps_path="/var/lmp/default-apps"
+	shortlist=""
+
+	if [ -f $default_apps_path ] ; then
+		for app in `cat $default_apps_path`; do
+			shortlist="${shortlist}${app},"
+		done
+		if [ ! -z $shortlist ] ; then
+			shortlist=${shortlist::-1}
+                fi
+	fi
+
+	aklite-apps run --apps "${shortlist}"
+}
+
+if [ -d /var/sota/reset-apps ] ; then
+	start_restorable_apps
 else
-	echo "No apps defined"
+	start_compose_apps
 fi
--- a/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start-recovery
+++ b/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start-recovery
@@ -1,15 +1,31 @@
 #!/bin/sh
 
-# Force compose down, restart docker and try again
+recover_compose_apps() {
+	# Force compose down, restart docker and try again
 
-for app in `ls /var/sota/compose-apps` ; do
-	cd /var/sota/compose-apps/${app}
-	docker compose down
-done
+	for app in `ls /var/sota/compose-apps` ; do
+		cd /var/sota/compose-apps/${app}
+		docker compose down
+	done
 
-# TODO: extend with support for restorable-apps, which should
-# allow us to remove /var/lib/docker completely
+	systemctl restart docker
+}
 
-systemctl restart docker
+recover_restorable_apps() {
+	systemctl stop docker
+	rm -rf /var/sota/compose-apps/*
+	rm -rf /var/lib/docker
+
+	systemctl restart compose-apps-register-images.service
+	systemctl start docker
+}
+
+if [ -d /var/sota/reset-apps ] ; then
+	recover_restorable_apps
+else
+	recover_compose_apps
+fi
+
 systemctl reset-failed compose-apps-early-start.service
+sleep 60
 systemctl restart compose-apps-early-start.service
--- a/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-register-images.service
+++ b/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-register-images.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Register images of Apps at a docker store repository
+Before=docker.service
+Before=lmp-device-auto-register.service
+ConditionPathExists=/var/sota/reset-apps
+ConditionPathExists=!/var/sota/sql.db
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/bin/aklite-apps register
+
+[Install]
+WantedBy=multi-user.target


mike-sul:
@doanac This PR works along with PR https://github.com/foundriesio/aktualizr-lite/pull/176.

mike-sul:
I restructured the change to:
1. make it easier to review;
2. avoid changes to the "compose apps" case as much as possible.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -11,6 +11,26 @@ include conf/machine/include/lmp-partner-custom.inc
...
+
+			REGISTRY_SECRETS_FILE="${APP_PRELOAD_REGISTRY_SECRETS_FILE}" \
+			TARGET_JSON_FILE="${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/sota/import/installed_versions" \
+			APP_SHORTLIST="${APP_SHORTLIST}" \
+			OCI_STORE_PATH="${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var/sota/reset-apps" \
+			TOKEN_FILE="${APP_PRELOAD_TOKEN_FILE}" \
+			LOG_FILE="${APP_PRELOAD_LOG_FILE}" \
+				${APP_PRELOADER}
+		else
+			bbwarn "Apps preloading is turned ON, but the specified preloader either doesn't exist or is not executable: ${APP_PRELOADER}"
```
doanac, line 27: probably should be `bbfatal`?

meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
```diff
@@ -61,6 +61,7 @@ FILES:${PN}-get = "${bindir}/${PN}-get"
 FILES:${PN}-lite = " \
                     ${bindir}/${PN}-lite \
                     ${nonarch_libdir}/tmpfiles.d/${PN}-lite.conf \
+                    ${bindir}/fioapp \
```
doanac, line 64: i'm guessing the name of this is going to change once aklite is updated. I guess the sha to aklite will also have to get bumped here.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start.bb
```diff
@@ -22,6 +24,8 @@ do_install() {
 	install -d ${D}${bindir}
 	install -m 0755 ${WORKDIR}/compose-apps-early-start ${D}${bindir}/
 	install -m 0755 ${WORKDIR}/compose-apps-early-start-recovery ${D}${bindir}/
+	install -m 0755 ${WORKDIR}/start-apps ${D}${bindir}/
+	install -m 0755 ${WORKDIR}/start-restorable-apps ${D}${bindir}/
```
doanac, line 28: maybe we could/should just have one script deal with both rather than splitting this up into so many pieces. I don't have a strong opinion. Sort of a distro maintenance question.

doanac (commented):
seems like the right direction.

meta-lmp-base/classes/lmp.bbclass
mike-sul: ok.

meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
mike-sul: Correct. I just wanted to check whether it works from end2end perspective as well as get some feedback before finalizing this work.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start.bb
mike-sul: OK. I actually implemented four flavors of the given implementation, the way you suggested is one of them. The overall approach works regardless of the flavor, I decided to go with the separate service scripts because it makes reviewing easier and arguably easier to do further support.

The four implementation options I tried are:

1. Dedicated systemd service and corresponding dedicated script for the restorable Apps. Which of the services to trigger is determined by  service conditions (e.g. `ConditionPathExists=/var/sota/reset-apps`). The current implementation is intact at all.

2. The current PR version. A single systemd service, two use-case specific scripts as well as an entry script with `if else` statement

3. A single systemd service, a single script that includes implementation for both use-cases as well as `if else`.

4. A single systemd service and a single script for both use-cases. `if else` is in the `aklite-app` utility that knows how to run both type of apps.


I played with all four approached, and none of them is ideal, each of them has its pros and cons. Based on my experience with it and taking into account a need in further support and maybe some customization, I incline towards 2 or 3. It implies that `aklite-apps` will run only restorable apps while compose apps will be started by means of `docker compose` and a script, effectively, as it is now.

doanac (commented):
i like how this has turned out.

mike-sul:
@doanac @ricardosalveti I tested the given change for the aklite version implemented here https://github.com/foundriesio/aktualizr-lite/pull/176. Once the aklite PR is merged I'll add a new commit to this PR to bump the new aklite version.

The eternal reboot issue seems irrelevant to this change/PR, I'll create a jira ticket with steps to reproduce it for the current meta-lmp version.

mike-sul:
I think I've done dev testing enough to be sure that it's ready for review and QA testing.
I added a workaround to the "reboot" issue to this PR. A proper fix should be implemented in another PR, the jira ticket is FFTK-1636.

meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
```diff
@@ -61,6 +61,7 @@ FILES:${PN}-get = "${bindir}/${PN}-get"
 FILES:${PN}-lite = " \
                     ${bindir}/${PN}-lite \
                     ${nonarch_libdir}/tmpfiles.d/${PN}-lite.conf \
+                    ${bindir}/aklite-apps \
```
ricardosalveti, line 64: nit, please have this line below 62 (${bindir}/${PN}-lite).

meta-lmp-base/recipes-containers/docker/files/docker.service
```diff
@@ -1,7 +1,7 @@
 [Unit]
 Description=Docker Application Container Engine
 Documentation=https://docs.docker.com
-After=network-online.target docker.socket firewalld.service containerd.service fio-docker-fsck.service
+After=network-online.target docker.socket firewalld.service containerd.service fio-docker-fsck.service compose-apps-register-images
```
ricardosalveti, line 4: I don't think it needs to be explicitly called after compose-apps-register-images here. I believe just having the other job with before docker should be enough. Also curious if this could cause a dep chain issue when a condition at compose-apps-register-images causes this job to not be executed.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-register-images.service
```diff
@@ -0,0 +1,14 @@
...
+Description=Register images of Apps at a docker store repository
+Before=docker.service
+Before=lmp-device-auto-register.service
+ConditionPathExists=/var/sota/reset-apps
+ConditionPathExists=!/var/sota/sql.db
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=/usr/bin/aklite-apps hack
```
ricardosalveti, line 11: Hack?

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
```diff
@@ -14,21 +14,43 @@ function start_compose_app() {
...
 	fi
-fi
+}
+
+function start_restorable_apps() {
+	default_apps_path="/var/lmp/default-apps"
+	shortlist=""
+
+	if [ -f $default_apps_path ] ; then
+	for app in `cat $default_apps_path`; do
```
ricardosalveti, line 43: Missing indentation at the for block.

ricardosalveti:
There are some bashisms which we should remove as well (bash is gplv3):

```
checkbashisms meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
possible bashism in meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start line 3 ('function' is useless):
function start_compose_app() {
possible bashism in meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start line 17 ('function' is useless):
function start_compose_apps() {
possible bashism in meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start line 38 ('function' is useless):
function start_restorable_apps() {
possible bashism in meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start line 44 (should be VAR="${VAR}foo"):
                        shortlist+="${app},"
```

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start-recovery
```diff
@@ -1,15 +1,31 @@
...
-systemctl restart compose-apps-early-start.service
+function recover_restorable_apps() {
+	systemctl stop docker
+	rm -rf /var/sota/compose-apps/*
+	rm -rf /var/lib/docker
+
+	systemctl restart compose-apps-register-images.service
+	systemctl start docker
+	systemctl reset-failed compose-apps-early-start.service
+	systemctl restart compose-apps-early-start.service
```
ricardosalveti, line 26: reset-failed and restart is common to both, can be done at the end of the script.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start-recovery
```diff
@@ -21,6 +22,7 @@ function recover_restorable_apps() {
 	systemctl restart compose-apps-register-images.service
 	systemctl start docker
 	systemctl reset-failed compose-apps-early-start.service
+	sleep 60
```
ricardosalveti, line 25: What is the consequence if we get a broken app which can cause this job to be executed in loop? I guess lmp will be running but docker will be constantly restarting on an infinite loop.

meta-lmp-base/recipes-containers/docker/files/docker.service
mike-sul: >I don't think it needs to be explicitly called after compose-apps-register-images here.

Yes, you're right, it's not necessary, I put it there because for some reason we put it for another services. I'll remove it.

> Also curious if this could cause a dep chain issue when a condition at compose-apps-register-images causes this job to not be executed.

`After` and `Before` is about ordering of servicing they do not define dependencies as `Requires` does. So, this job (`docker.service`) should be started regardless of `compose-apps-register-images` condition.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start-recovery
mike-sul: That's correct. At least a device won't be constantly restarting.
Unfortunately, I don't see a way to limit a number for attempts for `early-start` --> `early-start-recovery` --> `early start` cycle. We could limit it if we use `ExecStopPost` to invoke the recovery script.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-register-images.service
ricardosalveti: @mike-sul is hack the right argument here? If so, why is it called hack?

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start-recovery
ricardosalveti: Right, the problem is that at this stage the device is not even registered, so there won't be much that the user can do with it (assuming login is disabled, for example).

Having it rebooting could end up causing the recovery process to be initialized (once we enable it), which might be better. I understand that rebooting is not really helpful, but not rebooting is not a lot better.

I'm ok to keep the sleep for now, until we enable the recovery logic, we just need to make sure that this limitation gets documented somehow (@kprosise). Basically we need to explicitly say that the apps are required to be known to work and not conflict with each other, otherwise the device will be in a bad state.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-register-images.service
mike-sul: Yes, the `hack` is a correct argument here. It's a nice name for a command isn't it? :).

For some reason, neither the TAR image format, that is used for loading of images into a dockerd's store, nor dockerd does NOT allow to "pin"/map a loaded image to a digested image reference (image URL with hash). Our solution (compose app, aklite) refers to images by digested reference. So, even though all required images are loaded into a dockerd's store, it (dockerd) communicates to Registry to fetch image manifest because it does not known whether a given image referenced  as e.g. hub.f.io/factory/foo/app@sha256:<hash> really exists in its store or not. Once it downloads an image manifest it gets an image config hash and checks whether it's already in the store, if so, only then it knows that the given image already loaded.
Obviously, such behavior is not appropriate for our "early app startup in offline mode" case. So, this command `hack` "tells" `dockerd` that prelaoded images referenced by digested reference already present in its store what makes `dockerd` to not do any connections to Registry. Specifically, the command adds required mapping (image uri -> image config hash) to  `/var/lib/docker/image/overlay2/repositories.json`.

It's kind of dockerd hacking hence the command is called `hack`. Having said that, I don't have a strong opinion about it and open to any suggestions on the command naming.

^^ @doanac FYI.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start-recovery
mike-sul: > that at this stage the device is not even registered

Well, it depends, I can imagine a scenario when a device is preregistered before its first boot.

> so there won't be much that the user can do with it 

Yes, as I stated earlier, this workaround does not help end users much because their device does not function properly anyway.
However, I find it is useful for customers' engineers during dev and test phases of Apps development. It would be hard/inconvenient for them to figure out App failure reason if a device keeps rebooting.

It's just a workaround and a proper solution I think is out of the scope of the given PR/effort. IMO, a proper solution should limit a number of attempts to two, if it doesn't work after starting from scratch then it doesn't make any sense to keep trying.

quaresmajose (approved):

kprosise:
@ricardosalveti I opened a [Jira ticket](https://foundriesio.atlassian.net/browse/FFTK-1660) to document the limitation.

MrCry0:
> > Like, "force-register" or so on?
> 
> Maybe just `aklite-apps register`?

Looks good to me.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-register-images.service
doanac: looking at it now that a week has passed - i don't clearly remember what "hack" even does - so it's a bad name. I think we should name it better to help the next person that goes to debug this

ricardosalveti:
> Maybe just `aklite-apps register`?

Yeah, let's use register instead of hack as that has a way too generic meaning :-)

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-register-images.service
mike-sul: renamed the cmd to `register`.

mike-sul:
> > Maybe just `aklite-apps register`?
> 
> Yeah, let's use register instead of hack as that has a way too generic meaning :-)

done (now we are not hacking anything :)).

MrCry0 (approved):
LGTM

PR #685, opened 2022-06-13
recipes-samples: lmp-feature-optee fiovb is set with separate feature

Tim-Anderson:
Currently the lmp-feature-optee includes optee-fiovb but that should
only be enabled if there is an RPMB available (i.e. secured).


ricardosalveti (approved):
LGMT

PR #695, opened 2022-06-16
Add support for efitools and LockDown.efi

ricardosalveti:
Add efitools recipe and support for generating and installing LockDown.efi, which can be used to automatically enroll the UEFI certs used by the UEFI Secure Boot process.


MrCry0 (approved):
LGTM

quaresmajose (approved):

PR #696, opened 2022-06-20
base: systemd: only enable gnu-efi when efi is available

ricardosalveti:
Systemd cannot build with gnu-efi support when support for efi is not
enabled, so make it depend on it.


vanmaegima (approved):
LGTM

MrCry0 (approved):
LGTM

PR #704, opened 2022-06-29
build with clang toolchain by default

quaresmajose:
base: lmp: build with clang toolchain by default

For recipes that doesn't build we can go back to gcc toolchain
with TOOLCHAIN:pn-recipes = "gcc" in main local.conf or
TOOLCHAIN = "gcc" on himself in a bbappend

Depends on https://github.com/foundriesio/meta-lmp/pull/705

### DIFF
--- a/meta-lmp-base/conf/distro/include/lmp.inc
+++ b/meta-lmp-base/conf/distro/include/lmp.inc
@@ -83,6 +83,10 @@ INHERIT += "buildhistory"
 INHERIT += "image-buildinfo"
 BUILDHISTORY_COMMIT = "1"
 
+# Clang toolchain
+TOOLCHAIN ?= "clang"
+RUNTIME = "llvm"
+
 PREMIRRORS ??= "\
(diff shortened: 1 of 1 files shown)


ricardosalveti:
Nice, would you mind breaking this down into 2 prs? One for adding all the clang stuff and another for enabling it as the default toolchain?

quaresmajose:
Sure I will split it

angolini (approved):
I'm not sure I like the errors as comments, but I think I understand where they come from and I don't think they hurt, so LGTM

quaresmajose:
> I'm not sure I like the errors as comments, but I think I understand where they come from and I don't think they hurt, so LGTM

The reason for the errors in the commands comes from the upstream 
https://github.com/kraj/meta-clang/blob/master/conf/nonclangable.conf

Maybe I should put this one in draft because it only contain the last commit and all the others comes from https://github.com/foundriesio/meta-lmp/pull/705

MrCry0 (approved):

ricardosalveti (approved):
LGTM, and here we go :-)

PR #719, opened 2022-07-07
base: wireguard-tools: add nostamp for create_runtime_spdx

ricardosalveti:
SPDX data requires all runtime dependencies to be updated in order to
reflect the correct json files, and as it is depending on a kernel
module recipe (wireguard-module), we need to force nostamp for
create_runtime_spdx otherwise it will be ignored due sstate cache
filtering (siggen).

This can be reverted once the correct fix lands upstream.

For more information:
https://www.yoctoproject.org/irc/%23yocto.2022-07-06.log.html#t2022-07-06T20:25:15


angolini (approved):

MrCry0 (approved):
LGTM

PR #723, opened 2022-07-11
Fixes for the external kernel modules signing issue

ricardosalveti:
This is to fix the module not signed issue we noticed on imx8mm evk (wifi module is external).

This was caused by a race condition during the rm_work step, which erased the sysroot for make-mod-scripts, causing sign-file to error out due missing openssl library.

### DIFF
--- a/meta-lmp-base/conf/distro/include/lmp.inc
+++ b/meta-lmp-base/conf/distro/include/lmp.inc
@@ -118,6 +118,9 @@ WARN_QA:remove = "${WARN_TO_ERROR_QA}"
 ERROR_QA:append = " ${WARN_TO_ERROR_QA}"
 ERROR_QA:remove = "version-going-backwards"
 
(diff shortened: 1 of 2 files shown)


Tim-Anderson (approved):
LGTM

ricardosalveti:
Moving to draft as it is still not working correctly on rebuilds, investigating.

ricardosalveti:
The fix for the signature failure is indeed correct (caused by rm_work), but there is still a task dependency issue which is not causing the module to be rebuilt when the kernel is rebuilt but without any changes to the kernel itself (build caused by cache invalidation due other env changes), so it is just reusing the kernel modules from the cache, which is not signed.

ricardosalveti:
I think the easiest way forward here is to force a new kernel build or remove local cache (invalid), and the new builds should be fine.

PR #730, opened 2022-07-13
bsp: mfgtool-files: imx: delete bootfirmware_version during provisioning

ricardosalveti:
Prefer to delete bootfirmware_version instead of writing a default value
at provisioning to allow the correct initial value to be set during
first boot instead.

By setting to a default value (0) every first update will also cause the
boot firmware to be updated as a consequence, since the firmware version
will differ.


Tim-Anderson (approved):
LGTM

MrCry0 (approved):
LGTM. Agree

mwasilew:
Tested manually and it works as expected (at least on imx8mm-evk)

PR #731, opened 2022-07-13
base: recipes-sota: Add aklite offline update cmd

mike-sul:
- Add `aklite-offline` command to the aktualizr-lite package if `aklite-offline` feature is enabled/added to the package config.
  
- Bump aklite version that supports an offline update.

Signed-off-by: Mike <mike.sul@foundries.io>


ricardosalveti:
> I'm not sure if its a part of the aklite package or if it should be its own. I'm sure ricardo will know

I think it is fine to have it as part of the ak-lite package because then we won't need to change the image to install an additional package when enabled.

In this case it will be available if enabled via package config, otherwise the binary won't be part of the package.

quaresmajose (approved):

MrCry0 (approved):

PR #738, opened 2022-07-18
bsp: stm32mp15-eval: drop build of non-existent dts

ricardosalveti:
As Linux LmP is used by default, don't build non-existent dts, that
exists only in ST Linux fork.


quaresmajose (approved):
LGTM

PR #741, opened 2022-07-18
bsp: lmp-machine-custom: imx: prefer bsp recipe for gpu-viv

ricardosalveti:
Prefer the version that is aligned with the imx bsp instead of the fslc
one, in order to be aligned with our kernel which is also bsp based.


MrCry0 (approved):
LGTM

PR #747, opened 2022-07-20
bsp: wpa-supplicant: disable control port over nl80211 for imx8mm-evk

ricardosalveti:
wpa-supplicant 2.10 enables NL80211_EXT_FEATURE_CONTROL_PORT_OVER_NL80211 by
default if provided by the driver, but that causes a handshake issue with ath10k
(wireless device used on imx8mm-evka).

Can be reverted once better investigated and reproduced with latest firmware
and kernel (from 5.15 BSP).


Tim-Anderson (approved):
LGTM

PR #752, opened 2022-07-29
base: lmp-image-common: add curl util

MrCry0:
This tool is pretty small and way useful for troubleshooting to be
included in all LmP images.

Signed-off-by: Oleksandr Suvorov <oleksandr.suvorov@foundries.io>

PR #753, opened 2022-07-29
Alter: base: lmp-feture-debug: add curl tool

MrCry0:
This is an alternative to https://github.com/foundriesio/meta-lmp/pull/752

This tool is useful for troubleshooting.


ricardosalveti (approved):
LGTM, let's go with this change instead and users can use that fragment on their own factory.

PR #755, opened 2022-08-01
Add support of u-boot imx 2022.04

MrCry0:
Prepare recipes for enabling u-boot imx-2022.04 for imx boards.
Reduce the PR to only adding the u-boot imx-2022.04 support without switching boards to use it.
There will be other PRs for switching boards to this u-boot as long as they are ready.


angolini (approved):

vanmaegima (approved):
LGTM

PR #756, opened 2022-08-02
 create-spdx: avoid race conditions caused when ipk packages is written

quaresmajose:
- base: add create-spdx.bbclass from yocto-4.0.2

```
Add temporarily the create-spdx.bbclass from git rev a4bfb5ce

When [1] is merged upstream we can revert this.
[1] https://lists.openembedded.org/g/openembedded-core/message/168773

```

- create-spdx: ignore packing control files from ipk and deb
```
Otherwise spdx can have references for data that is not packed
in the package delivered because this contol data is temporarly
and only exist while the package is been write.

During do_package_write_ipk task in do_package_ipk the control
files is cleaned up at the end. This can create a race condiction
when the do_create_spdx task runs the add_package_files function
and these files is been deleted at same time in the
task do_package_write_ipk.
```

This PR will be reverted with https://github.com/foundriesio/meta-lmp/pull/757

### DIFF
--- a/meta-lmp-base/classes/create-spdx.bbclass
+++ b/meta-lmp-base/classes/create-spdx.bbclass
@@ -571,6 +571,7 @@ python do_create_spdx() {
                     pkgdest / package,
                     lambda file_counter: oe.sbom.get_packaged_file_spdxid(pkg_name, file_counter),
                     lambda filepath: ["BINARY"],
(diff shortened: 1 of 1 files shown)


angolini:
Could you, please, share the reason the PR is bringing a class from upstream? (So, in future we can understand when to stop using it)

angolini (approved):

ricardosalveti (approved):
LGTM

PR #765, opened 2022-08-10
Fix basic support for intel + TPM2

ricardosalveti:
Base changes should be sent upstream as well, but need to check when we get a master-next that is functional as the base versions are not the same.


MrCry0 (approved):
LGTM

quaresmajose (approved):
LGTM

PR #766, opened 2022-08-10
base: tpm2-pkcs11: set default storedir to /var/tpm2_pkcs11

ricardosalveti:
Upstream uses /etc/tpm2-pkcs11 by default, which is not really ideal (as
the default location for the store sqlite databse), so set to
/var/tpm2_pkcs11 instead, and create the respective tmpfiles logic in
order for the system to create the required dir during boot.


quaresmajose (approved):
LGTM

PR #767, opened 2022-08-10
base: systemd: use nonarch_libdir instead of nonarch_base_libdir

ricardosalveti:
Systemd expects to install the systemd-boot files under /usr/lib, so use
nonarch_libdir instead of nonarch_base_libdir in order to be compatible
with builds that are not using usrmerge.

### DIFF
--- a/meta-lmp-base/recipes-core/systemd/systemd_%.bbappend
+++ b/meta-lmp-base/recipes-core/systemd/systemd_%.bbappend
@@ -89,5 +89,7 @@ do_install:append() {
 	ln -sf ../systemd-timesyncd-update.service ${D}${systemd_system_unitdir}/sysinit.target.wants/systemd-timesyncd-update.service
 
 	# Remove systemd-boot as it is provided by a separated recipe and we can't disable via pkgconfig
-	rm -rf ${D}${nonarch_base_libdir}/systemd/boot
+	if ${@bb.utils.contains('PACKAGECONFIG', 'efi', 'true', 'false', d)}; then
+		rm -r ${D}${nonarch_libdir}/systemd/boot
+	fi
 }


Tim-Anderson (approved):
LGTM thanks!

MrCry0 (approved):
LGTM

meta-lmp-base/recipes-core/systemd/systemd_%.bbappend
```diff
@@ -89,5 +89,5 @@ do_install:append() {
 	ln -sf ../systemd-timesyncd-update.service ${D}${systemd_system_unitdir}/sysinit.target.wants/systemd-timesyncd-update.service
 
 	# Remove systemd-boot as it is provided by a separated recipe and we can't disable via pkgconfig
-	rm -rf ${D}${nonarch_base_libdir}/systemd/boot
+	rm -rf ${D}${nonarch_libdir}/systemd/boot
```
quaresmajose, line 92: I think is better to fail when the directory doesn't exist and if something change in future this will rise an error

```
if ${@bb.utils.contains('EFI_PROVIDER','systemd-boot','true','false',d)}; then
    rm -r ${D}${nonarch_libdir}/systemd/boot
fi
```

angolini (approved):

meta-lmp-base/recipes-core/systemd/systemd_%.bbappend
ricardosalveti: Yup, will update.

ricardosalveti: Done.

quaresmajose (approved):
LGTM

PR #774, opened 2022-08-16
bsp: kernel-modules: generalize imx-gpu-viv version

MrCry0:
bsp: kernel-modules: generalize imx-gpu-viv version
      
      As the bbappend recipe doesn't bind to a specific version of
      imx-gpu-viv, make it general.


quaresmajose:
Good catch! as the content of this append can by applied to any version is better to have the append agnostic to version
and rename it to `kernel-module-imx-gpu-viv_%.bbappend`

quaresmajose (approved):
LGTM

Tim-Anderson (approved):
LGTM

ricardosalveti (approved):
LGTM

MrCry0:
> Good catch! as the content of this append can by applied to any version is better to have the append agnostic to version and rename it to `kernel-module-imx-gpu-viv_%.bbappend`

Thanks, it's way better for this case :)

PR #780, opened 2022-08-17
Update Kria XSA / PM based on v2022.1

ricardosalveti:
* XSA update
* pm obj update as it needs to be specific to kria
* tpm2 init on boot


Tim-Anderson (approved):
LGTM

PR #786, opened 2022-08-19
Wireguard: invert the KERNEL_BUILTIN_WIREGUARD logic

quaresmajose:


meta-lmp-base/recipes-kernel/wireguard/wireguard-module_%.bbappend
```diff
@@ -5,6 +5,6 @@ python __anonymous() {
     pprovider = d.getVar("PREFERRED_PROVIDER_virtual/kernel")
     if pprovider != "linux-lmp" and pprovider != "linux-lmp-lts" and pprovider != "linux-lmp-rt" \
             and pprovider != "linux-lmp-dev" and pprovider != "linux-lmp-fslc-imx" \
-            and d.getVar("KERNEL_BUILTIN_WIREGUARD") != "1":
+            and d.getVar("KERNEL_BUILTIN_WIREGUARD") == "0":
```
ricardosalveti, line 8: We can probably remove the lines above as well, and just look for KERNEL_BUILTIN_WIREGUARD.

quaresmajose: right, whenever the variable KERNEL_BUILTIN_WIREGUARD is not defined as "0" the builtin kernel module will be used

MrCry0 (approved):
Could we add some details to commit messages on why that was done? (Like "As now a built-in kernel wireguard is supported for most machines, reverting logic reduces and simplifies code").
Besides that, LGTM

angolini:
Is the default value set somewhere?

quaresmajose:
> Is the default value set somewhere?

no default value set, it remains as it was before

angolini (approved):

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -546,6 +541,7 @@ OSTREE_KERNEL_ARGS:tegra194 ?= "\${cbootargs} ${OSTREE_KERNEL_ARGS_COMMON}"
 OSTREE_BOOTLOADER:tegra194 = "syslinux"
 
 # Allwinner (sun8i)
+KERNEL_BUILTIN_WIREGUARD:sun8i = "0"
```
ricardosalveti, line 544: I think sun8i is using a newer kernel as well:
```
conf/machine/include/sunxi.inc:PREFERRED_PROVIDER_virtual/kernel ?= "linux-mainline"
conf/machine/include/sunxi.inc:PREFERRED_VERSION_linux-mainline ?= "5.15.%"
```

quaresmajose: In fact, it was the only one I didn't test. but I agree that the out of tree module is not needed in this kernel

PR #791, opened 2022-08-23
base: u-boot: u-boot-fio_imx-2022.04 decrease preference

igoropaniuk:
Add DEFAULT_PREFERENCE = -1, so recipe is not built by default.


ricardosalveti (approved):
LGTM, thanks!

We're doing the switch to this version by default in a separated pr.

PR #792, opened 2022-08-23
tf-a-tools: Add temporarily build fixes until it is submitted upstream

quaresmajose:


meta-lmp-bsp/dynamic-layers/stm-st-stm32mp/recipes-bsp/trusted-firmware-a/tf-a-tools_2.6.bb
```diff
@@ -0,0 +1,31 @@
+SUMMARY = "Cert_create & Fiptool for fip generation for Trusted Firmware-A"
+LICENSE = "BSD-3-Clause"
+LIC_FILES_CHKSUM = "file://license.rst;md5=1dd070c98a281d18d9eefd938729b031"
+
+SRC_URI = "git://github.com/ARM-software/arm-trusted-firmware.git;protocol=https;branch=master \
+           file://0001-tools-allow-to-use-a-root-key-password-from-command-.patch \
+           file://0002-fix-fiptool-respect-OPENSSL_DIR.patch \
+           "
+
+SRCREV = "a1f02f4f3daae7e21ee58b4c93ec3e46b8f28d15"
```
igoropaniuk, line 10: Have you tried to switch to 2.7 release?

quaresmajose: not yet but I am going to try

quaresmajose: add the 2.7 version

quaresmajose:
```
RDEPENDS:${PN} += "libcrypto"
```
bitbake RDEPENDS only works for packages and native targets don't use packages so my assumption is right about RDEPENDS but the question so how can i fix this problem without using the RDEPENDS

ricardosalveti:
There is nothing specific to st here, and we have other boards using arm-trusted-firmware as well, so might be better to make it part of meta-lmp-base instead.

ricardosalveti:
> Please old on a bit as I need to confirm an issue that I see in our CI
> 
> ```
> | fiptool: error while loading shared libraries: libcrypto.so.3: cannot open shared object file: No such file or directory
> ```
> 
> So my affirmation `- RDEPENDS and FILES is not needed` can be wrong.

Are you getting this error with the native build? Usually DEPENDS is already enough for having it as part of the sysroot-deps list, unless it is not looking for the right path.

quaresmajose:
yes, fixed with https://github.com/STMicroelectronics/meta-st-stm32mp/pull/41/commits/6c5464b7c6429134c8d6c68ab1dc24bd699579e8 that is included in this PR

```
tf-a-tools: fix the RPATH for the native target
the libcrypto.so.3 is provide on the native sysroot of the fiptool
but the fiptool don't search the native sysroot so it will work
in hosts with the library libcrypto.so.3 and fail when the libcrypto.so.3
not found. this patch fix this host contamination bug.

| fiptool: error while loading shared libraries: libcrypto.so.3: cannot open shared object file: No such file or directory
```

meta-lmp-bsp/dynamic-layers/stm-st-stm32mp/recipes-bsp/trusted-firmware-a/tf-a-tools_2.6.bb
quaresmajose: meta-st upstream prefere to stay with the 2.6 so we can do the bump in our side

MrCry0 (approved):
LGTM

quaresmajose:
> There is nothing specific to st here, and we have other boards using arm-trusted-firmware as well, so might be better to make it part of meta-lmp-base instead.

For others boards we can use the arm-trusted-firmware from meta-arm https://git.yoctoproject.org/meta-arm/tree/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a.inc that have all the pieces needed (fiptool native 2.7 in this case). 
that's where i found the fix for the issue RPATH

For stm32mp we need to use their fiptool `tf-a-tools-native` because they packing it in a different recipe
https://github.com/STMicroelectronics/meta-st-stm32mp/blob/00c2494b66b8fa2c0fc2008fa802a2adbeed966e/classes/fip-utils-stm32mp.bbclass#L1

meta-lmp-bsp/dynamic-layers/stm-st-stm32mp/recipes-bsp/trusted-firmware-a/tf-a-tools_2.7.bb
```diff
@@ -4,10 +4,9 @@ LIC_FILES_CHKSUM = "file://license.rst;md5=1dd070c98a281d18d9eefd938729b031"
 
 SRC_URI = "git://github.com/ARM-software/arm-trusted-firmware.git;protocol=https;branch=master \
            file://0001-tools-allow-to-use-a-root-key-password-from-command-.patch \
-           file://0002-fix-fiptool-respect-OPENSSL_DIR.patch \
```
angolini: Maybe it's good to mention why this patch is being removed (is it upstream already?)

quaresmajose: These is a backported patch https://github.com/foundriesio/meta-lmp/pull/792/commits/5866016e47477e689ac9a10fa0e7fafdb469ed06#diff-b5e33ff710ac85f7839b8785da65e0f6063017c1cb7ae0f1b81720c8e9cefd83R12 from upstream version 2.7

```
Upstream-Status: Backport [https://github.com/ARM-software/arm-trusted-firmware/commit/dd14d0f63f7d049cbcbfd3f8128f05e9a351bbe5]
```

ricardosalveti:
> > There is nothing specific to st here, and we have other boards using arm-trusted-firmware as well, so might be better to make it part of meta-lmp-base instead.
> 
> For others boards we can use the arm-trusted-firmware from meta-arm https://git.yoctoproject.org/meta-arm/tree/meta-arm/recipes-bsp/trusted-firmware-a/trusted-firmware-a.inc that have all the pieces needed (fiptool native 2.7 in this case). that's where i found the fix for the RPATH issue
> 
> For stm32mp we need to use their fiptool `tf-a-tools-native` because they packing it in a different recipe https://github.com/STMicroelectronics/meta-st-stm32mp/blob/00c2494b66b8fa2c0fc2008fa802a2adbeed966e/classes/fip-utils-stm32mp.bbclass#L1

Sure, we can merge both later on (since we're also using 2.7 atf from upstream).

ricardosalveti (approved):
LGTM, just need to get a +1 from @igoropaniuk

igoropaniuk (changes requested):
LGTM

PR #798, opened 2022-08-25
Fix plug-and-trust-seteec and el2go-auto-register

ricardosalveti:
Fix libckteec.so.0 library name and update plug-and-trust-seteec to reduce the need for a custom config (in order to get it to work without major changes with openssl 3).


doanac (approved):

doanac (approved):
tested this on my device.

PR #803, opened 2022-08-26
optee-os-fio backport from main-next

quaresmajose:


ricardosalveti:
Since we own the fork I would prefer having these patches available in https://github.com/foundriesio/optee_os instead. We're also about to bump to 3.18, so it will be good check if they are also relevant there.

And in case relevant on optee-os master, mind also sending the patches to https://github.com/OP-TEE/optee_os? Then we can just tag as backported patches.

quaresmajose:
Ok so I will wait for the 3.18 merge and try again and if they will be relevant I will send it to our fork and mainline.

ricardosalveti:
Patches already included in https://github.com/foundriesio/optee_os/tree/3.18%2Bfio.

PR #804, opened 2022-08-26
bsp: kernel: mfgtool: enable missing configs for apalis-imx6

vanmaegima:
Without these configs, we have several peripherals failing to load.
Enable these in order to make udc available on kernel boot.


ricardosalveti (approved):
LGTM

PR #806, opened 2022-08-29
[fromup] recipes-kernel: correct wifi startup on am62xx

Tim-Anderson:


ricardosalveti:
Please update the patch subject to be "bsp: linux-lmp-ti-staging: ...".

Then extend the commit message saying where you got this patch from, possibly with a link. Please also add the Upstream-Status line inside the patch, like "Upstream-Status: Pending", if that is indeed the case.

Tim-Anderson:
updated

ricardosalveti (approved):
LGTM

PR #810, opened 2022-08-31
non clangable: add new bsp packages

quaresmajose:
- base: lmp: non-clangable: add trusted-firmware-a used in am6xxx
- base: lmp: non-clangable: add tf-a-fio used in stm32mp1


ricardosalveti (approved):
LGTM

PR #814, opened 2022-09-01
ostree: fix clang static linking 

quaresmajose:
glibc is built with gcc and hence encodes some libgcc specific builtins
which are not found when doing static linking with clang using compiler-rt,
so use libgcc to fix floting point emulation undefined reference:

- undefined reference to `__unordtf2'
- undefined reference to `__letf2'

this brings again https://github.com/foundriesio/meta-lmp/commit/a3892eb03dab13b3ef3daca692521c759e3f2283 but now only for static linked.


MrCry0 (approved):
WoW! LGTM

ricardosalveti (approved):
LGTM, nice one.

PR #822, opened 2022-09-14
stm32mp15-eval lmp-base fixes

igoropaniuk:


mike-scott (approved):
LGTM.  Nice work on the boot files fixes.

meta-lmp-bsp/recipes-bsp/u-boot/u-boot-base-scr.bbappend
```diff
@@ -1 +1,11 @@
 FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+do_install:append:stm32mp15-eval() {
+    install -m 0644 ${DEPLOYDIR}/boot.scr ${D}/boot.scr
+    install -m 0644 ${DEPLOYDIR}/uEnv.txt ${D}/uEnv.txt
+}
+
+FILES:${PN} += " \
```
quaresmajose, line 8: the FILES needs the same override `FILES:${PN}:stm32mp15-eval` otherwise will fail as this files not found in other machines

igoropaniuk: sorry, missed that. Will address in another PR

MrCry0: @quaresmajose @igoropaniuk we have a generic version of boot.cmd and uEnv.txt.in which deploys as boot.scr and uEnv.txt, so I think it's safe to keep a common FILES instruction.

quaresmajose: I am testing this MR now in https://github.com/foundriesio/lmp-manifest/pull/221 so in a couple of hours we will know.

quaresmajose: @MrCry0 is right and we don't need `FILES:${PN}:stm32mp15-eval` override as MR https://github.com/foundriesio/lmp-manifest/pull/221 build without any issue.
@igoropaniuk however, this way I think we don't need the full block of `FILES:${PN}`, tomorrow I will test

ricardosalveti: We always deploy both boot.scr and uEnv.txt, so I think it is fine to create a generic do_install function that install both, instead of having it just for stm32mp15-eval.

Then we can both everything to meta-lmp-base (from the bbappend).

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -484,6 +484,8 @@ EXTRA_IMAGEDEPENDS:append:stm32mp15-eval = "flashlayouts-stm32mp1"
 LMP_FLASHLAYOUT:stm32mp15-eval = "FlashLayout_stm32mp1-optee.tsv"
 LMP_FLASHLAYOUT_BOARD_NAME:stm32mp15-eval = "stm32mp157c-ev1"
 LMP_BOOT_FIRMWARE_FILES:stm32mp15-eval = "arm-trusted-firmware/tf-a-stm32mp157c-ev1-emmc.stm32 fip/fip-stm32mp157c-ev1-optee.bin"
+IMAGE_INSTALL:append:stm32mp15-eval = " kernel kernel-devicetree u-boot-base-scr"
+IMAGE_INSTALL:remove:sota:stm32mp15-eval = "kernel kernel-devicetree u-boot-base-scr"
```
ricardosalveti, line 488: We shouldn't change IMAGE_INSTALL here, please switch with MACHINE_ESSENTIAL_EXTRA_RDEPENDS instead (and no need to remove when sota is used).

Also use u-boot-default-script instead of u-boot-base-scr.

igoropaniuk: But why should we keep `kernel kernel-devicetree` for sota?

meta-lmp-bsp/recipes-bsp/u-boot/u-boot-base-scr.bbappend
igoropaniuk: @ricardosalveti but should we have both (boot.scr and uEnv.txt) in rootfs? For most platforms they are installed via WKS parser and `IMAGE_BOOT_FILES`.

ricardosalveti: I think it is fine to have them in the rootfs, even if they are also listed in IMAGE_BOOT_FILES because that one is only for copying into the vfat/boot partition.

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
ricardosalveti: I don't think installing them would cause any side effect (and I think kernel is already installed), but if you think it is better to have it not installed, then do a similar append/remove logic but using MACHINE_ESSENTIAL_EXTRA_RDEPENDS instead.

PR #797, opened 2022-08-25
LmP staging bitbake class

quaresmajose:
This class will implement some pending patches that we have
and some workarounds needed in LmP.

For now there is only fixes for:
- linux kernel modules dependencies and inheritance
- network during compile for go-mod.bbclass

Issue #245, opened 2020-12-10
fioconfig compile failure behind corporate proxy...


salerio-gs:
I am seeing the compile failure below. As far as I know my http_proxy and htts_proxy environment variables are good and the proxy is working fine. If I edit the fioconfig recipe and insert **GOPROXY=direct** infront of the compilation command I get a similar error but only after a good couple of minutes of compilation. Any idea's what's wrong there?

```
Build Configuration:
BB_VERSION           = "1.48.0"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "ubuntu-18.04"
TARGET_SYS           = "aarch64-lmp-linux"
MACHINE              = "raspberrypi4-64"
DISTRO               = "lmp"
DISTRO_VERSION       = "3.2"
TUNE_FEATURES        = "aarch64 armv8a crc crypto cortexa72"
TARGET_FPU           = ""
meta-lmp-base        = "HEAD:3beb1d4791967ea4f1123dea519fb29574d102c2"
meta-oe
meta-networking
meta-filesystems
meta-perl
meta-python          = "HEAD:b9dcf17700014dec679307dc1247882dccd439ca"
meta-virtualization  = "HEAD:9fe997733d9bad4ac24dfb41e91a0e06b9e82791"
meta-clang           = "HEAD:1dfc3a905ad0fed7a3b300aa51153365f9950fe7"
meta-updater         = "HEAD:075ca14b8800dcae1c2532389e5dd219f6e7c2dd"
meta-security        = "HEAD:4583ab9b08c1c582a9d7e8ab70d6685ed4720799"
meta-arm
meta-arm-toolchain
meta-arm-bsp         = "HEAD:1a0a9700f2e667752e985f57087272553b239aa4"
meta-freescale       = "HEAD:b9870755b89b8bc81d0fee60e3050155e94a9c46"
meta-freescale-3rdparty = "HEAD:d23ec3443d2a2fb36f29d0ef04d2806dab286676"
meta-raspberrypi     = "HEAD:09a3c11696c75593f8e475da5dfb401016c6aaca"
meta-riscv           = "HEAD:9a75ffbb21ebee61782576d524a454d04a9f91e2"
meta-rtlwifi         = "HEAD:ccf8cdc0e2282cf3989d35504989dc404bbe1370"
meta-intel           = "HEAD:5492270c796daf8b7a7cc9cd93880c2bb25523c0"
[...]

PR #824, opened 2022-09-15
LmP go-mod-network bitbake class

quaresmajose:
In LmP layer we need network support to build Go Modules.
So this classe will handle this cases without touching the upstream version.


MrCry0 (approved):

meta-lmp-base/classes/go-mod-network.bbclass
```diff
@@ -0,0 +1,17 @@
+# Foundries LmP Go Modules with network support
+#
+# In LmP layer we need network support to build Go Modules
```
angolini, line 3: Do we need to call the class lmp-go-mod-network? (only to make sure it's from lmp when inherit it from other files)

angolini (approved):
Nice!

meta-lmp-base/classes/go-mod-network.bbclass
quaresmajose: yeah we can do it, I will change the name after when https://github.com/foundriesio/lmp-manifest/pull/221 finish.

quaresmajose: thinking about it again, this class is like a wrapper around the upstream version go-mod.bbclass, the main ideia is to don't lost changes from the upstream version.
It is not very easy to see the connection between the two prefixing with lmp `go-mod -> lmp-go-mod-network`. I think is better to have this without lmp in the name because the class works with all go modules and not only the ones in meta-lmp, ate the end it is a ugly hack to only enable network during compile.

ricardosalveti:
> In LmP layer we need network support to build Go Modules.
> So this classe will handle this cases without touching the upstream version.

While our go-based recipes do need network, this class was created to replace the one from oe-core in order to allow any go-lang based recipe that requires network to work under a proxy (as described at https://github.com/foundriesio/meta-lmp/issues/245).

If we change our own recipes this will keep working as expected, but others might break users requiring proxy.

quaresmajose:
> > In LmP layer we need network support to build Go Modules.
> > So this classe will handle this cases without touching the upstream version.
> 
> While our go-based recipes do need network, this class was created to replace the one from oe-core in order to allow any go-lang based recipe that requires network to work under a proxy (as described at #245).
> 
> If we change our own recipes this will keep working as expected, but others might break users requiring proxy.

I didn't know the reasons behind its existence. Thanks for explaining.
So the patch needs more test and I need to understand the proxy requirements and see if it still necessary.

```
diff --git a/meta-lmp-base/classes/go-mod-network.bbclass b/meta-lmp-base/classes/go-mod-network.bbclass
index 2da7136a..9f031a8a 100644
--- a/meta-lmp-base/classes/go-mod-network.bbclass
+++ b/meta-lmp-base/classes/go-mod-network.bbclass
@@ -13,5 +13,5 @@ python bb_utils_export_proxies() {
     bb.utils.export_proxies(d)
 }
 
-do_compile[prefuncs] += "bb_utils_export_proxies"
 do_compile[network] = "1"
```

With a quick test I can build `fioconfig` without calling `bb.utils.export_proxies(d)`

ricardosalveti:
> So the patch needs more test and I need to understand the proxy requirements and see if it still necessary.

Try to enable network usage via proxy in OE and see if it works or not (you can also set to an invalid proxy and give it a try, if it works then you know it is not using the proxy).

quaresmajose:
I have tested the proxy without calling the `b.utils.export_proxies(d)` and the bitbake fetcher can handle the proxy without issues.
```
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
@@ -45,3 +45,11 @@ FILES:${PN} += " \
        ${systemd_system_unitdir}/fioconfig-extract.service \
        ${datadir}/fioconfig/handlers \
 "
+
+# https://wiki.yoctoproject.org/wiki/Working_Behind_a_Network_Proxy
+export http_proxy='http://myproxy.example.com:1080/'
+export https_proxy='https://myproxy.example.com:1080/'
+export ftp_proxy='http://myproxy.example.com:1080/'
+export ALL_PROXY='socks://myproxy.example.com:1080/'
+export all_proxy='socks://myproxy.example.com:1080/'
+export no_proxy='example.com'
```

the use of an invalid will cause a build fail:
```
| HASH[moduleIndex]: 6df15e245826e2936cc7a414d4d5fc4b5aa3dfc69238aff2bcf2ad37f765a3cc
| go: github.com/ThalesIgnite/crypto11@v1.2.1 (replaced by github.com/doanac/crypto11@v1.2.2-0.20200715151421-f3d2e17ac497): Get "https://proxy.golang.org/github.com/doanac/crypto11/@v/v1.2.2-0.20200715151421-f3d2e17ac497.mod": proxyconnect tcp: dial tcp: lookup myproxy.example.com: no such host
| WARNING: exit code 1 from a shell command.
ERROR: Task (/oe/lmp-manifest/repo/build/conf/../../layers/meta-lmp/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb:do_compile) failed with exit code '1'
```

So I will close this PR and will implement the network access during build as part of PR https://github.com/foundriesio/meta-lmp/pull/797

quaresmajose:
Tested without proxy with `bitbake fioconfig --skip-setscene` and works in the same way
```
diff --git a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
index d048f876..0c38ba5d 100644
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
@@ -45,3 +45,11 @@ FILES:${PN} += " \
        ${systemd_system_unitdir}/fioconfig-extract.service \
        ${datadir}/fioconfig/handlers \
 "
+
+# https://wiki.yoctoproject.org/wiki/Working_Behind_a_Network_Proxy
+unset http_proxy
+unset https_proxy
+unset ftp_proxy
+unset ALL_PROXY
+unset all_proxy
+unset no_proxy
```

PR #837, opened 2022-09-26
base: lmp: non-clangable: add imx-atf

ricardosalveti:
Build is not reproducible, causing the binary hash to change at every clean build (and causing unexpected boot firmware updates).


quaresmajose (approved):
LGTM

PR #839, opened 2022-09-27
lmp-boot-firmware: limit bootfirmware_version length

jeromebrunet:
When the bootfirmware_version is not fixed by PV, the length of bootfirmware_version depends on the number of files passed in LMP_BOOT_FIRMWARE_FILES

Having a variable length is no problem for u-boot env. With backend such as fiovb it is a problem because the storage size of the variable is fixed - 128b in the current boot-common.cmd.in.

Make a single md5sum out of the current bootfirmware_version content so keeps on changing if any of the files in LMP_BOOT_FIRMWARE_FILES changes and length does not exceed 128b.


ricardosalveti (approved):
LGTM

MrCry0 (approved):

quaresmajose (approved):
LGTM, thanks @jeromebrunet

PR #844, opened 2022-10-03
base: vulkan-loader: mx8mm-nxp-bsp doesn't support vulkan

quaresmajose:
Vulkan is disabled on meta-freescale for 8M Mini since ee92b5a,
so add the same logic to not recommends libvulkan-imx on this machines.
https://github.com/Freescale/meta-freescale/commit/ee92b5a8b2cad45de267d3cf6dae9251f5ed11fd

This append is version specific because it's a workaround until
upstream found and merge a solution then we can drop this patch.
https://github.com/Freescale/meta-freescale/pull/1249

This also fix:
```
ERROR: Nothing RPROVIDES 'libvulkan-imx' (but /srv/oe/build/conf/../../layers/openembedded-core/meta/recipes-graphics/vulkan/vulkan-loader_1.3.216.0.bb RDEPENDS on or otherwise requires it)
ERROR: Required build target 'lmp-base-console-image' has no buildable providers.
Missing or unbuildable dependency chain was: ['lmp-base-console-image', 'weston', 'virtual/libgles2', 'mesa', 'vulkan-loader', 'libvulkan-imx']
```

### DIFF
--- a/meta-lmp-base/recipes-graphics/vulkan/vulkan-loader_1.3.204.1.bbappend
+++ b/meta-lmp-base/recipes-graphics/vulkan/vulkan-loader_1.3.204.1.bbappend
@@ -1,0 +1,1 @@
+RRECOMMENDS:${PN}:remove:imxgpu:mx8mm-nxp-bsp = "libvulkan-imx"


quaresmajose:
This should be needed as well on kirkstone as it is cherry picked https://github.com/Freescale/meta-freescale/commit/a7a6451991e5fef0168eb012a36b4ededb1c2776

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -347,6 +347,7 @@ OSTREE_SPLIT_BOOT:mx8mm-nxp-bsp = "${@bb.utils.contains('DISTRO_FEATURES', 'luks
 OSTREE_DEPLOY_USR_OSTREE_BOOT:mx8mm-nxp-bsp = "${@bb.utils.contains('DISTRO_FEATURES', 'luks', '1', '0', d)}"
 WKS_FILE:sota:mx8mm-nxp-bsp = "${@bb.utils.contains('DISTRO_FEATURES', 'luks', 'sdimage-imx8-spl-split-boot-sota.wks.in', 'sdimage-imx8-spl-sota.wks.in', d)}"
 ## iMX8MM EVK
+DISTRO_FEATURES:remove:imx8mm-lpddr4-evk = "vulkan"
```
MrCry0, line 350: I'd move it to the common imx8mm part and use the upstream classifier `mx8mm-nxp-bsp` - we can add support of other imx8mm boards in the future, so such generalization simplifies steps for adding such support.

quaresmajose: I agree with the more generic use of `mx8mm-nxp-bsp` machine override and the move to common imx8mm

quaresmajose: Fixed, thanks for point that.

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -338,6 +338,7 @@ PREFERRED_PROVIDER_u-boot-default-script:mx8mm-generic-bsp = "u-boot-base-scr"
 PREFERRED_PROVIDER_u-boot-default-script:sota:mx8mm-generic-bsp = "u-boot-ostree-scr-fit"
 KERNEL_IMAGETYPE:sota:mx8mm-generic-bsp = "fitImage"
 KERNEL_CLASSES:sota:mx8mm-generic-bsp = " kernel-lmp-fitimage "
+DISTRO_FEATURES:remove:mx8mm-nxp-bsp = "vulkan"
```
ricardosalveti, line 341: Hm, what is actually trying to build/bring vulkan on imx8mm?

DISTRO_FEATURES shouldn't cause the build to fail necessarily, as it is just a generic option for the distribution to support a functionality, and the bsp should be the one then deciding if there is a matching support on the driver level.

A better use at the bsp level is to leverage COMBINED_FEATURES instead, I would say.
```
COMBINED_FEATURES = "${@oe.utils.set_intersect('DISTRO_FEATURES', 'MACHINE_FEATURES', d)}"
COMBINED_FEATURES[vardeps] += "DISTRO_FEATURES MACHINE_FEATURES"
```

quaresmajose: > Hm, what is actually trying to build/bring vulkan on imx8mm?

lmp-xwayland distro has vulkan enabled in DISTRO_FEATURES 

```
DISTRO_FEATURES:append = " x11 wayland opengl vulkan"
```

https://github.com/foundriesio/meta-lmp/blob/main/meta-lmp-base/conf/distro/lmp-xwayland.conf


> DISTRO_FEATURES shouldn't cause the build to fail necessarily, as it is just a generic option for the distribution to support a functionality, and the bsp should be the one then deciding if there is a matching support on the driver level.

I agree that this should be handled in the bsp layer but is not the case in meta-freescale

ricardosalveti: It seems that this is what is causing the failure:
```
meta-freescale/recipes-graphics/vulkan/vulkan-loader_%.bbappend:RRECOMMENDS:${PN}:append:imxgpu = " libvulkan-imx"
```

If possible I would prefer fixing meta-freescale instead.

quaresmajose: As the bsp doesn't support vulkan on this machine (i.e: libvulkan-imx is not provided) the only solution I can see is using the vulkan cpu backend Lavapipe provided by mesa to continue using vulkan DISTRO_FEATURES

quaresmajose: please drop my last comment and I will try to fix the issue in meta-freescale

quaresmajose:
Reopen as it will unblock the bumping of meta-freescale https://github.com/foundriesio/lmp-manifest/pull/224. This PR is a workaround and can be dropped when a solution found upstream https://github.com/Freescale/meta-freescale/pull/1249.

MrCry0 (approved):

angolini:
I would wait until we have a good meta-freescale base, so I can rework the bump again.
It looks like there are more pieces other than the machine override (the downgrade in the vulkan version)
Anyway, thanks @quaresmajose for this PR :smile:

quaresmajose:
> the downgrade in the vulkan version

yeah, another PR will come up with the downgrade

quaresmajose:
kirkstone is now fixed https://github.com/Freescale/meta-freescale/pull/1250

PR #849, opened 2022-10-10
Use the lmp mirrors in PREMIRRORS and keep it in sync with MIRRORS

quaresmajose:
- base: lmp: adds lmp mirrors to PREMIRRORS

As we have the PREMIRRORS assigned with ??= the Yocto mirrors takes precedence so ours is disabled, change that and adds our mirrors.

- base: lmp: keep lmp PREMIRRORS in sync with MIRRORS


ricardosalveti (approved):
LGTM

PR #850, opened 2022-10-11
WIP: lmp-el2go-auto-register: fix PKCS#11 slot initialization

mwasilew:
This patch fixes the potential issues when running the auto-registration script on the device:

1. 1st slot gets overwritten If the 1st slot is not labeled 'aktualizr' and lmp-el2go-auto-register can't log into it with it's default PIN it would overwrite it. This might mean data loss for some othe deamon which initialized the slot

2. Objects not stored in proper slots When writing to pkcs#11 lmp-el2go-auto-register was not using the slot label consistently. This means that some objects were written in proper slot while some other not. This would only happen if the 1st slot (0x0) was initialized with default PIN.

3. has_labels method doesn't use slot label For this reason it would always try to access 1st slot (0x0) and ignore the label. If slot with label 'aktualizr' was initialized in different position this would mean the certificate/key objects would not be stored properly.

Signed-off-by: Milosz Wasilewski <milosz.wasilewski@foundries.io>

### DIFF
--- a/meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
+++ b/meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
@@ -4,6 +4,7 @@
 # Copyright (c) 2022, Foundries.io Ltd.
 
 from contextlib import contextmanager
+import re
 import os
 import logging
 import subprocess
@@ -18,6 +19,7 @@ PIN = os.environ.get("PKCS11_PIN", "87654321")
 SO_PIN = os.environ.get("PKCS11_SOPIN", "12345678")
 SOTA_DIR = os.environ.get("SOTA_DIR", "/var/sota")
 HANDLERS = os.environ.get("HANDLERS", "aws-iot,aktualizr-lite")
+TOKEN_LABEL = os.environ.get("TOKEN_LABEL", "aktualizr")
 
 REPO_ID = os.environ["REPOID"]
 
@@ -46,7 +48,7 @@ class Pkcs11:
             "--key-type=EC:prime256v1",
             f"--id={slot}",
             f"--label={label}",
-            "--token-label=aktualizr",
+            f"--token-label={TOKEN_LABEL}",
         ]
         run_quietly(args)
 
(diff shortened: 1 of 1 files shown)


mwasilew:
This is still WIP. When testing I discovered that aklite also has similar assumptions. Aklite needs to be fixed before this patch can be merged to avoid regressions.

meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
```diff
@@ -76,20 +83,59 @@ class Pkcs11:
...
+        slot_dict = cls._init_slot_dictionary(out.decode())
+        empty_slot = None
+        for slot_id, slot in slot_dict.items():
+            token_state = slot.get("token state")
+            if token_state == "uninitialized":
+                empty_slot = slot_id
+                break
+        if empty_slot is None:
+            # raise exception as there are no more available slots
+            raise Pkcs11Error("No free PKCS11 slots")
```
doanac, line 135: I don't think you are gaining anything from a custom error class. Seems like you can just throw a RuntimeError .

Other than that - I think this is beyond my understanding of PKCS and you should have the platform team review

mwasilew:
@rsalveti @ldts could you take a look?

meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
ricardosalveti: +1 for a generic error here.

meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
```diff
@@ -256,3 +302,4 @@ def main():
 
 if __name__ == "__main__":
     main()
+
```
ricardosalveti, line 305: Drop extra line.

ricardosalveti:
PKCS#11 logic seems correct. I know it is a bit of a pain to depend and parse the tooling output, but this should improve quite a bit from what we have now.

Ideally we should be using the pkcs11 interface directly, and same for lmp-device-register (https://github.com/foundriesio/lmp-device-register/issues/20), but this would be for another day.

meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
mwasilew: done

meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
mwasilew: done

mwasilew:
lmp-device-register suffers from the same issue as lmp-el2go-auto-register. It will initialize token in the 1st slot and potentially overwrite the token that is already there.

doanac:
> Ideally we should be using the pkcs11 interface directly

Also - the ssscli code turns out to be a python project. So we could probably use its lib directly and do some big cleanups here.

angolini:
What is the status of this PR?

doanac:
everything with regards to el2g can be deleted

PR #859, opened 2022-10-14
lmp-staging: serialize do_compile of heavy tasks

quaresmajose:
This is a work around to not compile rust and clang at same time as it is very huge tasks.

It uses the bitbake lockfiles to run the refered tasks one at a time. The bitbake UI can be misleading since it will inform you that the tasks are working at the same time, which is not true because only one is working and all the others are blocked.
I have a pending bitbake patch to improve this situation.

### DIFF
--- a/meta-lmp-base/classes/lmp-staging.bbclass
+++ b/meta-lmp-base/classes/lmp-staging.bbclass
@@ -17,6 +17,9 @@ python __anonymous() {
 
     if bb.data.inherits_class('go-mod', d):
         d.appendVarFlag('do_compile', 'network', '1')
+
+    if pn in ["clang", "rust"]:
+        d.appendVarFlag('do_compile', 'lockfiles', " ${TMPDIR}/lmp-hack-avoid-oom-do_compile.lock")
 }
 
 inherit ${INHERIT_KERNEL_MODSIGN}
--- a/meta-lmp-base/classes/lmp-staging.bbclass
+++ b/meta-lmp-base/classes/lmp-staging.bbclass
@@ -8,6 +8,8 @@
 INHERIT_KERNEL_MODSIGN = ""
 
 python __anonymous() {
+    pn = d.getVar('PN')
+
     if bb.data.inherits_class('module', d):
         d.appendVar('DEPENDS', ' virtual/kernel')
         if 'modsign' in d.getVar('DISTRO_FEATURES'):


meta-lmp-base/classes/lmp-staging.bbclass
```diff
@@ -15,6 +15,10 @@ python __anonymous() {
 
     if bb.data.inherits_class('go-mod', d):
         d.appendVarFlag('do_compile', 'network', '1')
+
+    pn = d.getVar('PN')
+    if pn in ["clang", "rust"]:
```
ricardosalveti, line 20: Will this catch native runs as well? Or is that already safe with the normal dependency chain locks?

quaresmajose: no, this only lock the target ones. but yes, the dependency chain already block the native ones.

`clang` needs `clang-cross-x86_64`  that pulls `clang-native`
```
bitbake-getvar -r clang DEPENDS
DEPENDS="pkgconfig-native cmake-native  binutils zlib zstd libffi libxml2 libxml2-native ninja-native swig-native python3-native  clang-cross-x86_64 python3 compiler-rt libcxx libedit libedit-native ncurses ninja-native"
```

`rust` needs `rust-native` and `clang-native`

```
bitbake-getvar -r rust DEPENDS
DEPENDS=" clang-cross-x86_64 virtual/libc  compiler-rt libcxx  file-native python3-native rust-llvm (=1.63.0) python3-native   rust-native libstd-rs clang-native compiler-rt-native libcxx-native"
```

So the order without this patch is:

1. `clang-native`
2. `rust-native`
3. `clang` and `rust` that can run at same time

angolini (approved):

meta-lmp-base/classes/lmp-staging.bbclass
```diff
@@ -8,6 +8,8 @@
 INHERIT_KERNEL_MODSIGN = ""
 
 python __anonymous() {
+    pn = d.getVar('PN')
```
MrCry0, line 11: I'd combine both commits in a single - introducing and using variables in the same commit is fine IMHO.

quaresmajose: This bitbake class has mainly workaround that will be reverted asap or when fixed upstream. In the first patch they went in the same commit, but have it in a separate patch add the possibility to revert one and not the other.
The `PN` is a very common variable so will always be there and meanwhile I already have new patches that makes uses  it.

ricardosalveti (approved):
LGTM

PR #868, opened 2022-10-20
luks2: support encryption in CI and online re-encryption on first boot

ricardosalveti:
Still not fully functional due https://gitlab.com/cryptsetup/cryptsetup/-/issues/780.

After that is fixed upstream, the only other piece missing would be the systemd job responsible for calling reencrypt --resume-only.

### DIFF
--- a/meta-lmp-base/recipes-support/luks-reencryption/luks-reencryption/luks-reencryption
+++ b/meta-lmp-base/recipes-support/luks-reencryption/luks-reencryption/luks-reencryption
@@ -7,7 +7,7 @@ set -e
 
 [ $(whoami) = "root" ] || { echo "E: You must be root" && exit 1; }
 
-DEVICE=$(lsblk -f | awk '/crypto_LUKS/ {print $1}' | awk '{sub(/^[^a-zA-Z]*/, ""); print}')
+DEVICE=$(lsblk -f | awk '/crypto_LUKS/ {print $1; exit}' | awk '{sub(/^[^a-zA-Z]*/, ""); print}')
 DEVICE="/dev/${DEVICE}"
 
 # Avoid using the PIN for OP-TEE supported PKCS#11 user authentication


meta-lmp-base/recipes-crypto/cryptsetup/cryptsetup_git.bb
```diff
@@ -0,0 +1,109 @@
+SUMMARY = "Manage plain dm-crypt and LUKS encrypted volumes"
```
quaresmajose, line 1: Carry cryptsetup with a bb append can be more adequate and easy to see the changes needed.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -140,6 +140,36 @@ IMAGE_CMD:ostree:append () {
...
+IMAGE_CMD:ota-ext4:append () {
+	if [ "${OSTREE_OTA_EXT4_LUKS}" = "1" ]; then
+		if [ -z "${OSTREE_OTA_EXT4_LUKS_PASSPHRASE}" ]; then
+			bbfatal "Unable to find passphrase for LUKS-based ota-ext4 (define OSTREE_OTA_EXT4_LUKS_PASSPHRASE)"
+		fi
+		cp ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.ota-ext4 ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.ota-ext4-orig
+
+		block_size=`dumpe2fs -h ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.ota-ext4 | grep "^Block size" | cut -d ':' -f 2 | tr -d ' '`
+		block_count=`dumpe2fs -h ${IMGDEPLOYDIR}/${IMAGE_NAME}${IMAGE_NAME_SUFFIX}.ota-ext4 | grep "^Block count" | cut -d ':' -f 2 | tr -d ' '`
+		luks_size=33554432 # 32m
```
Tim-Anderson, line 153: My only comment here is, do you expect to hard code this to 32 megabytes. and is that only for the initial image and will expand correct on first boot?

Tim-Anderson (commented):
Are you completely replacing the recipe because it is not correct in OE?

ricardosalveti:
> Are you completely replacing the recipe because it is not correct in OE?

Newer version, this is tracking git atm.

Once the required fixes land upstream I will update the recipe properly.

meta-lmp-base/classes/lmp.bbclass
ricardosalveti: Yes, this is the recommended amount for the luks2 metadata. When doing the resize on first boot it takes this into account as well (which is why we need to call cryptsetup resize as well).

Tim-Anderson (approved):
LGTM

ldts:
The problem is that systemd-cryptenroll for TPM2 does not generate the tpm2 secret that can be used for the LUKS re-encryption.
I'll send a systemd patch to fix this locally

ldts:
also, any problem with doing all the re-encryption form the initramfs? seems safer

ldts:
> The problem is that systemd-cryptenroll for TPM2 does not generate the tpm2 secret that can be used for the LUKS re-encryption. I'll send a systemd patch (just a few lines) + the initframds update to fix this locally

https://github.com/ldts/meta-lmp/commit/518812f127df2e11191370b65ceb0439ebd723bf

PR #869, opened 2022-10-21
Initial support for JP5

ricardosalveti:
Still not using systemd-boot by default, but required work for enabling it is also part of this PR.


meta-lmp-base/conf/distro/include/non-clangable.inc
```diff
@@ -99,6 +99,9 @@ TOOLCHAIN:pn-trusted-firmware-a = "gcc"
 
 # jetson-agx-xavier-devkit
 TOOLCHAIN:pn-cboot-t19x = "gcc"
```
quaresmajose, line 101: as cboot is been dropped we can drop this as well

quaresmajose (approved):
LGTM with a minor comment

MrCry0 (approved):
LGTM. (with jose's note)

meta-lmp-base/conf/distro/include/non-clangable.inc
ricardosalveti: yup, will remove.

ricardosalveti: done

quaresmajose (approved):
LGTM

PR #870, opened 2022-10-24
Garagesign: add retries and sleep configuration

quaresmajose:


meta-lmp-base/classes/image_types_ostree.bbclass
```diff
@@ -260,7 +261,12 @@ IMAGE_CMD:garagesign () {
                 push_success=1
                 break
             else
-                bbwarn "Push to garage repository has failed with errcode ${errcode}, retrying"
+                bbwarn "Push to garage repository has failed with errcode ${errcode}, retrying ${push_retries}/${GARAGE_PUSH_RETRIES}"
+                if [ "${GARAGE_PUSH_RETRIES_SLEEP}" -ne "0" ]; then
+                    sleep="$(($RANDOM%${GARAGE_PUSH_RETRIES_SLEEP}+1))""
+                    bbdebug 1 "Push to garage repository in ${sleep} seconds"
+                    sleep ${sleep}
+                fi
```
doanac, line 269: i don't have a strong opinion on this change. not sure if it will actually help in practice or not.

otherwise, LGTM

meta-lmp-base/classes/image_types_ostree.bbclass
```diff
@@ -178,6 +178,7 @@ IMAGE_CMD:ostreepush () {
 }
 
 GARAGE_PUSH_RETRIES ?= "3"
+GARAGE_PUSH_RETRIES_SLEEP ?= "0"
```
ricardosalveti, line 181: We can probably have this set to 10 by default, or similar.

meta-lmp-base/classes/image_types_ostree.bbclass
ricardosalveti: Yeah, seems ok, please propose upstream in meta-updater and we can move the review discussion there.

quaresmajose: I will do that after some tests on main-next

meta-lmp-base/classes/image_types_ostree.bbclass
quaresmajose: My ideia is propose this in meta-updater with the sleep disabled to don't change the current implementation.
We can change it after in our side but if upstream like to have this I can change the default to 10.

I am testing with 10 in main-next https://github.com/quaresmajose/meta-updater/commit/06d1404743008ea868812ac2b05fc3af997a3094

PR #885, opened 2022-11-03
Switch u-boot imx-2022.04 to 2022.04+lf-5.15.52-2.1.0-fio

MrCry0:


meta-lmp-bsp/recipes-bsp/u-boot/u-boot-fio/imx8mm-lpddr4-evk/lmp.cfg
```diff
@@ -110,3 +110,4 @@ CONFIG_POWER_I2C=y
 # CONFIG_VIDEO_LINK is not set
 # CONFIG_DISTRO_DEFAULTS is not set
 # CONFIG_SHA_HW_ACCEL is not set
+# CONFIG_SET_DFU_ALT_INFO is not set
```
ricardosalveti, line 113: We will need a similar change for sec and others, might be better to fix this in u-boot itself.

Tim-Anderson (approved):
LGTM

meta-lmp-bsp/recipes-bsp/u-boot/u-boot-fio/imx8mm-lpddr4-evk/lmp.cfg
ricardosalveti: I see what happened, the standard config this is fine because EFI capsule update is also defined, which has a weak definition for the function. I think we should just disable other configs like DFU, let me have a quick look.

ricardosalveti:
Closing in favor of https://github.com/foundriesio/meta-lmp/pull/888.

Doing a wider config cleanup there, taking your u-boot update patch.

meta-lmp-bsp/recipes-bsp/u-boot/u-boot-fio/imx8mm-lpddr4-evk/lmp.cfg
MrCry0: The problem is this can't be reproduced in the pure u-boot. If CONFIG_SET_DFU_ALT_INFO  is not in defconfig, it stays disabled. But when the recipe u-boot-fio builds, this option turns out enabled. I'll investigate this behavior tomorrow.

ricardosalveti: Happens with our fragment and not in pure u-boot as the EFI options will add a weak function definition and won't cause the problem. The issue is related to the removals we had + additions on the latest bsp.

Which is why I just disabled the new options in #888, no need to spend more time on this I would say.

PR #889, opened 2022-11-04
bsp: lmp-machine-custom: imx: drop preferred rev for imx-gpu-viv

ricardosalveti:
No need to set preferred version for kernel-module-imx-gpu-viv as the one provided by the BSP is now newer.


quaresmajose (approved):
LGTM, I have updated the version in main-next but is better to drop it as only a single version is provided

MrCry0 (approved):
LGTM

angolini (approved):

PR #892, opened 2022-11-04
base: lmp: remove image-buildinfo as we don't use it anymore

quaresmajose:
This also adds a dependencie BBLAYERS that rebuild many recipes from scracth when we cange the build dir.


ricardosalveti (approved):
LGTM

PR #900, opened 2022-11-09
base: images: install dtc package

igoropaniuk:
Install dtc package by default for all images, as this likely to be used on all platforms for exporting device tree during debugging.

### DIFF
--- a/meta-lmp-base/recipes-samples/images/lmp-feature-debug.inc
+++ b/meta-lmp-base/recipes-samples/images/lmp-feature-debug.inc
@@ -9,4 +9,5 @@ CORE_IMAGE_BASE_INSTALL += " \
     minicom \
     devmem2 \
     curl \
+    dtc \
 "


quaresmajose (approved):
LGTM

MrCry0:
We have `./meta-lmp-base/recipes-samples/images/lmp-feature-debug.inc`. Wouldn't that be the better place for `dtc` package?

ricardosalveti:
> We have `./meta-lmp-base/recipes-samples/images/lmp-feature-debug.inc`. Wouldn't that be the better place for `dtc` package?

Yeah, I would also prefer having dtc as part of lmp-feature-debug.inc.

Just because changing common will affect every user of lmp (on every factory). If needed we can include lmp-feature-debug.inc as part of lmp-base-console-image by default, since this image is only used by the public factory.

quaresmajose:
> > We have `./meta-lmp-base/recipes-samples/images/lmp-feature-debug.inc`. Wouldn't that be the better place for `dtc` package?
> 
> Yeah, I would also prefer having dtc as part of lmp-feature-debug.inc.
> 
> Just because changing common will affect every user of lmp (on every factory). If needed we can include lmp-feature-debug.inc as part of lmp-base-console-image by default, since this image is only used by the public factory.

This look a better approach imo too

igoropaniuk:
done, moved to `lmp-feature-debug.inc`

MrCry0 (approved):
LGTM

Tim-Anderson (approved):
LGTM

ricardosalveti (approved):
LGTM

PR #902, opened 2022-11-09
base: layer.conf: exclude os-release in the siggen as it is abisafe

quaresmajose:
The recipe os-release only install a file on target so we cat exclude it from the siggen and avoid rebuild packages that releies on the files they provided by the os-release.

Currently one bigside effects of this issue is rebuilding systemd and consequently everything that depends on that.

A list of recipes that are completely stable and will never change. The ABI for the recipes in the list are presented by output from the tasks run to build the recipe. Use of this variable is one way to remove dependencies from one recipe on another that affect task signatures and thus force rebuilds when the recipe changes.
https://docs.yoctoproject.org/ref-manual/variables.html?highlight=siggen_exclude_safe_recipe_deps#term-SIGGEN_EXCLUDERECIPES_ABISAFE

### DIFF
--- a/meta-lmp-base/conf/layer.conf
+++ b/meta-lmp-base/conf/layer.conf
@@ -31,3 +31,6 @@ BBFILES_DYNAMIC += " \
     tpm-layer:${LAYERDIR}/dynamic-layers/tpm-layer/*/*/*.bb \
     tpm-layer:${LAYERDIR}/dynamic-layers/tpm-layer/*/*/*.bbappend \
 "
+
+# A list of recipes that are completely stable and will never change.
+SIGGEN_EXCLUDERECIPES_ABISAFE += "os-release"


quaresmajose:
This patch needs to be proposed upstream in oe-core.

ricardosalveti:
> The recipe os-release only install a file on target so we cat exclude it from the siggen and avoid rebuild packages that releies on the files they provided by the os-release.

It just installs one file but that file is dynamically generated based on certain variables that are not necessarily stable (e.g. upstream uses DATETIME).

In our case the content of os-release will change on every CI build due IMAGE_ID and IMAGE_VERSION, so in that context it is expected for it to cause this sstate chain to be invalidated.

quaresmajose:
> > The recipe os-release only install a file on target so we cat exclude it from the siggen and avoid rebuild packages that releies on the files they provided by the os-release.
> 
> It just installs one file but that file is dynamically generated based on certain variables that are not necessarily stable (e.g. upstream uses DATETIME).
> 
> In our case the content of os-release will change on every CI build due IMAGE_ID and IMAGE_VERSION, so in that context it is expected for it to cause this sstate chain to be invalidated.

The `os-release` will change on every CI build as before and it will be installed as well as before in the target image.
But what this patch does is not triggering a systemd rebuild for every CI build. The systemd currently rebuild because they have `RRECOMMENDS:${PN} = "os-release"` and this fire-up a systemd rebuild for every `os-release` with is not correct because it is a runtime dependence and the output of systemd build is the same as before.

ricardosalveti:
> > > The recipe os-release only install a file on target so we cat exclude it from the siggen and avoid rebuild packages that releies on the files they provided by the os-release.
> > 
> > 
> > It just installs one file but that file is dynamically generated based on certain variables that are not necessarily stable (e.g. upstream uses DATETIME).
> > In our case the content of os-release will change on every CI build due IMAGE_ID and IMAGE_VERSION, so in that context it is expected for it to cause this sstate chain to be invalidated.
> 
> The `os-release` will change on every CI build as before and it will be installed as well as before in the target image. But what this patch does is not triggering a systemd rebuild for every CI build. The systemd currently rebuild because they have `RRECOMMENDS:${PN} = "os-release"` and this fire-up a systemd rebuild for every `os-release` with is not correct because it is a runtime dependence and the output of systemd build is the same as before.

Yup, makes sense, I had the impression this was a harder dependency on systemd but everything gets parsed at runtime, so it makes sense to mark it abi safe.

ricardosalveti (approved):
LGTM

MrCry0 (commented):
Good catch! LGTM.

PR #916, opened 2022-11-14
Fix kernel initrd images rebuild on CI

quaresmajose:
- Fix sstate-cache invalidation of the kernel initrd images

### DIFF
--- a/meta-lmp-base/conf/distro/include/lmp.inc
+++ b/meta-lmp-base/conf/distro/include/lmp.inc
@@ -42,6 +42,9 @@ DISTRO_FEATURES_DEFAULT = "acl argp bluetooth ext2 ipv4 ipv6 largefile usbgadget
 DISTRO_FEATURES_BACKFILL_CONSIDERED = "pulseaudio ldconfig"
 DISTRO_FEATURES:append = " pam usrmerge virtualization alsa"
 
+# List of distro-specific packages to add to all images with packagegroup-distro-base
+DISTRO_EXTRA_RRECOMMENDS += "base-files-issue"
+
 # Default IMA policy (tcb)
 IMA_POLICY ?= "ima-policy-tcb"
 
--- a/meta-lmp-base/recipes-core/base-files/base-files-issue.bb
+++ b/meta-lmp-base/recipes-core/base-files/base-files-issue.bb
@@ -1,0 +1,47 @@
+FILESEXTRAPATHS:prepend := ":${COREBASE}/meta/recipes-core/base-files/base-files:"
+
+def get_pv(d):
+    import re
+    corebase = d.getVar('COREBASE')
+    bb_dir = os.path.join(corebase, 'meta', 'recipes-core', 'base-files')
+    if os.path.isdir(bb_dir):
+        re_bb_name = re.compile(r"base-files_([0-9.]*)\.bb")
+        for bb_file in os.listdir(bb_dir):
+            result = re_bb_name.match(bb_file)
+            if result:
+                return result.group(1)
+    bb.fatal("Cannot find base-files recipe in %s" % bb_dir)
+
+PV := "${@get_pv(d)}"
+
+require recipes-core/base-files/base-files_${PV}.bb
+
+SUMMARY += " (only contains /etc/issue*)"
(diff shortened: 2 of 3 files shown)


quaresmajose:
Another options can be disabling this only in the initramfs images

ricardosalveti:
> Another options can be disabling this only in the initramfs images

That would be preferred.

quaresmajose:
> > Another options can be disabling this only in the initramfs images
> 
> +1

I have been testing this and it be much complicated to solve. It is possible to not installing the `/etc/issue` on the initramfs with some modifications on the `base-files.bbappend`, this is what is present on the current state of the patch.

Anyway the target initramfs image signature still depend on the `base-files` recipe because they are used/installed buy the image. So any change on `base-files` recipe triggers a rebuild of the initramfs.

I will revert it back to the original version of the patch:
```
FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"

+BASEFILESISSUEINSTALL = ""
```

quaresmajose:
The real source of true is still the `/etc/os-release` so we didn't lose much with this drop that is basically the lmp prompt
```
Linux-microPlatform 4.0.10-1-90
```

quaresmajose:
This PR fixes only the initrd images rebuilding but as a side effect the kernel is not rebuild anymore because of this too.

MrCry0 (approved):
With the minor note, LGTM

angolini (approved):

meta-lmp-base/conf/distro/include/lmp.inc
```diff
@@ -39,6 +39,9 @@ DISTRO_FEATURES_DEFAULT = "acl argp bluetooth ext2 ipv4 ipv6 largefile usbgadget
 DISTRO_FEATURES_BACKFILL_CONSIDERED = "pulseaudio ldconfig"
```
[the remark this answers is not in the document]
quaresmajose: fixed! thanks

quaresmajose:
Validated on https://github.com/foundriesio/lmp-manifest/pull/348

igoropaniuk (approved):
LGTM

MrCry0:
@ricardosalveti .

quaresmajose:
@MrCry0 this is not ready yet and needs more testing

ricardosalveti:
@quaresmajose back to what we discussed, the 2 signing variables to track are actually the ones for UEFI secure boot (UEFI_SIGN_KEYDIR/DB.key, similar how intel did on meta-intel) and fitimage signing (UBOOT_SIGN_KEYDIR/UBOOT_SIGN_KEYNAME).

If we cover both (u-boot fitimage and uefi based boot) then we're good here.

quaresmajose:
Thanks, I'm back to this today and hope I can adress and fix all pending issue

quaresmajose:
Depends on https://github.com/foundriesio/meta-lmp/pull/1311

meta-lmp-base/recipes-core/base-files/base-files-issue.bb
```diff
@@ -0,0 +1,45 @@
...
+"
+
+pkg_preinst:${PN} () {
+    :
+}
+
+do_install () {
+    install -d ${D}${sysconfdir}
+    do_install_basefilesissue
+    return 0
```
ricardosalveti, line 34: is return 0 needed here?

quaresmajose: I have no ideia how this got here, I don't see the reason for this.
But will do some tests because this PR It's been a while and I don't remember.

quaresmajose: The reason fo this is because we only need to call the `do_install_basefilesissue` as anyother files was already been installed.

This snip from the main recipe will be executed as well because of the `require recipes-core/base-files/base-files_${PV}.bb` and we don't carry the `${WORKDIR}/nsswitch.conf`  so it will fail.
```
do_install:append:libc-glibc () {
	install -m 0644 ${WORKDIR}/nsswitch.conf ${D}${sysconfdir}/nsswitch.conf
}
```

quaresmajose: Added a comment about this

ricardosalveti (approved):
LGTM

PR #920, opened 2022-11-16
os-release: bypass the allarch_package_arch_handler

quaresmajose:
This is a fixup for 96bc0769 os-release: set PACKAGE_ARCH machine dependent

To set the PACKAGE_ARCH we need to bypass the allarch_package_arch_handler from allarch.bbclass otherwise they will set the PACKAGE_ARCH=all regardless of what is defined elsewhere.

### DIFF
--- a/meta-lmp-base/recipes-core/os-release/os-release.bbappend
+++ b/meta-lmp-base/recipes-core/os-release/os-release.bbappend
@@ -17,6 +17,11 @@ IMAGE_ID = "${LMP_FACTORY_IMAGE}"
 IMAGE_VERSION = "${H_BUILD}"
 
 PACKAGE_ARCH = "${MACHINE_ARCH}"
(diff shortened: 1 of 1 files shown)


angolini (approved):

MrCry0 (approved):
LGTM

quaresmajose:
> Should we expect less reuse of sstate cache once this is applied? (I mean, exactly on the transition)

As this recipe use the git commit hash of lmp-manifest in `VERSION`, `VERSION_ID` and `PRETTY_NAME` they rebuild the package from scratch on every change on this variables. In conclusion, on CI they don't sstate cache at all for this package.

This change is useful for debugging, especially locally, as it will store one sstate-cache per machine instead of only one.
But I am thinking in exclude this package totally from sstate-cache as in reality it's only a waste of time, perhaps in a future interaction.

PR #927, opened 2022-11-18
base: optee-os-fio: deploy build configuration

MrCry0:
Publish the final configuration used as part of the recipe build into the LmP deploy folder.

Having the conf.mk file published is quite useful when debugging configuration issues without having to build it locally.

### DIFF
--- a/meta-lmp-base/recipes-security/optee/optee-os-fio.inc
+++ b/meta-lmp-base/recipes-security/optee/optee-os-fio.inc
@@ -71,6 +71,11 @@ do_deploy() {
 
     # Link tee.bin so it can be consumed by recipes such as imx-boot
     ln -sf optee/tee-pager_v2.bin ${DEPLOYDIR}/tee.bin
+
+    # Deploy optee_os config used during build
+    install -D -m 644 ${B}/conf.mk ${DEPLOYDIR}/${PN}-config-${MACHINE}-${PV}-${PR}
+    ln -sf ${PN}-config-${MACHINE}-${PV}-${PR} ${DEPLOYDIR}/${PN}-config-${MACHINE}
+    ln -sf ${PN}-config-${MACHINE}-${PV}-${PR} ${DEPLOYDIR}/${PN}-config
 }
 
 addtask deploy before do_build after do_install


meta-lmp-base/recipes-security/optee/optee-os-lmp.inc
```diff
@@ -0,0 +1,9 @@
+# LmP common include for optee_os
+
+# Also deploy optee_os config used during build
+do_deploy:append() {
+    install -D -m 644 ${B}/conf.mk ${DEPLOYDIR}/${PN}-config-${MACHINE}-${PV}-${PR}
+    cd ${DEPLOYDIR}
```
ricardosalveti, line 6: Since it is just 2 lines, avoid 'cd' and just link using ${DEPLOYDIR} on both lines.

meta-lmp-base/recipes-security/optee/optee-os-fio.inc
```diff
@@ -12,6 +12,7 @@ DEPENDS = "python3-pyelftools-native dtc-native"
 DEPENDS:append:toolchain-clang = " compiler-rt"
 
 require optee-fio.inc
+include optee-os-lmp.inc
```
ricardosalveti, line 15: One thing to check, this will end up generating configs for both optee-os-fio and optee-os-tadevkit right?

Not sure we can avoid that, unless we check for tadevkit in PN.

MrCry0: @ricardosalveti Sure, tadevkit. I'm checking this.

quaresmajose: I think this is not a problem because the config is deployed as `${DEPLOYDIR}/${PN}-config-${MACHINE}-${PV}-${PR}` so `PN` is already there.

ricardosalveti: Yup, won't conflict, but there is no need to have the ones from tadevkit (but ok if avoiding it makes the logic more complicated than what already is).

MrCry0: > One thing to check, this will end up generating configs for both optee-os-fio and optee-os-tadevkit right?

This doesn't affect `optee-os-tadevkit` because that recipe redefines `deploy`:
```
do_deploy() {
    echo "Do not inherit do_deploy from optee-os."
}
```

meta-lmp-base/recipes-security/optee/optee-os-lmp.inc
MrCry0: Done.

meta-lmp-base/recipes-security/optee/optee-os-lmp.inc
```diff
@@ -0,0 +1,9 @@
...
+    ln -sf ${PN}-config-${MACHINE}-${PV}-${PR} ${PN}-config-${MACHINE}
+    ln -sf ${PN}-config-${MACHINE}-${PV}-${PR} ${PN}-config
+}
```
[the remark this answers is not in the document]
MrCry0: Done

ricardosalveti (approved):
LGTM

PR #966, opened 2022-12-04
[RFC] Lmp-common.cfg

igoropaniuk:
Introduce lmp-common.cfg, which contains common options for all machines for lmp distro. This approach helps to avoid duplication of Kconfig options and is less prone to errors (lower possibility to miss vital config option).

Some obvious common options are moved to that file(analyzing number of occurrences in `cat ./u-boot-fio/*/lmp.cfg | sort | uniq -c | sed -n -e 's/^ *<number_of_occurences> \(.*\)/\1/p'`)

Same is planned to be done for lmp-base and lmp-mfgtool configs


meta-lmp-base/recipes-bsp/u-boot/u-boot-fio/lmp-common.cfg
```diff
@@ -0,0 +1,8 @@
+CONFIG_FIT=y
+CONFIG_USE_BOOTCOMMAND=y
+CONFIG_BOOTDELAY=-2
```
MrCry0, line 3: I'm not sure about this option. When debugging, I often set it to a positive value for a particular board. Others may do the same. Having this option in a machine config file would be more convenient.

LGTM for the rest.

angolini (approved):

angolini:
I was expecting to see a bigger `lmp-common.cfg`. Are you planning to expand it in the future?

igoropaniuk:
> I was expecting to see a bigger lmp-common.cfg. Are you planning to expand it in the future?

Yes, for now I moved some obvious non-intrusive config options, moving others require testing for side effects

PR #981, opened 2022-12-12
bsp: u-boot-xlnx: k26: disable redundant env

ricardosalveti:
Redundant environment is not properly supported yet by our default boot.cmd and userspace environment manipulation logic, so just disable it for now.

Same is already done for uz.


igoropaniuk (approved):
LGTM

PR #983, opened 2022-12-12
bsp: optee-os-fio: apalis-imx6: increase ddr size

ricardosalveti:
Increase ddr size used by OP-TEE to 2GB in order to cover both the 1GB and 2GB modules.

This is required in order for the shared memory used by the kernel to be compatible with the optee memory mapping range.


ricardosalveti:
With this change I'm able to run xtests on apalis-imx6 (2GB):

```
+-----------------------------------------------------
29844 subtests of which 0 failed
123 test cases of which 0 failed
0 test cases were skipped
TEE test application done!
```

MrCry0 (approved):
LGTM

quaresmajose (approved):
LGTM

PR #993, opened 2022-12-20
base: rc: docker: Patch docker to fsync layer

mike-sul:
Signed-off-by: Mike <mike.sul@foundries.io>

### DIFF
--- a/meta-lmp-base/recipes-containers/docker/files/0001-Fsync-layer-once-extracted-to-file-system.patch
+++ b/meta-lmp-base/recipes-containers/docker/files/0001-Fsync-layer-once-extracted-to-file-system.patch
@@ -1,0 +1,37 @@
+From aef0dbdaeef9a2e56f0df3a9dfa41d4cdb88f108 Mon Sep 17 00:00:00 2001
+From: Mike Sul <mike.sul@foundries.io>
+Date: Tue, 20 Dec 2022 11:45:13 +0100
+Subject: [PATCH] Fsync layer once extracted to file system
+
+- Sync an overall file system just after layer unpack while the control
+  flow is within the layer "change root".
+
+Signed-off-by: Mike <mike.sul@foundries.io>
+---
+ pkg/chrootarchive/diff_unix.go | 3 +++
+ 1 file changed, 3 insertions(+)
+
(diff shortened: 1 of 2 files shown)


mike-sul:
@doanac @mike-scott @ricardosalveti This one works without visible performance impact.

meta-lmp-base/recipes-containers/docker/files/0001-Fsync-layer-once-extracted-to-file-system.patch
```diff
@@ -0,0 +1,42 @@
...
+ 	"path/filepath"
+ 	"runtime"
+ 
+@@ -77,6 +79,9 @@ func applyLayer() {
+ 		fatal(err)
+ 	}
+ 
++	if syncErr := exec.Command("/bin/sync").Run(); syncErr != nil {
++		logrus.Errorf(">>> FIO: failed to sync the extracted layer directory")
++	}
```
doanac, line 36: Isn't calling https://pkg.go.dev/syscall#Sync the same thing w/o the fork/exec and chance of an error?

mike-sul: I think so, however `syscall.Sync()` doesn't return any error. I suppose it's not really important to catch the "sync" error since it's kind of best effort anyway.
I just wanted to make sure that the "sync" is actually executed without any error and check how it impact `docker pull` performance.

mike-sul: Switched to `syscall.Sync()`, +- the same performance.

I think it's not important to catch the sync error, anyway the log output if printed in this spot is not visible because it's executed in "change root/root jail" env and its stderr is printed only if an overall extraction process exits with non-zero.

doanac: this looks good - my point about catching the error was that there isn't one - the error fork/exec would produce would only be if /bin/sync didn't exist or something.

doanac (approved):

ricardosalveti (approved):
LGTM

PR #995, opened 2022-12-27
bsp: linux-lmp: apalis-imx6: fix HDMI interface

MrCry0:
Add a patch to make HDMI interface work.

Note, this patch disables a lcdif panel. If the panel is connected to a board, it is safe to enable it back - in this case, it won't break other interfaces.


quaresmajose (approved):
LGTM

angolini (approved):
What is the reason for adding this patch in the recipe and not merged into the Linux repository?

I'm only curious, the PR LGTM

MrCry0:
> What is the reason for adding this patch in the recipe and not merged into the Linux repository?
> 
> I'm only curious, the PR LGTM

@angolini we use a pure repo for mainline Linux kernel + patches. I presume it's because of the few numbers of our kernel patches. And this patch shouldn't be upstreamed as well.

ricardosalveti (approved):
LGTM

PR #998, opened 2023-01-04
Drop mxm-mwifiex-setup recipe

quaresmajose:
We can use yocto variables for this propose:
- https://docs.yoctoproject.org/ref-manual/variables.html?highlight=kernel_module_autoload#term-KERNEL_MODULE_AUTOLOAD
- https://docs.yoctoproject.org/ref-manual/variables.html?highlight=kernel_module_autoload#term-KERNEL_MODULE_PROBECONF


MrCry0 (approved):
LGTM

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -373,6 +373,9 @@ OSTREE_BOOTLOADER:imx8mm-lpddr4-evk-ebbr ?= "grub"
 IMAGE_EFI_BOOT_FILES:sota:imx8mm-lpddr4-evk-ebbr ?= "${@make_efi_dtb_boot_files(d)}"
 WKS_FILE:sota:imx8mm-lpddr4-evk-ebbr ?= "efidisk-sota.wks.in"
 # iMX8MM EVKB revision
+KERNEL_MODULE_AUTOLOAD:imx8mm-lpddr4-evk-ebbr += "moal"
+KERNEL_MODULE_PROBECONF:imx8mm-lpddr4-evk-ebbr += "moal"
```
angolini, line 377: Is there a reason to add this driver only for the ebbr variation of the board?

angolini (commented):
Do you know if the KERNEL_MODULE_* variables also deal with firmware?

I think I still prefer to keep this as a MACHINE_FEATURE

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
ricardosalveti: Right, it shouldn't be ebbr specific.

`VARIABLE:board += ""`

IIRC this will also potentially cause issues with the assignment (var ?=, var:board = ""), which is why we usually just append here instead.

quaresmajose: > Is there a reason to add this driver only for the ebbr variation of the board?

This replace the previous `MACHINE_FEATURES:append:imx8mm-lpddr4-evk = " mxm-mwifiex-load"` that uses the `imx8mm-lpddr4-evk` machine override, a few lines below there is another but now for `imx8mp-lpddr4-evk`.

quaresmajose: > IIRC this will also potentially cause issues with the assignment (var ?=, var:board = ""), which is why we usually just append here instead.

Changed to use the append

quaresmajose:
> Do you know if the KERNEL_MODULE_* variables also deal with firmware?

I think not, this variables is handled on `kernel-module-split.bbclass` that don't know nothing about firmware.

> 
> I think I still prefer to keep this as a MACHINE_FEATURE

Is this configuration to auto load the kernel module still required? Why the `moal` can't be probed by the kernel and loaded by udev? I don't have any board to test this.

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
quaresmajose: > > Is there a reason to add this driver only for the ebbr variation of the board?
> 
> This replace the previous `MACHINE_FEATURES:append:imx8mm-lpddr4-evk = " mxm-mwifiex-load"` that uses the `imx8mm-lpddr4-evk` machine override, a few lines below there is another but now for `imx8mp-lpddr4-evk`.

misunderstood this part and ebbr variation is my fault! now fixed

quaresmajose:
> As far as I understand, `MACHINE_FEATURES` describes the hardware. There are firmwares that are only installed when some tag is present in the `MACHINE_FEATURES`
> 
> As far as I understand, moal module is a generic wifi driver, and I would say it is a common dependency for other drivers.
> 
> So I understand we can configure the moal driver probe using the OE variable, but I would not remove the MACHINE_FEATURES.
> 
> Are you facing some log errors due to changes in some firmware due to the latest meta-freescale bump, @quaresmajose ?

The MACHINE_FEATURES have now all the pieces on meta-freescale and we have our adaptation in https://github.com/foundriesio/lmp-manifest/pull/277 so lmp is aligned.

This PR is for auto load the kernel modules in this two machines, I can change it to be done with a MACHINE_FEATURES.
My question is why we need this auto load and modprobe config in LmP and meta-freescale machines works out of the box?

angolini:
I'm not entirely aware of the context of this PR. I thought it was only an enhancement, not a bug fix. I think the suggestion was to have the OE variables ready to load the module instead of something else.

We can discuss this more offline, but I would need to understand what is happening.

quaresmajose:
> I'm not entirely aware of the context of this PR. I thought it was only an enhancement, not a bug fix. I think the suggestion was to have the OE variables ready to load the module instead of something else.
> 
> We can discuss this more offline, but I would need to understand what is happening.

and you are right, this is a different way of doing the same thing and is not a bug. we can talk offline

MrCry0:
> > I'm not entirely aware of the context of this PR. I thought it was only an enhancement, not a bug fix. I think the suggestion was to have the OE variables ready to load the module instead of something else.
> > We can discuss this more offline, but I would need to understand what is happening.
> 
> and you are right, this is a different way of doing the same thing and is not a bug. we can talk offline

@angolini has a reason. I agree it would be better to keep a machine feature, just replacing the whole recipe with OE variables. So we can drop our conf files and replace the whole recipe `mxm-mwifiex-setup` with:

```
KERNEL_MODULE_AUTOLOAD:append = " moal"
KERNEL_MODULE_PROBECONF:append = " moal"
module_conf_moal = "options moal mod_para=nxp/wifi_mod_para.conf drvdbg=0x6"
```

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -373,7 +373,9 @@ OSTREE_BOOTLOADER:imx8mm-lpddr4-evk-ebbr ?= "grub"
 IMAGE_EFI_BOOT_FILES:sota:imx8mm-lpddr4-evk-ebbr ?= "${@make_efi_dtb_boot_files(d)}"
 WKS_FILE:sota:imx8mm-lpddr4-evk-ebbr ?= "efidisk-sota.wks.in"
 # iMX8MM EVKB revision
-MACHINE_FEATURES:append:imx8mm-lpddr4-evk = " mxm-mwifiex-load"
+KERNEL_MODULE_AUTOLOAD:append:imx8mm-lpddr4-evk = " moal"
+KERNEL_MODULE_PROBECONF:append:imx8mm-lpddr4-evk = " moal"
+module_conf_moal:imx8mm-lpddr4-evk = "options moal mod_para=nxp/wifi_mod_para.conf drvdbg=0x6"
```
ricardosalveti, line 378: Would be good to confirm indeed if this customization for mod_para is still needed as it seems it is not required anymore by looking at the BSP (maybe the default path it looks now is nxp/wifi_mod_para.conf?).

ricardosalveti: And also if autoloading is still required as well, @MrCry0 should be able to confirm.

quaresmajose:
Convert this to WIP because the variable KERNEL_MODULE_AUTOLOAD and KERNEL_MODULE_PROBECONF in yocto needs be adapted because they use "/etc" hard-coded in kernel-module-split.bbclass and we need to have this in ${libdir}

quaresmajose:
> Convert this to WIP because the variable KERNEL_MODULE_AUTOLOAD and KERNEL_MODULE_PROBECONF in yocto needs be adapted because they use "/etc" hard-coded in kernel-module-split.bbclass and we need to have this in ${libdir}

I have a patch prepared for oe-core so i will merge this in main-next to test

PR #1002, opened 2023-01-09
Tegra initrd flasher fixes

quaresmajose:


quaresmajose:
These fixes should unblock https://github.com/foundriesio/lmp-manifest/pull/278

meta-lmp-base/recipes-sota/ostree-kernel-initramfs/ostree-kernel-initramfs_%.bbappend
```diff
@@ -17,3 +17,5 @@ do_install:append() {
 
 INITRAMFS_RECOVERY_IMAGE ?= ""
 do_install[depends] += "virtual/kernel:do_deploy ${@['${INITRAMFS_RECOVERY_IMAGE}:do_image_complete', ''][d.getVar('INITRAMFS_RECOVERY_IMAGE') == '']}"
+
+INITRAMFS_FSTYPES:remove:tegra = "cpio.gz.cboot"
```
ricardosalveti, line 21: The problem on moving to a specific recipe is that the general setting of INITRAMFS_FSTYPES via append on the bsp layer influences every initrd that gets generated by the build, and why we removed on lmp-machine-custom (via remove).

The issue:
```
meta-tegra/conf/machine/include/tegra234.inc:INITRAMFS_FSTYPES:append = "${@'' if d.getVar('INITRAMFS_IMAGE_BUNDLE') == '1' else ' cpio.gz.cboot cpio.gz.cboot.bup-payload'}"
meta-tegra/conf/machine/include/tegra194.inc:INITRAMFS_FSTYPES:append = "${@'' if d.getVar('INITRAMFS_IMAGE_BUNDLE') == '1' else ' cpio.gz.cboot cpio.gz.cboot.bup-payload'}"
```

This forces this variable to be expanded on every recipe using it.

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -608,7 +608,7 @@ DISTRO_FEATURES:append:tegra = " opengl"
 DISTRO_FEATURES_DEFAULT:remove:tegra = "modsign"
 PREFERRED_PROVIDER_virtual/kernel:tegra ?= "linux-tegra"
 IMAGE_TEGRAFLASH_FS_TYPE:sota:tegra = "ota-ext4"
-INITRAMFS_FSTYPES:remove:tegra = "cpio.gz.cboot cpio.gz.cboot.bup-payload"
+INITRAMFS_FSTYPES:remove:tegra = "cpio.gz.cboot.bup-payload"
```
ricardosalveti, line 611: We can probably let as before here and create a tegra-initrd-flash-initramfs.bbappend adding what is missing.

quaresmajose: The issue with the `tegra-initrd-flash-initramfs` is because the `image_types_tegra.bbclass` expect a `*.cboot` initrd but as we remove here is not possible to add it again.
To add the `cpio.gz.cboot` only for the `tegra-initrd-flash-initramfs` we need a new variable here
```
INITRAMFS_FSTYPES_REMOVE = "cpio.gz.cboot cpio.gz.cboot.bup-payload"
INITRAMFS_FSTYPES:remove:tegra = "$INITRAMFS_FSTYPES_REMOVE"
```
and on the tegra-initrd-flash-initramfs.bbappend we override to don't drop the `*.cboot` initrd
```
INITRAMFS_FSTYPES_REMOVE = "cpio.gz.cboot.bup-payload"
```

quaresmajose:
Another way can be change the assignment upstream to use a weak `IMAGE_TEGRAFLASH_INITRD_FLASHER ?=`  and in lmp change it to  to use the default `cpio.gz` but don't know if this will work.

https://github.com/OE4T/meta-tegra/commit/40b259e8d84cd1d2c0b54d246529572c0e6f3ed0#diff-eba3b42a4b688730e5000092bb2b752a86b8bd73170e271e200f6f5bca97dbb9R74

meta-lmp-base/recipes-sota/ostree-kernel-initramfs/ostree-kernel-initramfs_%.bbappend
quaresmajose: yup, and maybe the `ostree-kernel-initramfs` is the reason we need to remove this types on lmp-machine-custom as this recipe refuses to build with it.

ricardosalveti: Right, the reason is that in LmP we don't want to build any special type of initrd for the normal rootfs, since we're booting either via UEFI or u-boot, and on both the normal type is already enough.

ricardosalveti:
> Another way can be change the assignment upstream to use a weak `IMAGE_TEGRAFLASH_INITRD_FLASHER ?=` and in lmp change it to to use the default `cpio.gz` but don't know if this will work.

I would also check with upstream if this override can be set without append, as that is what causes the issue on the first place (as append ends up affecting every recipe).

quaresmajose:
It builds with our initrd type so I will change this PR to for that 
```
diff --git a/meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc b/meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
index 45d038d0..85c5ba57 100644
--- a/meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
+++ b/meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
@@ -609,6 +609,7 @@ DISTRO_FEATURES_DEFAULT:remove:tegra = "modsign"
 PREFERRED_PROVIDER_virtual/kernel:tegra ?= "linux-tegra"
 IMAGE_TEGRAFLASH_FS_TYPE:sota:tegra = "ota-ext4"
 INITRAMFS_FSTYPES:remove:tegra = "cpio.gz.cboot cpio.gz.cboot.bup-payload"
+IMAGE_TEGRAFLASH_INITRD_FLASHER:tegra = "${DEPLOY_DIR_IMAGE}/${TEGRAFLASH_INITRD_FLASH_IMAGE}-${MACHINE}.cpio.gz"
 INITRD_IMAGE:tegra = ""
 IMAGE_TEGRAFLASH_KERNEL:tegra = "${DEPLOY_DIR_IMAGE}/${KERNEL_IMAGETYPE}"
 IMAGE_FSTYPES:remove:tegra = "wic wic.gz wic.bmap"
```

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -618,6 +618,7 @@ OSTREE_BOOTLOADER:tegra ?= "syslinux"
 CORE_IMAGE_BASE_INSTALL:remove:tegra = "resize-helper"
 OSTREE_DEPLOY_DEVICETREE:tegra = "1"
 PREFERRED_PROVIDER_virtual/optee-os:tegra = "optee-os"
+#MACHINE_EXTRA_RDEPENDS:remove = "tegra-redundant-boot"
```
ricardosalveti, line 622: It is commented out here.

ricardosalveti: And it should probably be tegra specific.

ricardosalveti (approved):
LGTM

PR #1003, opened 2023-01-11
Backport main-next non-clangable

quaresmajose:

PR #1006, opened 2023-01-17
Fix known clang issues in LmP

MrCry0:
It supersedes PR  https://github.com/foundriesio/meta-lmp/pull/1003
- Add strip in clang excludes for Linux kernels
- Add a description and references to meta-clang
- Sort a list


angolini:
Do we know the root cause of those errors?

I don't see anything to comment on the commits, but I feel like a more verbose explanation might be helpful in the future.

angolini (approved):

ricardosalveti (approved):
LGTM

quaresmajose:
> Do we know the root cause of those errors?
> 
> I don't see anything to comment on the commits, but I feel like a more verbose explanation might be helpful in the future.

This is introduced somewhere in oe-core kirkstone but don't know yet where.

PR #1014, opened 2023-01-24
base: non-clangable: jailhouse: force gcc

ricardosalveti:
Jailhouse includes a kernel module, and the build fails if built with a different toolchain than used by the main kernel recipe.


ricardosalveti:
Will merge in order to fix the failures from https://github.com/foundriesio/lmp-manifest/pull/290.

PR #1015, opened 2023-01-24
Jailhouse: move config fragment to kernel features

quaresmajose:


quaresmajose:
Tested on am62xx-evm and imx8mp-lpddr4-evk.

MrCry0 (approved):
.

angolini:
I always get confused with the commit order presented here in github, so in case I'm wrong, please ignore my comment. But it looks to me that the first commit should be the last. First, you make the file available for every machine, and then you remove it from one specific machine.

angolini:
I wonder if it's not time to migrate the `cfg` file to the `lmp-kernel-cache` repository instead of keeping it as a file in the layer. Because this PR is making this accessible for all machines.

I know controlling kernel features from the distro POV always generates a lot of discussions, and I'm not trying to stress that. I'm good with whatever is proposed.

ricardosalveti:
Yeah, I would prefer this one to be available in lmp-kernel-cache, and included as we include modsign (I added this one as part of linux-lmp-ti-staging because the kernel is still 5.10 based, but we can also move it to kernel cache).

`meta-security/meta-integrity/recipes-kernel/linux/linux_ima.inc:KERNEL_FEATURES:append = " ${@bb.utils.contains('DISTRO_FEATURES', 'modsign', ' features/ima/modsign.scc', '', d)}"`

@quaresmajose would you mind creating a features file for jailhouse and add in lmp-kernel-cache for both 5.10 and 5.15 (and master)?

quaresmajose:
@ricardosalveti and @angolini I will move the fragment to lmp-kernel-cache and made the required adaption on this PR

MrCry0 (changes requested):
I think we don't really need to add and then revert commits :)

quaresmajose:
I am testing this PR on https://github.com/foundriesio/lmp-manifest/pull/291 and if everything looks good I cleanup this PR

ricardosalveti (approved):
Nice, thanks for cleaning this up.

PR #1020, opened 2023-01-26
AM6XXX: publish boot firmware files

ricardosalveti:


ricardosalveti:
Opening as a draft because this set is causing a huge amount of spdx errors (the usual removal before they are created). I believe this is because of the do_deploy dependencies at lmp-boot-firmware, which causes the dependency chain to have a different order, presenting the issue.

@quaresmajose another one for you to investigate as you are playing with a possible fix there.

quaresmajose:
This is another proof that the spdx issue is related with the do_deploy task, hope I can fix it shortly.

quaresmajose:
@ricardosalveti maybe this can now be rebased and merged.

quaresmajose (approved):
LGTM

MrCry0 (approved):
LGTM

PR #1027, opened 2023-02-01
Lmp staging: fix the compilation lock of the expensive recipes 

quaresmajose:


meta-lmp-base/classes/lmp-staging.bbclass
```diff
@@ -5,15 +5,15 @@
 #
 # Copyright 2022 (C) Foundries.IO LTD
```
MrCry0, line 6: We need to add 2023 here as well.

meta-lmp-base/classes/lmp-staging.bbclass
```diff
@@ -5,15 +5,15 @@
...
 
-INHERIT_KERNEL_MODSIGN = ""
+LMPSTAGING_INHERIT_KERNEL_MODSIGN = ""
```
MrCry0, line 8: Could you expand the commit message with an explanation of why we need these changes?

quaresmajose: The idea is to clarify that these variables are used in this classe. I will add it to the commit

meta-lmp-base/classes/lmp-staging.bbclass
quaresmajose: I think is better to remove the year instead of have a requirement to update it every year.

MrCry0 (approved):
LGTM

meta-lmp-base/classes/lmp-staging.bbclass
ricardosalveti: Copyright is related to the year range that is valid for it, so we can have it 2022-2023 atm.

ricardosalveti:
Small typo at the commit message from https://github.com/foundriesio/meta-lmp/pull/1027/commits/1befe767ef994f7dd360842603275ebc3eff0230

s/beackported/backported/g

meta-lmp-base/classes/lmp-staging.bbclass
quaresmajose: updated the year, in 2024 will remove this part.

quaresmajose:
> Small typo at the commit message from [1befe76](https://github.com/foundriesio/meta-lmp/commit/1befe767ef994f7dd360842603275ebc3eff0230)
> 
> s/beackported/backported/g

Also fixed

ricardosalveti (approved):
LGTM

PR #1031, opened 2023-02-02
k3: additional jailhouse fixes

ricardosalveti:
- Add missing memory node for k3
- Install inmate dtbs
- k3: build kernel Image along with fitImage


quaresmajose (approved):
LGTM

PR #1033, opened 2023-02-06
bsp: bootbin: versal: op-tee loaded by tf-a

ldts:
OP-TEE is not executed by the PLM but by TF-A instead.


ricardosalveti:
Loaded by PLM but executed by TF-A? Or is TF-A the one responsible for moving the memory around before jumping into it?

ldts:
The PLM copies it to the target location in DDR (https://github.com/ARM-software/arm-trusted-firmware/blob/master/plat/xilinx/versal/include/platform_def.h#L48). TF-A just jumps into it.

ldts:
image

ricardosalveti (approved):
LGTM

PR #1037, opened 2023-02-07
bsp: u-boot-ti-staging: am62xx-evm: use vfat for env

ricardosalveti:
Move the default environment from an emmc offset to the first vfat partition (sdcard) as a way to facilitate the environment management process on clean flash.


Tim-Anderson (approved):
LGTM

PR #1039, opened 2023-02-08
base: lmp: enable uninative

quaresmajose:
Attempts to isolate the build system from the host distribution’s C library in order to make re-use of native shared state artifacts across different host distributions practical.

https://docs.yoctoproject.org/ref-manual/classes.html?highlight=uninative#uninative-bbclass


MrCry0 (approved):
LGTM

PR #1041, opened 2023-02-09
Provision robustness

Tim-Anderson:
Changing from STM32CubeProgrammer_CLI version 2.11.0 to version 2.12.0 caused the provision.sh script to not work.

This PR makes it work correctly for these 2 versions of the tools and only those as well as making sure that any command that fails in running this script will not allow the script to continue but exit with an appropriate error.


ricardosalveti (approved):
LGTM, but needs approval from either @mike-scott or @igoropaniuk here.

igoropaniuk (approved):
LGTM

MrCry0 (approved):

PR #1054, opened 2023-02-28
bsp: initramfs-ostree-lmp-image: tegra: install tegra-firmware-xusb

ricardosalveti:
USB support requires firmware to be available in the initrd.

### DIFF
--- a/meta-lmp-bsp/recipes-core/images/initramfs-ostree-lmp-image.bbappend
+++ b/meta-lmp-bsp/recipes-core/images/initramfs-ostree-lmp-image.bbappend
@@ -1,0 +1,2 @@
+# USB support requires firmware to be available in the initrd
+PACKAGE_INSTALL:append:tegra = " tegra-firmware-xusb"


quaresmajose (approved):
LGTM, but we can also use the CONFIG_EXTRA_FIRMWARE for this one and made it available in all variant of lmp distros

ricardosalveti:
That ends up causing the kernel itself to be larger, and also not how it was done in the end in meta-tegra (even for the initrd flasher), but it could indeed be a more generic approach (maybe we can propose that to meta-tegra).

ricardosalveti:
Will merge to unblock this one.

quaresmajose:
> That ends up causing the kernel itself to be larger, and also not how it was done in the end in meta-tegra (even for the initrd flasher), but it could indeed be a more generic approach (maybe we can propose that to meta-tegra).

I can try to submit something.
Look like an improvement to make it more generic but in addition to be larger also introduces a dependency on kernel so to update the firmware we need to update the kernel.

ricardosalveti:
Indeed, and that might not necessarily be a good thing. Check with upstream, but low prio :-)

PR #1056, opened 2023-03-01
Add support for linux-tegra-rt

ricardosalveti:
Same as linux-tegra, but with the RT (available in the linux/rt-patches folder) applied, as suggested by meta-tegra.

### DIFF
--- a/meta-lmp-bsp/dynamic-layers/tegra/recipes-kernel/linux/linux-tegra-rt_5.10.bb
+++ b/meta-lmp-bsp/dynamic-layers/tegra/recipes-kernel/linux/linux-tegra-rt_5.10.bb
@@ -1,0 +1,6 @@
+require recipes-kernel/linux/linux-tegra_5.10.bb
+
+# Use BSP with rt-patches applied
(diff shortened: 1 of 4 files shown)


meta-lmp-bsp/dynamic-layers/tegra/recipes-kernel/linux/linux-tegra-rt_5.10.bb
```diff
@@ -0,0 +1,6 @@
+require recipes-kernel/linux/linux-tegra_5.10.bb
+
+# Use BSP with rt-patches applied
+SRC_REPO = "github.com/foundriesio/linux.git;protocol=https"
+SRCBRANCH = "oe4t-patches${LINUX_VERSION_EXTENSION}-rt"
```
quaresmajose, line 5: It is a good idea to have the variable `LINUX_VERSION_EXTENSION` here and thus follow the kernel upstream branch.

quaresmajose (approved):
LGTM

PR #1057, opened 2023-03-01
Aligning SDMA firmware version and load logic with NXP BSP

ricardosalveti:
Have SDMA provided by firmware-imx (latest revision) and add firmwared as it is required for early loading of the firmware when booting the system.


quaresmajose (approved):

meta-lmp-bsp/dynamic-layers/freescale-layer/recipes-bsp/firmware-imx/firmware-imx_8.%.bbappend
```diff
@@ -0,0 +1,4 @@
+do_install:append() {
```
angolini, line 1: if this is needed here, I wonder if this is needed upstream.
I'm up to reworking the firmware-imx_8 recipe mainline. I'm adding a note here, so that when I'm working on this, I remember where it's coming from

angolini (approved):
LGTM

PR #1060, opened 2023-03-06
Boot script: support flexible boot media

arnout:
Currently, the boot script and OSTree's uEnv.txt have hardcoded the paths to the root filesystem. Because of this, it is not possible to use the same image both for booting from eMMC and from SD card; instead, a separate machine must be created for each.

This PR adds flexibility to the boot script so the same script and rootfs can be used on different boot media. For this, it relies on `devnum` to be set dynamically, which is the case e.g. for the stm32mp15 boards.

This currently only works for eMMC and SD card, not for other boot media (USB flash, SATA, ...). To support those, the root device must be determined more flexibly than `/dev/mmcblkXXX`.

Only tested on stm32mp15-disco with lmp-base-console-image, and on our custom board with lmp-factory-image.


meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -642,9 +642,9 @@ PREFERRED_PROVIDER_virtual/optee-os:tegra = "optee-os"
 MACHINE_EXTRA_RDEPENDS:remove:tegra = "tegra-redundant-boot"
 MACHINE_FEATURES:append:tegra = " optee"
 ## jetson-agx-xavier-devkit (tegra194)
-OSTREE_KERNEL_ARGS:tegra194 ?= "\${cbootargs} root=/dev/mmcblk0p1 rootwait rootfstype=ext4 mminit_loglevel=4 console=tty0 console=ttyTCU0,115200 fbcon=map:0 sdhci_tegra.en_boot_part_access=1"
```
quaresmajose, line 645: tegra machines still needs the root kernel arg

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -642,9 +642,9 @@ PREFERRED_PROVIDER_virtual/optee-os:tegra = "optee-os"
...
+OSTREE_KERNEL_ARGS:tegra194 ?= "\${cbootargs} rootwait rootfstype=ext4 mminit_loglevel=4 console=tty0 console=ttyTCU0,115200 fbcon=map:0 sdhci_tegra.en_boot_part_access=1"
 ## jetson-agx-orin-devkit (tegra234)
-OSTREE_KERNEL_ARGS:tegra234 ?= "root=/dev/mmcblk0p1 rootwait rootfstype=ext4 mminit_loglevel=4 console=tty0 console=ttyTCU0,115200 fbcon=map:0"
```
quaresmajose, line 647: tegra machines still needs the root kernel arg

quaresmajose (commented):
Thanks for the contribution @arnout.

On tegra machine for bootload we use syslinux and not u-boot so the tegra changes should be dropped.

arnout:
> On tegra machine for bootload we use syslinux and not u-boot so the tegra changes should be dropped.

Both tegra changes reverted.

quaresmajose (approved):

MrCry0 (approved):
Nice.

igoropaniuk:
@arnout I think we can just use label here, which is set of by libostree in uEnv.txt, and just drop explicit setting of `root` in `bootargs` in `lmp-machine-custom.inc`:


```
$ cat /boot/loader/uEnv.txt
kernel_image=/boot/ostree/lmp-40240ba5992626f25b9e29cf617ba5bde2d5e0d738714bd57fecca3fa5d2c2c8/vmlinuz-5.15.64-lmp-standard
bootdir=/boot/ostree/lmp-40240ba5992626f25b9e29cf617ba5bde2d5e0d738714bd57fecca3fa5d2c2c8/
ramdisk_image=/boot/ostree/lmp-40240ba5992626f25b9e29cf617ba5bde2d5e0d738714bd57fecca3fa5d2c2c8/initramfs-5.15.64-lmp-standard.img
bootargs=console=ttySTM0,115200 root=LABEL=otaroot rootfstype=ext4 ostree=/ostree/boot.1/lmp/40240ba5992626f25b9e29cf617ba5bde2d5e0d738714bd57fecca3fa5d2c2c8/0
```

arnout:
> @arnout I think we can just use label here

Our use case was (among others) to be able to boot from SD card while something was already flashed to eMMC. If you use LABEL=, it's going to mount the *first* filesystem with that label - so most likely the one on eMMC (regardless of what you booted from). Then the initramfs will not find the ostree hash (because it's actually on SD card) and boot fails. (Note: this is just an assumption, I haven't tested, I could be wrong).

igoropaniuk:
@arnout for this particular case (when both eMMC and SD are used) this definetely make sense, thanks for the explanation!

LGTM

igoropaniuk (approved):

ricardosalveti (approved):
LGTM

PR #1061, opened 2023-03-06
bsp: bootimg-sota-efi: allow dynamic bootable partition

ricardosalveti:
Remove the hardcoded bootable flag on the first partition and instead search for active partitions when defining the bootable one.

This allows the ESP partition to be created as second or third partition (for cases where the boot firmware partition needs to be defined first).


MrCry0 (approved):
LGTM

quaresmajose (approved):

angolini (approved):

PR #1062, opened 2023-03-07
Fix u-boot option names

MrCry0:


quaresmajose (approved):

ricardosalveti:
Looks like the last commit was added by accident?

MrCry0:
> Looks like the last commit was added by accident?

That commit reduces the size of a u-boot container significantly so that we can add and test some other features in `lmp-base` distro.

ricardosalveti:
> That commit reduces the size of a u-boot container significantly so that we can add and test some other features in `lmp-base` distro.

Sure, it is a good patch, just out of the context from this PR.

Mind having a new pr just for that one?

PR #1065, opened 2023-03-12
bsp: optee: extend pkcs11 ta heap size

igoropaniuk:
Extend default PKCS11 TA heap size to 64Kb, as on some platforms the PKCS11 TA runs out of memory when generating the RSA keys [1] Set 256Kb PKCS11 TA heap size when SE05x support is enabled. Before it was set only for imx8mm-lpddr4-evk MACHINE.

[1] https://github.com/OP-TEE/optee_os/issues/5022


MrCry0 (approved):

quaresmajose (approved):

meta-lmp-bsp/recipes-security/optee/optee-os-fio-bsp.inc
```diff
@@ -91,11 +91,14 @@ EXTRA_OEMAKE:append:stm32mp15-eval = " \
 "
 
 # Additional Settings for SE05X
+EXTRA_OEMAKE:append = " \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'se05x', 'CFG_PKCS11_TA_HEAP_SIZE=262144', 'CFG_PKCS11_TA_HEAP_SIZE=65536', d)} \
```
ricardosalveti, line 95: I think it is probably a good idea to set with and without se050, as then we can have the same value set all over.

ricardosalveti: As we discussed, better indeed to let it only when se050 is used, and for that you can move this config to 'meta-lmp-base/recipes-security/optee/optee-os-fio-se05x.inc', which is only included when se050 is enabled.

igoropaniuk: The problem is that default value for now is 32KB. This is why I added it exactly here:
1. Increase to 64KB if se05x is not used
2. Set to 256KB is se05x is used


If I move this config meta-lmp-base/recipes-security/optee/optee-os-fio-se05x.inc, I'll still need this condition here, which will check for se05x

MrCry0: @igoropaniuk will this work?
In `optee-os-fio-bsp.inc`:
```
PKCS11_TA_HEAP_SIZE ??= 65536
EXTRA_OEMAKE:append = " CFG_PKCS11_TA_HEAP_SIZE=${PKCS11_TA_HEAP_SIZE}"
```
In `optee-os-fio-se05x.inc`:
```
PKCS11_TA_HEAP_SIZE = 262144
```

igoropaniuk: @MrCry0 thanks for suggestions, redone!

MrCry0 (approved):
LGTM

meta-lmp-bsp/recipes-security/optee/optee-os-fio-bsp.inc
igoropaniuk: @ricardosalveti comment addressed

ricardosalveti (approved):
LGTM

PR #1068, opened 2023-03-15
bsp: optee-os-fio: enable CAAM for imx7ulpea-com

MrCry0:
Now the


quaresmajose (approved):
LGTM, tested locally and fixes the link error

MrCry0:
> LGTM, tested locally and fixes the link error

Thanks! We may wait for @vanmaegima 's approvement, Vanessa knows more about 7ulp ;-)

ricardosalveti:
Yeah, thanks for looking at this one, we should be good as we only want to make sure we're able to build correctly for it at this moment, as we will later move to the optee imx bsp version instead (only for ulp).

ricardosalveti (approved):
LGTM

PR #1073, opened 2023-03-16
Updates for TI AM62xx jailhouse support

ricardosalveti:


quaresmajose (approved):

angolini (approved):

meta-lmp-bsp/recipes-bsp/u-boot/u-boot-ostree-scr-fit/am62xx-evm/boot.cmd
```diff
@@ -3,12 +3,15 @@ setenv bootlimit 3
 setenv devtype mmc
 setenv devnum ${mmcdev}
 
+# Use "#conf-overlayname.dtbo"
```
MrCry0, line 6: Is it correct? `"#conf-overlayname.dtbo"`, not `"#conf-ti_overlayname.dtbo"`

ricardosalveti: > I think the correct template is `"#conf-ti_overlayname.dtbo"`, isn't it?

Sure, updated to follow that suggestion.

MrCry0: > The problem there is that it will fail when empty, and the reason why the variable has #conf in it.


Ah, the `dtoverlay` is optional. Sure, you're right. Thanks for the explanation!

PR #1074, opened 2023-03-16
Ti BSP update to 08.06.00.007

ricardosalveti:
Tested on am62/64.


MrCry0 (approved):
We may need to update kmeta on the top of https://github.com/foundriesio/lmp-kernel-cache/pull/101

LGTM

PR #1075, opened 2023-03-17
base: fioconfig: Bump version

doanac:
Includes changes for:

 * Improved retry handling when cert rotation fails
 * Move from umbracle/ecies to foundries/ecies fork

### DIFF
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
@@ -11,7 +11,7 @@ SRC_URI = "git://${GO_IMPORT};protocol=${GO_IMPORT_PROTO};branch=main \
 	file://fioconfig.path \
 	file://fioconfig-extract.service \
 "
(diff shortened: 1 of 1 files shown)


doanac:
@rsalveti - I *think* the systemd service definition is okay. However, its missing a `StartLimitInterval=0`. But since the restart interval is 10s, I don't think fioconfig can get rate-limited and disabled, right?

@mwasilew - this will need some testing.

ricardosalveti:
` DefaultStartLimitIntervalSec= defaults to 10s. DefaultStartLimitBurst= defaults to 5.`

In the end I think it will take more than 10s, so won't necessarily get in a failed state here.

The question that follows is how do we want fioconfig service logic to be managed in case of failures? I guess we can always restart if failed in a range, which would cover the expected failure use case, and fail hard when in a crash loop (causing the rest of the system to reboot?).

ricardosalveti (approved):
LGTM, will merge so we can do a bit more testing here.

PR #1077, opened 2023-03-21
Fix for alsa with ostree env

MrCry0:


meta-lmp-base/recipes-multimedia/alsa/alsa-utils_%.bbappend
```diff
@@ -0,0 +1,11 @@
...
+# file which is required for some sound systems (i.e. wm8960 on imx8mp-evk)
+# to work properly.
+
+python __anonymous() {
+    if bb.utils.contains('IMAGE_CLASSES', 'image_types_ostree', True, False, d):
+        d.setVar("localstatedir", d.getVar("prefix"))
+}
+
+EXTRA_OECONF:append = " --with-asound-state-dir=${localstatedir}/lib/alsa"
+EXTRA_OEMAKE:append = " SYS_ASOUND_DIR=${localstatedir}/lib/alsa"
```
ricardosalveti, line 11: One problem I see is that this is also used by alsactl when saving custom mixer data, and that will fail when stored in /usr/lib.

I think it is ok to have it using /var/lib/alsa, as long the directory is created at runtime via tmpfiles.d and that the default asound.state file can be consumed from another location (or if that file gets copied over together with the process that is responsible for creating the folder at runtime).

Then we can have the same solution for both lmp-base and lmp.

MrCry0: I need to test this variant. The question is where alsactl is called first, from udev rules without path to the asound.state or from post-install package script, where we alter the path to the config file during image building.

Tim-Anderson: so on all devices that have sound (in this case imx boards) we should enable alsa by default or make it a MACHINE_FEATURE

MrCry0: @ricardosalveti I changed the solution preserving default behavior.

meta-lmp-base/recipes-samples/images/lmp-feature-alsa.inc
```diff
@@ -0,0 +1,5 @@
+# Basic ALSA packages
+CORE_IMAGE_BASE_INSTALL += " \
+    alsa-utils \
+    alsa-state \
+"
```
ricardosalveti, line 5: I think these are already installed by default once alsa is available in DISTRO_FEATURES:

```
oe-core/meta/recipes-core/packagegroups/packagegroup-base.bb:

SUMMARY:packagegroup-base-alsa = "ALSA sound support"
RDEPENDS:packagegroup-base-alsa = "\
    alsa-utils-alsactl \
    alsa-utils-amixer \
    ${VIRTUAL-RUNTIME_alsa-state}"
...
RDEPENDS:packagegroup-base = "\
    packagegroup-distro-base \
    packagegroup-machine-base \
    \
    module-init-tools \
    ${@bb.utils.contains('MACHINE_FEATURES', 'apm', 'packagegroup-base-apm', '',d)} \
    ${@bb.utils.contains('MACHINE_FEATURES', 'acpi', 'packagegroup-base-acpi', '',d)} \
    ${@bb.utils.contains('MACHINE_FEATURES', 'keyboard', 'packagegroup-base-keyboard', '',d)} \
    ${@bb.utils.contains('MACHINE_FEATURES', 'phone', 'packagegroup-base-phone', '',d)} \
    \
    ${@bb.utils.contains('COMBINED_FEATURES', 'alsa', 'packagegroup-base-alsa', '',d)} \

```

meta-lmp-base/recipes-multimedia/alsa/alsa-utils/0001-alsactl-add-fallback-for-restoring-from-asound.state.patch
```diff
@@ -0,0 +1,66 @@
...
+From: Oleksandr Suvorov <oleksandr.suvorov@foundries.io>
+Date: Thu, 23 Mar 2023 22:03:14 +0200
+Subject: [PATCH] alsactl: add fallback for restoring from asound.state
+
+On ostree-based disros asound.state cannot be installed into /var
+directory. Instead, this config installs into /usr/lib/alsa.
+
+Support restoring vendor-specific sound configuration from
+/usr/lib/alsa/asound.state if /var/lib/alsa/asound.state is
+unavailable.
```
ricardosalveti, line 11: I would probably make the commit message a bit different and not necessarily as ostree specific, since the problem we have is that alsa-utils is assuming that a file which is system specific to be available in a rw folder that can be removed by the user at any point in time, causing a different system behavior (from initial flash).

What you added is support to restore the state from a read-only location when the var content is not available, which I think it is valid and we can even try to upstream.

meta-lmp-base/recipes-multimedia/alsa/alsa-utils_%.bbappend
ricardosalveti: Do we still need "--with-asound-state-dir=${localstatedir}/lib/alsa" and "SYS_ASOUND_DIR=${localstatedir}/lib/alsa"?

meta-lmp-base/recipes-multimedia/alsa/alsa-utils_%.bbappend
```diff
@@ -0,0 +1,10 @@
+# Try to restore /usr/lib/alsa/asound.state if /var/lib/alsa/asound.state
+# is unavailable (it happens on ostree-based distro).
+
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+EXTRA_OECONF:append = " --with-init-asound-state-dir=${prefix}/lib/alsa \
+                        --with-asound-state-dir=${localstatedir}/lib/alsa "
+EXTRA_OEMAKE:append = " SYS_ASOUND_DIR=${localstatedir}/lib/alsa"
+
+SRC_URI += " file://0001-alsactl-add-fallback-for-restoring-from-asound.state.patch "
```
ricardosalveti, line 10: With += you don't need empty space before/after the line.

meta-lmp-base/recipes-bsp/alsa-state/alsa-state/alsa-state-init
```diff
@@ -0,0 +1,43 @@
+#! /bin/sh
+#
```
ricardosalveti, line 2: Do we use this script at all? Since we're not using sysvinit. I know it could be called by systemd, but curious if that is the current use case.

meta-lmp-base/recipes-multimedia/alsa/alsa-utils_%.bbappend
MrCry0: > Do we still need "--with-asound-state-dir=${localstatedir}/lib/alsa" and "SYS_ASOUND_DIR=${localstatedir}/lib/alsa"?

Not mandatory, but what if we build a distro with a variable directory other than /var? :)

meta-lmp-base/recipes-bsp/alsa-state/alsa-state.bbappend
```diff
@@ -0,0 +1,25 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+# ostree-based distros should not commit any files into /var
+# Thus, install a vendor-specific asound.state file into
+# /usr/lib/alsa/ for ostree images.
+VAR_STATEDIR  = "${localstatedir}/lib/alsa"
+INIT_STATEDIR = "${@bb.utils.contains('IMAGE_CLASSES', 'image_types_ostree', "${prefix}", "${localstatedir}", d)}/lib/alsa"
```
ricardosalveti, line 7: I would just make this the default behavior in lmp, with and without ostree.

It is a good change that is not really depending on ostree, but it was caused by the expectations from it.

meta-lmp-base/recipes-multimedia/alsa/alsa-utils_%.bbappend
ricardosalveti: Then we will have way bigger problems :D

meta-lmp-base/recipes-bsp/alsa-state/alsa-state/alsa-state-init
MrCry0: It runs only in sysvinit systems. Well, if we want to try to upstream such solution, it should generic, not only for systemd systems.

meta-lmp-base/recipes-bsp/alsa-state/alsa-state.bbappend
MrCry0: Ok, fixing.

Tim-Anderson (approved):
one thing that is missing is adding fio to the audio group

Otherwise LGTM

meta-lmp-base/recipes-samples/images/lmp-feature-alsa.inc
MrCry0: @ricardosalveti Not really. Only `alsa-state` and 2 utils `alsactl` and `amixer` are installed with the MACHINE_ or COMBINED_FEATURES alsa. Neither aplay/arecord nor test sounds, sample configs etc are installed with that feature got set. So I'm going to use COMBINED_FEATURES instead of DISTRO_FEATURES but we still need to include the full `alsa-utils` package, not just some subpackages.

MrCry0:
> one thing that is missing is adding fio to the audio group

@Tim-Anderson Thanks for pointing that out! Fixed.

meta-lmp-base/recipes-multimedia/alsa/alsa-utils_%.bbappend
MrCry0: Fxed

meta-lmp-base/recipes-multimedia/alsa/alsa-utils_%.bbappend
MrCry0: Ok, got rid that stuff :)

meta-lmp-base/recipes-samples/images/lmp-feature-alsa.inc
ricardosalveti: Sure, but in this case we can safely drop at least alsa-state, right? Since packagegroup-base-alsa will be installed anyway.

meta-lmp-base/recipes-samples/images/lmp-image-common.inc
```diff
@@ -43,6 +43,9 @@ CORE_IMAGE_BASE_INSTALL:append:sota = " \
     ostree-recovery-initramfs \
 "
 
+# Common features
+require ${@bb.utils.contains('COMBINED_FEATURES', 'alsa', 'lmp-feature-alsa.inc', '', d)}
```
ricardosalveti, line 47: I think the correct would be distro features instead, combined is a way for yocto to match what is available in the hardware (matching distro with machine features).

meta-lmp-base/recipes-samples/images/lmp-feature-alsa.inc
```diff
@@ -0,0 +1,11 @@
+# LmP specific settings for ALSA
+
+# Add user fio to audio group to control audio devices
+inherit extrausers
+EXTRA_USERS_PARAMS = "usermod -a -G audio fio"
```
ricardosalveti, line 5: This was already done as part of https://github.com/foundriesio/meta-lmp/commit/9742ad8cae71fa87d00f4078c42ad8d932328e82

meta-lmp-base/recipes-samples/images/lmp-feature-alsa.inc
MrCry0: Done.

meta-lmp-base/recipes-samples/images/lmp-image-common.inc
MrCry0: I didn't want to alter `packagegroup-base.bb`. But yeah, distro is the better place, as it was from the beginning. Ok, fixing.

MrCry0: Also, I re-thought using the full alsa-utls package, it may be needed for debug build only, so I moved alsa-utils to the feature-debug, adding only `alsa-utils-aplay` (which includes `arecord` as well) to make alsa subsystem usable in LmP images.

meta-lmp-base/recipes-samples/images/lmp-feature-alsa.inc
MrCry0: Fixed.

meta-lmp-base/recipes-core/packagegroups/packagegroup-base.bbappend
```diff
@@ -0,0 +1,9 @@
+# Enable using alsa as a DISTRO feature.
+
+PACKAGES += " \
+    ${@bb.utils.contains("DISTRO_FEATURES", "alsa", "packagegroup-base-alsa", "", d)} \
```
ricardosalveti, line 4: I think this can probably be dropped as packagegroup-base-alsa is already installed when alsa is available distro features.

We depend on packagegroup-base-extended, which depends on packagegroup-base, which depends on packagegroup-base-alse when it is also available in machine features (when really needed).

meta-lmp-base/recipes-core/packagegroups/packagegroup-base.bbappend
```diff
@@ -0,0 +1,9 @@
...
+"
+
+RDEPENDS:packagegroup-base-alsa += "\
+    alsa-utils-aplay \
```
ricardosalveti, line 8: This is fine to keep.

meta-lmp-base/recipes-core/packagegroups/packagegroup-base.bbappend
MrCry0: I still can't find that part in the OE code which "converts" DISTRO features into groups dependencies, but yeah, it works, so I dropped it. Done.

meta-lmp-base/recipes-multimedia/alsa/alsa-utils/0001-alsactl-add-fallback-for-restoring-from-asound.state.patch
```diff
@@ -0,0 +1,66 @@
...
++		  -e 's,@initasoundrcfile\@,$(INIT_ASOUND_STATE_DIR)/asound.state,g' \
+ 							< $< > $@ || rm $@
+ 
+ alsa-state.service: alsa-state.service.in
+diff --git a/configure.ac b/configure.ac
+index 92aec3a..8a14f58 100644
+--- a/configure.ac
++++ b/configure.ac
+@@ -419,6 +419,12 @@ AC_ARG_WITH([asound-state-dir],
+         [ASOUND_STATE_DIR="/var/lib/alsa"])
```
quaresmajose, line 52: @MrCry0 why not change `ASOUND_STATE_DIR` to use `/usr` instead of add a new `INIT_ASOUND_STATE_DIR`?

MrCry0: We ought to preserve the original behavior. Moreover, usually `alsactl` can't write into `/usr/lib/alsa/`.
So that the tool should store the configuration into `/var/lib/alsa/` and should try to restore the config from `/var/lib/alsa`. And only if there is no default config file, try to load an initial one from `/usr/lib/alsa`.

quaresmajose: Thanks for explaining and also I forget that `/usr` is read-only!

quaresmajose (approved):

ricardosalveti (approved):
LGTM, thanks for working on this!

PR #1083, opened 2023-03-27
tegra: support creating wic images along with tegraflash

ricardosalveti:


MrCry0 (approved):

quaresmajose (approved):
LGTM, maybe one day we could switch from the `resize-helper` script to [systemd-repart](https://www.freedesktop.org/software/systemd/man/systemd-repart.html) that can  automatically grow and add partitions

ricardosalveti:
> LGTM, maybe one day we could switch from the `resize-helper` script to [systemd-repart](https://www.freedesktop.org/software/systemd/man/systemd-repart.html) that can automatically grow and add partitions

Only when it supports MBR and after we support systemd in the initrd.

ricardosalveti:
Tested by adding the following to conf/local.conf (or conf/machine/include/lmp-factory-custom.inc on a factory):

```
TEGRAFLASH_ROOTFS_EXTERNAL:jetson-agx-xavier-devkit = "1"
TEGRA_EXTERNAL_ROOTFS_WIC:jetson-agx-xavier-devkit = "1"
ROOTFS_DEVICE_FOR_INITRD_FLASH:jetson-agx-xavier-devkit = "nvme0n1"
```

ricardosalveti:
@MrCry0 @quaresmajose mind having another look? added the initrd related patches, thanks!

quaresmajose (approved):
LGTM. The `tegra-flash-init` and the `tegra-helper-scripts` look appropriate to submit to meta-tegra, should I do that?

MrCry0 (approved):
Still LGTM

ricardosalveti:
@quaresmajose sure, it would be good to check what is nvidia bsp / jetpack and what is meta-tegra specific there, and propose the changes, as it should indeed be generic. Please have a look and let me know if you need a hand in there.

quaresmajose:
> @quaresmajose sure, it would be good to check what is nvidia bsp / jetpack and what is meta-tegra specific there, and propose the changes, as it should indeed be generic. Please have a look and let me know if you need a hand in there.

Ok, I will do that.

PR #1088, opened 2023-03-29
bsp: edk2-firmware-tegra: fix line endings

ricardosalveti:
Git converted the line endings when moving the patch around (git format-patch), which causes a do_patch failure.

Fix by using the original patches I created locally, which have the right line endings.


ricardosalveti:
Will merge to unblock latest lmp-manifest update.

PR #1089, opened 2023-03-30
Add back support for grub-based installers and images (for arm64)

ricardosalveti:
Needed on arm64 as the current sd-boot version does not support loading initrds via LINUX_INITRD_MEDIA_GUID (only via sd-stub, which requires UKI).

### DIFF
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
@@ -264,9 +264,9 @@ mkdir /src_root
 mkdir -p /boot
 
 # Path /rnu/media/<label>-sdX (check wks used)
-install_mount="install-${live_dev_name}1"
-otaboot_mount="otaboot-${live_dev_name}2"
-rootfs_mount="image-${live_dev_name}3"
+install_mount=`ls /run/media | grep install-`
+otaboot_mount=`ls /run/media | grep otaboot- || true`
+rootfs_mount=`ls /run/media | grep image-`
 
 # Handling of the target root partition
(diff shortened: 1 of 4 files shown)


meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
```diff
@@ -214,7 +214,19 @@ while true; do
...
     elif [ "$answer" = "n" ]; then
-        echo "Not erasing current partition table for device ${device}, assuming ${bootfs} as boot/ESP and ${rootfs} as rootfs."
+        echo "Not erasing current partition table for device ${device}. Use ${bootfs} as boot/ESP and ${rootfs} as rootfs? (y - default / n)"
+        echo
+        read answer
+        if [ "$answer" = "n" ]; then
+            echo "Please define the partition to be used as boot/ESP (e.g. ${bootfs}): "
+            echo
+            read answer
+            bootfs=${answer}
```
quaresmajose, line 224: given that the user can change the default value maybe is better to validate answer for `bootfs` and `rootfs`
```
while true; do
  echo "Please define the partition to be used as boot/ESP (e.g. ${bootfs}): "
  echo
  read answer
    if [ -b ${answer} ]; then
      break
    else
      echo "Invalid block device !"
    fi
done
```

ricardosalveti: Had that initially but it kind of looks ugly to have a set of nested loops just to confirm when this option is already for advanced users.

But I agree that it is better to check, I'm sure people will make mistakes here.

quaresmajose (approved):
LGTM

PR #1103, opened 2023-04-17
Revert "base: lmp: add make-mod-scripts to RM_WORK_EXCLUDE"

quaresmajose:
This reverts commit 336254ea5acc67dda182c407789f5414dcfe1fec.

Fixed upstream [1] with RM_WORK_EXCLUDE_ITEMS.
[1] https://lists.openembedded.org/g/openembedded-core/message/180113


quaresmajose:
Patch not accepted and the fix for this will be link the tools statically that it is unlikely to be backported so we need this on kirkstone

PR #1115, opened 2023-04-20
[WIP] base: recipes-core: initrdscripts: clear RPMB if aktualizr is configured to use PKCS11

bravl:
If aktualizr-lite was configured to used RPMB in order to store it certifactes, the slot used by aktualizr-lite should be cleared in order to prevent keys and certificates for being exposed after a factory reset.

### DIFF
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/ostree_factory_reset
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/ostree_factory_reset
@@ -15,6 +15,8 @@ ostree_factory_reset_enabled() {
 factory_reset() {
 	keep_sota=$1
 	keep_docker=$2
+	keep_rpmb="false"
+	aktualizr_uses_rpmb="false"
 
 	msg "Performing factory reset..."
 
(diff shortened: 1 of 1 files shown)


bravl:
Still WIP, need to further test but wanted to validate if this would be an acceptable approach.

meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/ostree_factory_reset
```diff
@@ -34,15 +36,34 @@ factory_reset() {
...
+	fi
+
+	# Cleanup RPMB before /var gets removed since we need some
+	# parameters which are stored in /var/sota/sota.toml
+	if [ "${keep_rpmb}" = "false" ] && [ "${aktualizr_uses_rpmb}" = "true" ]; then
+		ptool_module=$(grep module sota/sota.toml | cut -d'"' -f 2)
+
+		# Reinitializing a PKCS11 token destroys all objects which
+		# were associated with that slot.
+		pkcs11-tool --module ${ptool_module} --init-token --label "unused" --slot ${AKTUALIZR_SLOT}
```
ricardosalveti, line 62: Here I assume you are using the pkcs11 OP-TEE implementation (due RPMB access), so this call will only really work in a generic way with the right module, as we could have other pkcs11 providers that are not necessarily provided by OP-TEE (e.g. softhsm2, pkcs11-tpm2, etc).

Since we can assume we're using pkcs11 OP-TEE we can search if the module is libckteec, and skip otherwise.

You might also need the sopin to initialize, but needs testing (it is also available in sota.toml if needed).

One other thing, since the pkcs11 OP-TEE implementation requires that the pkcs11 trusted application to be loaded and operational in optee, you will need to make sure tee-supplicant is running before executing this command (libckteec will request the TA to be loaded via tee-supplicant on first access). You can use https://github.com/foundriesio/meta-lmp/blob/main/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/cryptfs_pkcs11 as a guide, since it is also executed as part of the initrd.

meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/ostree_factory_reset
```diff
@@ -34,15 +36,34 @@ factory_reset() {
...
 		msg "Keeping current ${OSTREE_DISTRO} SOTA content"
 		FIND_EXCLUDE="${FIND_EXCLUDE} ! -name 'sota'"
+		keep_rpmb="true"
 	fi
 	if [ "${keep_docker}" = "true" ]; then
 		msg "Keeping current docker content"
 		FIND_EXCLUDE="${FIND_EXCLUDE} ! -name 'lib'"
 	fi
 
+	cd ${OSTREE_VAR}
```
ricardosalveti, line 46: Not related to this PR, but to help testing this approach you can just call `sh` and iterate over the commands manually.

ricardosalveti:
> Still WIP, need to further test but wanted to validate if this would be an acceptable approach.

Yup, thanks for starting this one, this is the approach I would take as well.

PR #1117, opened 2023-04-21
docker: fix dockerd failed start conditions

mwasilew:
When dockerd fails to start after OTA, systemd will reboot the system to force rollback. This patch allows dockerd 3 attempts before rebooting.


quaresmajose (approved):

MrCry0 (approved):
LGTM

PR #1120, opened 2023-04-24
base: u-boot-ostree-scr-fit: update fiovb.bootfirmware_version

igoropaniuk:
Update a value of fiovb.bootfirmware_version during boot firmware upgrade procedure alongside with bootfirmware_version.


ricardosalveti (approved):
LGTM

PR #1122, opened 2023-04-24
base: packagegroup-security-tpm2: remove trousers

ricardosalveti:
Remove trousers from tpm2 related packagegroup as it is relevant with tpm 1.2 only, which we don't officially support in LmP.

### DIFF
--- a/meta-lmp-base/recipes-core/packagegroups/packagegroup-security-tpm2.bbappend
+++ b/meta-lmp-base/recipes-core/packagegroups/packagegroup-security-tpm2.bbappend
@@ -1,0 +1,2 @@
+# Trousers is relevant with TPM 1.2 (not supported by LmP)
+RDEPENDS:packagegroup-security-tpm2:remove = "trousers"


MrCry0 (approved):

meta-lmp-base/recipes-core/packagegroups/packagegroup-security-tpm2.bbappend
```diff
@@ -0,0 +1,2 @@
+# Trousers is relevant with TPM 1.2
```
angolini, line 1: What I understand is that Trousers is only relevant for TPM 1.2, and because LMP does not use 1.2, it can be removed.

however, the comment is ambiguous why are we removing something relevant? Maybe use "Trousers is relevant only for TMP1.2 (not supported by LmP)" instead

angolini (approved):
only minor suggestion

meta-lmp-base/recipes-core/packagegroups/packagegroup-security-tpm2.bbappend
ricardosalveti: The issue is that this packagegroup is tpm2 specific, so it doesn't really make much sense to have trousers also added as part of this packagegroup (there is one just for tpm, which implies tpm 1.2).

ricardosalveti: Extended the note.

PR #1125, opened 2023-04-26
base: move var clean up from base-files to lmp.bbclass

ricardosalveti:
Var directories, created by base-files, are also used by the rootfs task while performing the image bootstrap process (via packages), so they can't be removed at the base-files package level.

Move the removal logic to lmp.bbclass via ROOTFS_POSTPROCESS_COMMAND task, which is executed only when ota-ext4 is set in IMAGE_FSTYPES (only way to know when an ostree image is to be created), after all required rootfs packages are installed. Only dirs that can't be removed at this stage are /var/lib and /var/cache as they are used to store packaging related metadata.


MrCry0 (approved):

quaresmajose (approved):
LGTM

PR #1130, opened 2023-05-03
CAAM changes and fixes required for imx8mn

ricardosalveti:
* Enable CAAM on imx8mn fixing boot issues
* Force CAAM as disabled on other targets, until properly validated with the rest of the stack (as it depends on the hw support, device-tree, u-boot and so on)


ricardosalveti:
@MrCry0 once we bump to the 6.1 tag we can start enabling caam on every target, and also validate sha hw support, besides RSA (which requires SPL support).

All of that comparing and checking if there is no major size change due size restrictions we have on some platforms.

quaresmajose (approved):

PR #1131, opened 2023-05-03
base: linux-lmp-rt: switch to 6.1 as default rt

ricardosalveti:
Follow linux-lmp and switch to the 6.1 series by default.


quaresmajose:
With this we can also drop the 5.15?

ricardosalveti:
> With this we can also drop the 5.15?

Yup, planning on removing the 5.15 based recipes after the release.

MrCry0 (approved):
LGTM

PR #1132, opened 2023-05-03
Update TI bsp

quaresmajose:


ricardosalveti (approved):
Just missing one change in order to boot correctly but I will follow that up in another PR.

One other major difference here is that the default build for AM62XX is now HS-FS, which won't work right on GP (most AM62XX variants we have).

Had to force another tiboot3.bin for my GP device to continue to boot:
```
# TI
IMAGE_BOOT_FILES:append:am62xx-evm = " tiboot3-am62x-gp-evm.bin;tiboot3.bin"
IMAGE_BOOT_FILES:remove:am62xx-evm = "tiboot3.bin"
```

This will have to be documented properly.

quaresmajose:
> Just missing one change in order to boot correctly but I will follow that up in another PR.
> 
> One other major difference here is that the default build for AM62XX is now HS-FS, which won't work right on GP (most AM62XX variants we have).
> 
> Had to force another tiboot3.bin for my GP device to continue to boot:
> 
> ```
> # TI
> IMAGE_BOOT_FILES:append:am62xx-evm = " tiboot3-am62x-gp-evm.bin;tiboot3.bin"
> IMAGE_BOOT_FILES:remove:am62xx-evm = "tiboot3.bin"
> ```
> 
> This will have to be documented properly.

I have to do some tests on my board

PR #1136, opened 2023-05-04
base: u-boot/fip: fix bootfirmware version without firmware node

ricardosalveti:
Create the /firmware/bootloader node using auto-path, in order to automatically create all nodes required for the node path (useful for cases where /firmware is not defined by the board).


ricardosalveti:
Force merging for the release, tested locally.

PR #1141, opened 2023-05-08
[backporting] Update RPI kernel linux-lmp-rpi to 5.15

quaresmajose:


ricardosalveti:
Usually there is a need to match the rpi firmware with the kernel version used as well, so something we might want to review as well.

quaresmajose:
> Usually there is a need to match the rpi firmware with the kernel version used as well, so something we might want to review as well.

I will take a look the other pieces, Thanks for pointing that.

MrCry0 (approved):
LGTM

ricardosalveti (approved):
LGTM

PR #1143, opened 2023-05-08
base: lmp-device-register: bump to 263d11e

ricardosalveti:
Relevant changes:
- 263d11e main: refactor lmp-device-register

This version does not depend on pkcs11-tool for pkcs11 initialization and key creation, using libp11 directly instead.


quaresmajose (approved):

quaresmajose:
> Relevant changes:
> 
> * 263d11e main: refactor lmp-device-register
> 
> This version does not depend on pkcs11-tool for pkcs11 initialization and key creation, using libp11 directly instead.

if not specified it will build with pkcs11-too so we should need to use:
```
EXTRA_OECMAKE += "-DDISABLE_PKCS11=1"
```

https://github.com/foundriesio/lmp-device-register/pull/41/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR60

doanac (commented):
Can we drop the openssl-bin RDEPENDS?

ricardosalveti:
> if not specified it will build with pkcs11-too so we should need to use:

It doesn't depend on pkcs11-tool (runtime dependency, which was already missing before), but it still depends on libp11 (that was the main work done by Jorge).

angolini (approved):

MrCry0 (approved):

PR #1148, opened 2023-05-17
base: lmp: sota_var_cleanup: use the ostree_rmdir_helper

quaresmajose:
It will help removing directories with content inside withou failing.


meta-lmp-base/classes/lmp.bbclass
```diff
@@ -19,8 +19,13 @@ sota_var_cleanup() {
 	if ${@bb.utils.contains('IMAGE_FSTYPES', 'ota-ext4', 'true', 'false', d)}; then
 		# Remove /var stuff (ignored by ostree)
 		cd ${IMAGE_ROOTFS}/var
-		rmdir -v backups spool local lib/misc
-		rmdir -v volatile ${@'' if oe.types.boolean('${VOLATILE_LOG_DIR}') else 'log'}
+		ostree_rmdir_helper backups
+		ostree_rmdir_helper spool
+		ostree_rmdir_helper local
+		ostree_rmdir_helper lib/misc
+		ostree_rmdir_helper volatile
```
MrCry0, line 26: the original logic removes either volatile or log folder. Is it ok to remove volatile unconditionally?

MrCry0 (approved):

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -19,8 +19,13 @@ sota_var_cleanup() {
...
+		if not oe.types.boolean('${VOLATILE_LOG_DIR}'):
```
ricardosalveti, line 27: This won't parse correctly as it is shell and not a python code block.

quaresmajose: right, it fails to parse. fixed and tested locally and `ostree_rmdir_helper` works with arguments

MrCry0 (approved):

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -19,8 +19,12 @@ sota_var_cleanup() {
...
 		# Remove /var stuff (ignored by ostree)
 		cd ${IMAGE_ROOTFS}/var
-		rmdir -v backups spool local lib/misc
-		rmdir -v volatile ${@'' if oe.types.boolean('${VOLATILE_LOG_DIR}') else 'log'}
+		ostree_rmdir_helper backups
+		ostree_rmdir_helper spool
+		ostree_rmdir_helper local
+		ostree_rmdir_helper lib/misc
+		ostree_rmdir_helper volatile
+		ostree_rmdir_helper ${@'' if oe.types.boolean('${VOLATILE_LOG_DIR}') else 'log'}
```
ricardosalveti, line 27: This will call helper with empty inputs when VOLATILE_LOG_DIR is not set, we should probably have an if here instead to avoid surprises later on.

quaresmajose: I have tested it explicit call one only with just `ostree_rmdir_helper` and it works with an empty arg, because it check if the arg is a directory and not a link before removing and when the arg is empty the functions does nothing.

PR #1150, opened 2023-05-19
base: linux-lmp-dev: update template to 6.4.0-rc2

MrCry0:
Update template variables to use 6.4.0-rc2 by default, as that is the current development release.


ricardosalveti (approved):
LGTM

PR #1164, opened 2023-06-02
base: aktualizr: set default curl timeout for ostree

ricardosalveti:
Ostree already has a default curl timeout set of 780 seconds, and also the support for fetching a custom timeout via the environment variable OSTREE_CURL_TIMEOUT, so in order to allow user customization, set OSTREE_CURL_TIMEOUT to 780 by default.


mike-scott (approved):
LGTM

MrCry0 (approved):

mike-sul:
I don't think it's safe to allow OSTREE_CURL_TIMEOUT customization. Please, see the slack thread in the `ota` channel for details.

angolini (approved):

mike-scott:
@mike-sul @ricardosalveti not sure if this should be dropped or re-designed or we fix the upstream issue with the timeout?

mike-sul:
> if this should be dropped or re-designed or we fix the upstream issue
with the timeout?

Let's take it offline.

PR #1165, opened 2023-06-03
base: uboot-fitimage: avoid using obsolete egrep

MrCry0:
egrep is obsolescent. It is now an alias for "grep -E" [1]. Update the command to avoid wrapping and showing warning.

[1]
$ egrep -V
egrep: warning: egrep is obsolescent; using grep -E grep (GNU grep) 3.8


angolini (approved):

ricardosalveti (approved):
LGTM

PR #1170, opened 2023-06-07
base: custom-sota-client : Add the missing configuration folder

caiotpereira:
The custom-sota-client requires the folder /etc/sota/conf.d/ available. Otherwise, the service will not start. 
Install an empty folder /etc/sota/conf.d/ .


quaresmajose (approved):

ricardosalveti (approved):
LGTM

PR #1172, opened 2023-06-08
[all relevant imx] Switch to kernel 6.1

MrCry0:
This PR switches all imx machines that use linux-lmp-fslc-imx kernel to 6.1.

This version includes both arm64/arm32 archs.

This PR is splitter to smaller ones:
https://github.com/foundriesio/meta-lmp/pull/1194
https://github.com/foundriesio/meta-lmp/pull/1195

### DIFF
--- a/meta-lmp-bsp/recipes-kernel/linux/linux-lmp-fslc-imx_6.1.bb
+++ b/meta-lmp-bsp/recipes-kernel/linux/linux-lmp-fslc-imx_6.1.bb
@@ -30,5 +30,3 @@ SRC_URI:append:imx8mm-lpddr4-evk = " \
 SRC_URI:append:mx8mn-nxp-bsp = " \
     file://0001-FIO-internal-arm64-dts-imx8mn-evk.dtsi-re-add-blueto.patch \
 "
-
-DEFAULT_PREFERENCE = "-1"


meta-lmp-bsp/conf/machine/apalis-imx8.conf
```diff
@@ -55,7 +55,6 @@ IMX_DEFAULT_BOOTLOADER = "u-boot-fio"
 
 MACHINE_FIRMWARE:append = " linux-firmware-pcie8997"
 MACHINE_FIRMWARE:append = " firmware-imx-vpu-imx8"
-MACHINE_FIRMWARE:append = " firmware-imx-8"
```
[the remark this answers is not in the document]
MrCry0: > on main-next I have replace this with `imx-boot-firmware-files`
> 
> [b1b6f87](https://github.com/foundriesio/meta-lmp/commit/b1b6f8708a82e8a783d9c2523abcc8613748f24c)

IIUC It is handled properly upstream: https://github.com/Freescale/meta-freescale/pull/1514/commits/90df14f9af28fd55783a689a00ac9e20f674ca4e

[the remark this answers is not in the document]
quaresmajose: Thanks, I will drop it too instead of the replace.

quaresmajose (approved):

meta-lmp-bsp/conf/machine/apalis-imx8.conf
```diff
@@ -55,7 +55,8 @@ IMX_DEFAULT_BOOTLOADER = "u-boot-fio"
...
```
angolini, line 58: I'm a bit confused as firmware-imx-8 seems to be dropped only on meta-freescale master, and we are still working with meta-freescale kirkstone, right?

angolini (approved):
LGTM

meta-lmp-bsp/conf/machine/apalis-imx8.conf
MrCry0: > I'm a bit confused as firmware-imx-8 seems to be dropped only on meta-freescale master, and we are still working with meta-freescale kirkstone, right?

meta-freescale kirkstone doesn't include support for NXP BSP LF6.1, so we should switch to master for that. Otavio suggested doing this.

quaresmajose:
If we are ready to switch it to all kernels we can drop the `DEFAULT_PREFERENCE = "-1"` on the 6.1 recipe and with that we don't need anymore the `PREFERRED_VERSION_linux-lmp-fslc-imx`.

MrCry0:
> If we are ready to switch it to all kernels we can drop the `DEFAULT_PREFERENCE = "-1"` on the 6.1 recipe and with that we don't need anymore the `PREFERRED_VERSION_linux-lmp-fslc-imx`.

@quaresmajose thanks, I need to sleep a little more :)

meta-lmp-bsp/conf/machine/apalis-imx8.conf
angolini: But is this on preparation for switching to master?
I think we would still use a lmp linux recipe, so it's not dependent on the meta-freescale being on master or kirkstone

quaresmajose:
> > If we are ready to switch it to all kernels we can drop the `DEFAULT_PREFERENCE = "-1"` on the 6.1 recipe and with that we don't need anymore the `PREFERRED_VERSION_linux-lmp-fslc-imx`.
> 
> @quaresmajose thanks, I need to sleep a little more :)
> 
> Done.

I think I'm the one who needs to sleep a bit too :)
My suggestion only work if we remove the old version git recipe `linux-lmp-fslc-imx_git.bb` because bitbake prefer the `git` version over the numeric, so what you have before was better.

MrCry0:
> > > If we are ready to switch it to all kernels we can drop the `DEFAULT_PREFERENCE = "-1"` on the 6.1 recipe and with that we don't need anymore the `PREFERRED_VERSION_linux-lmp-fslc-imx`.
> > 
> > 
> > @quaresmajose thanks, I need to sleep a little more :)
> > Done.
> 
> I think I'm the one who needs to sleep a bit too :) My suggestion only work if we remove the old version git recipe `linux-lmp-fslc-imx_git.bb` because bitbake prefer the `git` version over the numeric, so what you have before was better.

@quaresmajose just tested. Nope, it chooses 6.1 :)

But yeah, I doubt the optimal numerating. @ricardosalveti what do you suggest? _git and _6.1? _5.15 and _6.1? 5.15 and _git?

I can't drop 5.15 since the -rt recipe is still on it. Once we bump -rt to 6.1 too, we can drop 5.15 for linux-lmp-fslc-imx.

meta-lmp-bsp/conf/machine/apalis-imx8.conf
MrCry0: > But is this on preparation for switching to master? I think we would still use a lmp linux recipe, so it's not dependent on the meta-freescale being on master or kirkstone
> 
> in case this is already for the preparation to switch to master, we also need: [angolini@174080a](https://github.com/angolini/meta-lmp/commit/174080a0bbe11c5c7ece14606a645bbc9d33676a)
> 
> But I would prefer to keep the "switching to master" PR separated from the "updating kernel to 6.1" PR

Yes, we have to switch to master. It is still compatible with kirkstone and supports NXP BSP LF6.1

quaresmajose:
> @quaresmajose just tested. Nope, it chooses 6.1 :)

I have also used the current state of this PR (without the `DEFAULT_PREFERENCE = "-1"`) on main-next and it build fine using the 6.1.

ricardosalveti:
> But yeah, I doubt the optimal numerating. @ricardosalveti what do you suggest? _git and _6.1? _5.15 and _6.1? 5.15 and _git?

We can drop _git (and 5.15 as part of this commit as well), better start having version specific recipes for kernel.

ricardosalveti:
Needs rebase.

Which meta-freescale master rev were you using here? So I can test it all together.

MrCry0:
> Needs rebase.
> 
> Which meta-freescale master rev were you using here? So I can test it all together.

For building, I use the HEAD of master. For now, 2805e245f6d49d2db4028698dcd116a38c37235f

Last time I tested when meta-freescale was just few commits behind.

MrCry0:
> Needs rebase.

@ricardosalveti done

PR #1173, opened 2023-06-08
Further patches for kernel 6.1

MrCry0:
Add more patches to fix issues with building imx8mp and imx8mq:
- select hdmi helper cdns-hdmi-core driver
- select v4l2_* modules for mxc-mipi-csi2_yav driver


quaresmajose (commented):
These patches are in the layer but are not used on the kernel recipe

MrCry0:
> These patches are in the layer but are not used on the kernel recipe

Thanks! Fixed.

quaresmajose (approved):

PR #1198, opened 2023-06-27
bsp: u-boot-ti-staging: am62: boot on closed boards

ldts:
Disable signature verification based on fused keys.

The kernel/dtb is therefore not verified and susceptible of being attacked.

This is a temporay solution.


ricardosalveti (approved):
LGTM

PR #1199, opened 2023-06-28
Changes for TI BSP 09.00.00.004

ricardosalveti:


quaresmajose (approved):
LGTM, tested the build on lmp-wayland and lmp-xwayland distro flavors

PR #1201, opened 2023-06-28
k3: u-boot: place root of trust keys outside the source repository

ldts:
Allow u-boot's binman to read the signing keys from anywhere in the filesystem.
i.e: specifying the following config option in uboot:

`CONFIG_SIGN_KEY_PATH="/tmp/`

### DIFF
--- a/meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
+++ b/meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
@@ -5,6 +5,7 @@ include recipes-bsp/u-boot/u-boot-lmp-common.inc
 SRC_URI:append = " \
     ${@bb.utils.contains('MACHINE_FEATURES', 'jailhouse', 'file://0003-HACK-lib-lmb-Allow-re-reserving-post-relocation-U-Bo.patch', '', d)} \
     file://lib-zlib-Fix-a-bug-when-getting-a-gzip-header-extra-field.patch \
+    file://k3-accept-filesystem-path-to-the-RoT-key.patch \
 "
 
 SRC_URI:append:am64xx-evm = " \
--- a/meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
+++ b/meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
@@ -27,3 +27,7 @@ python() {
     # we need to set the DEPENDS as well to produce valid SPDX documents
     fix_deployed_depends('do_install', d)
 }
+
+# Root of Trust Key directory
+K3_ROT_KEYS ?= "CONFIG_SIGN_KEY_PATH=${TOPDIR}/conf/keys"
+EXTRA_OEMAKE += "$K3_ROT_KEYS"
--- a/meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging/k3-accept-filesystem-path-to-the-RoT-key.patch
+++ b/meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging/k3-accept-filesystem-path-to-the-RoT-key.patch
@@ -1,0 +1,74 @@
+From 7afececd745b26a68822fe6c37addb0492356714 Mon Sep 17 00:00:00 2001
+From: Jorge Ramirez-Ortiz <jorge@foundries.io>
+Date: Wed, 28 Jun 2023 10:09:03 +0200
+Subject: [PATCH] k3: accept filesystem path to the RoT key
+
+Upstream-Status: Pending
+
+Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>
+---
+ arch/arm/dts/Makefile       | 10 ++++++++--
+ arch/arm/dts/k3-binman.dtsi |  4 ++--
+ board/ti/common/Kconfig     |  7 +++++++
+ 3 files changed, 17 insertions(+), 4 deletions(-)
+
+diff --git a/arch/arm/dts/Makefile b/arch/arm/dts/Makefile
+index 1b9c5f04c3..01b5ae36fe 100644
+--- a/arch/arm/dts/Makefile
++++ b/arch/arm/dts/Makefile
+@@ -1375,8 +1375,14 @@ targets += $(dtb-y)
+ # Add any required device tree compiler flags here
+ DTC_FLAGS += -a 0x8
+ 
+-PHONY += dtbs
+-dtbs: $(addprefix $(obj)/, $(dtb-y))
++PHONY += update_sign_keys dtbs
++
++update_sign_keys:
++ifeq ($(CONFIG_ARCH_K3),y)
++	${shell sed -i "s|TI_KEYS|$(CONFIG_SIGN_KEY_PATH)|g" $(srctree)/arch/arm/dts/k3-binman.dtsi}
++endif
++
++dtbs: update_sign_keys $(addprefix $(obj)/, $(dtb-y))
+ 	@:
+ 
+ clean-files := *.dtb *.dtbo *_HS
+diff --git a/arch/arm/dts/k3-binman.dtsi b/arch/arm/dts/k3-binman.dtsi
+index 97a3573bdb..c595203e9b 100644
+--- a/arch/arm/dts/k3-binman.dtsi
++++ b/arch/arm/dts/k3-binman.dtsi
+@@ -13,14 +13,14 @@
+ 	custMpk {
+ 		filename = "custMpk.pem";
+ 		blob-ext {
+-			filename = "../keys/custMpk.pem";
++			filename = "TI_KEYS/custMpk.pem";
+ 		};
+ 	};
+ 
+ 	ti-degenerate-key {
+ 		filename = "ti-degenerate-key.pem";
+ 		blob-ext {
+-			filename = "../keys/ti-degenerate-key.pem";
++			filename = "TI_KEYS/ti-degenerate-key.pem";
+ 		};
+ 	};
+ };
+diff --git a/board/ti/common/Kconfig b/board/ti/common/Kconfig
+index 49edd98014..4ff12d6c5b 100644
+--- a/board/ti/common/Kconfig
++++ b/board/ti/common/Kconfig
+@@ -49,3 +49,10 @@ config TI_COMMON_CMD_OPTIONS
+ 	imply CMD_SPI
+ 	imply CMD_TIME
+ 	imply CMD_USB if USB
++
++config SIGN_KEY_PATH
++       string "Path to the sign key"
++       depends on ARCH_K3
++       default "../keys"
++       help
++         Path to the folder containing the sign keys
+-- 
+2.34.1
+


meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
```diff
@@ -17,4 +18,15 @@ SRC_URI:append:am62xx-evm = " \
     file://lmp.cfg \
 "
 
+PATCHTOOL = "git"
```
ricardosalveti, line 21: Is this needed?

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
```diff
@@ -17,4 +18,15 @@ SRC_URI:append:am62xx-evm = " \
...
+
 PACKAGECONFIG[optee] = "TEE=${STAGING_DIR_HOST}${nonarch_base_libdir}/firmware/tee-pager_v2.bin,,optee-os-fio"
+
+python () {
+    if os.path.exists("${TOPDIR}/conf/keys/custMpk.pem"):
```
ricardosalveti, line 26: So here we're hardcoding the key name, we should have it under a variable instead.

Since this is the RoT key, we can probably make it specific to it, like:

K3_ROT_KEYNAME
K3_ROT_KEYDIR

And set it to ?= "" by default, allowing the customer to replace it.

ricardosalveti:
This would be yet another key to document (@angolini ).

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ldts: it helps since we are updating file in the source tree (a dtsi file)

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ldts: um I dont think we should replace the name of the ke:  it means something for TI, is documented in KeyWriter; the way I see it it doesnt bring any benefits being able to replace the name.

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ricardosalveti: Sure, but if not needed, I would prefer for it to not be added as it is yet another line we would have to maintain in the bbappend.

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ricardosalveti: Would it make any difference in the end? If you would prefer having a hardcoded name, then we should have at least a customized K3_ROT_KEYDIR (as we would have one dir for the public build and another for the factory build).

ldts: I just don't think changing the name adds any value - remember that this key must match the one in the keywriter software package (which was given the same name). Matching names just keeps it coherent and easier to follow.

you mean:
`"${TOPDIR}/${K3_ROT_KEYDIR}/custMpk.pem"` ?

angolini:
> This would be yet another key to document (@angolini ).

@ldts @ricardosalveti I can work on this. Is this key being created with a new factory yet?

angolini (approved):
no other comments - LGTM

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ricardosalveti: Sure.

In this case it would be the entire dir, as we have for other similar variables (https://github.com/foundriesio/partner-setup/blob/main/partner_api/templates/meta-subscriber-overrides/conf/machine/include/lmp-factory-custom.inc).

Something like:

`K3_ROT_KEYDIR = "${TOPDIR}/conf/factory-keys"`

Then when using it you can just append /custMpk.pem.

ricardosalveti:
> @ldts @ricardosalveti I can work on this. Is this key being created with a new factory yet?

Not yet, we would still need a change in partner-setup.

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ldts: ah I get your point, doing it now

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
```diff
@@ -17,12 +18,22 @@ SRC_URI:append:am62xx-evm = " \
...
 python() {
     # we need to set the DEPENDS as well to produce valid SPDX documents
     fix_deployed_depends('do_install', d)
+
+    if os.path.exists("${TOPDIR}/conf/keys/custMpk.pem"):
+        d.setVar('K3_ROT_KEYS', 'CONFIG_SIGN_KEY_PATH=${TOPDIR}/conf/keys')
+    else:
+        d.setVar('K3_ROT_KEYS', '')
 }
+
```
quaresmajose, line 38: We can consider the lmp keys the default ones and if the user whats to use the TI keys they just need to unset it with `K3_ROT_KEYS=""` in another bbappend
```
K3_ROT_KEYS ?= "CONFIG_SIGN_KEY_PATH=${TOPDIR}/conf/keys"

EXTRA_OEMAKE += "$K3_ROT_KEYS"
```

quaresmajose: With that we don't need the python snip to setVar

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
quaresmajose: by default bitbake uses the quilt for patching but it works without issues in any source tree, one diference is with quilt it don't produce commits in the source tree

ldts: Not having git in a tree that we patch and then update with sed is confusing when working/developing on local trees. So in terms of usability I'd say we have to have this.

But now, can you explain a bit more what does the maintenance task involve? why is having this line an issue? I dont understand sorry...

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ldts: It is not so much saving a python snip but the logic that I disagree with.

IMO the default keys should be the u-boot keys in the project.  And by default I mean them being available without having to specify any options (set or unset)

And then, if the user prefers to use their own keys, then let them figure out how by finding out the variable they need to set.

Preemptively asking them to unset a variable is a time waster IMO.

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ricardosalveti: @quaresmajose forgot about that one, this will change the commit hash which is also later part of the u-boot version string, which is indeed not desirable.

ricardosalveti: > Not having git in a tree that we patch and then update with sed is confusing when working/developing on local trees. So in terms of usability I'd say we have to have this.

For dev it makes sense, but as @quaresmajose wrote above, this will change the git version identifier as it gets applied as a commit, which will also change the hash string that is part of the u-boot version (which you can read during boot), which is not ideal (and why OE uses quilt by default, to avoid changing the underlying tree identifier).

angolini:
> > @ldts @ricardosalveti I can work on this. Is this key being created with a new factory yet?
> 
> Not yet, we would still need a change in partner-setup.

Please, try to remember to mention me on that PR when that happens. So I can start the doc change.

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
quaresmajose: Another consequence of this is breaking the binary reproducibility because it will change the commit hash randomly when applying the patch. Because of this each build will have a different hash and as the hash is encoded on the u-boot version the binary is not the same.

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
quaresmajose: I also think it's preferable for the process to be aligned with all the other board and because to build meta-lmp we have a strict dependency on the lmp-manifest the keys will always be there.
To use the key available on the tree the user can set `CONFIG_SIGN_KEY_PATH="../keys"` or don't use at all the `CONFIG_SIGN_KEY_PATH` on the recipe.

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging_%.bbappend
ldts: I don't see how this is a problem since releases are based on manifests and issues are reported based on releases and not tree commits.
But anyway, will drop.

ldts:
changes done as requested.

MrCry0 (approved):

meta-lmp-bsp/dynamic-layers/meta-ti-bsp/recipes-bsp/u-boot/u-boot-ti-staging/0001-k3-path-to-the-sign-key.patch
```diff
@@ -0,0 +1,72 @@
+From 5516b9366be6a81b036874233e83f42a584319f4 Mon Sep 17 00:00:00 2001
+From: Jorge Ramirez-Ortiz <jorge@foundries.io>
+Date: Wed, 28 Jun 2023 10:09:03 +0200
+Subject: [PATCH] k3: path to the sign key
+
```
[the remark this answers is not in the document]
ricardosalveti: @ldts it seems this is still missing?

MrCry0 (approved):

quaresmajose (approved):
LGTM

PR #1204, opened 2023-06-29
bsp: linux-firmware: add SRCREV_FORMAT

quaresmajose:
The SRCREV_FORMAT variable must be set when multiple SCMs are used.

| bb.data_smart.ExpansionError: Failure expanding variable SRCPV, expression was ${@bb.fetch2.get_srcrev(d)} which triggered exception FetchError: Fetcher failure: The SRCREV_FORMAT variable must be set when multiple SCMs are used. | The SCMs are:
| git://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git;protocol=https;branch=main | git://github.com/NXP/imx-firmware.git;protocol=https;branch=lf-5.15.71_2.2.0;destsuffix=imx-firmware;name=imx-firmware; | The variable dependency chain for the failure is: SRCPV -> PV -> BP -> FILESPATH


meta-lmp-bsp/recipes-kernel/linux-firmware/linux-firmware_%.bbappend
```diff
@@ -18,6 +18,7 @@ SRC_URI:append:imx-nxp-bsp = "\
     git://github.com/NXP/imx-firmware.git;protocol=https;branch=${IMX_FIRMWARE_BRANCH};destsuffix=imx-firmware;name=imx-firmware; \
 "
 SRCREV_imx-firmware ?= "982bb10dfabfb9e7b9dc106c59a4fbb2c45bfb44"
+SRCREV_FORMAT = "imx-firmware"
```
ricardosalveti, line 21: Should we use just the main SRCURI or combine both here? In the end this imx-firmware is just an extra.

quaresmajose: This was a quick fix on main-next, maybe the better solution would be to submit upstream all those we have in this bbapppend.

quaresmajose: It should be the two but because main repo uses just one they don't set the name so we don't know it.
We can add a name to the default to use the two but look a bit ugly:
```
SRC_URI:remove =  "git://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git;protocol=https;branch=main"
SRC_URI:append = " git://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git;protocol=https;branch=main;name=main"
SRCREV_FORMAT = "main_imx-firmware"
```

ricardosalveti: Right, I think the right thing to do here would instead remove all these extra appends (as we have logic here for multiple socs) into separated recipes, as then we would probably be able to drop this append entirely.

ricardosalveti: Since I don't think it makes sense for the stm32mp15 build to have a linux-firmware package that has the hash from imx-firmware, maybe to start we could duplicate the original SRC_URI to add a name to it, and have SRCREV_FORMAT = "main" only, what do you think?

quaresmajose: Currently this only affect main-next and what I have on SRCREV_FORMAT solves the problem for now but as I am here it will be better to separated it to new recipes so I will do that.

ricardosalveti: Thanks, it would be a nice change for us anyway :-)

quaresmajose:
This is not need anymore

PR #1205, opened 2023-06-29
Add basic support for beagleplay

ricardosalveti:
Board is am62xx based.


meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -647,45 +647,47 @@ WKS_FILE_DEPENDS:append:k3 = " virtual/bootloader"
...
+PREFERRED_PROVIDER_u-boot-default-script:sota:am62xx = "u-boot-ostree-scr-fit"
+KERNEL_CLASSES:sota:am62xx = " kernel-lmp-fitimage "
+IMAGE_BOOT_FILES:append:am62xx = " boot.itb"
+UBOOT_ENTRYPOINT:am62xx = "0x82000000"
+UBOOT_LOADADDRESS:am62xx = "0x82000000"
+UBOOT_DTB_LOADADDRESS:am62xx = "0x88000000"
+UBOOT_DTBO_LOADADDRESS:am62xx = "0x89000000"
+EFI_PROVIDER:am62xx = ""
+LMP_BOOT_FIRMWARE_FILES:am62xx = "tispl.bin u-boot.img tiboot3.bin"
+LMP_BOOT_FIRMWARE_FILES:append:sota:am62xx = " boot.itb"
```
quaresmajose, line 664: This AM62x block is also common with the AM64x with exception of the UBOOT_DTBO_LOADADDRESS that is not defined on the AM64x.

quaresmajose (approved):

meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
ricardosalveti: Yup, but we don't have a common am6x override which we can use for both, for example.

quaresmajose: Right, this part is missing.

PR #1214, opened 2023-07-04
base: layer.conf: exclude os-release siggen dependencie on the mfgtool-initramfs

quaresmajose:
This patch is a follow-up of 24d1016e, the fsl-image-mfgtool-initramfs is used on the distro lmp-mfgtool and without this patch the initramfs is rebuilded which consequently triggers a kernel rebuild.


MrCry0 (approved):

ricardosalveti (approved):
LGTM

PR #1217, opened 2023-07-05
base: fioconfig: Support SOTA_CLIENT customization on aktualizr-toml-update handler.

caiotpereira:
This commit fixes the `fioconfig` handler allowing to restart of the  SOTA CLIENT service when the Custom SOTA Client is used instead `aktualizr-lite`.

Previously the `fioconfig` handler had a hard-coded `systemctl restart aktualizr-lite` . 
In cases where the Custom SOTA Client is used and the `fioconfig` receives any configuration from `fioctl` , it was impossible to restart the Custom SOTA Client service to apply the new configurations.

- Bump `fioconfig` rev to `c9469923f88cec496a9abce194d698fda0f3573d`.
- The final image still retains the `aktualizr-toml-update` handler filename.

### DIFF
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
@@ -11,7 +11,7 @@ SRC_URI = "git://${GO_IMPORT};protocol=${GO_IMPORT_PROTO};branch=main \
 	file://fioconfig.path \
 	file://fioconfig-extract.service \
 "
-SRCREV = "0b14975bbcfa85590432a9e71b1eb0125d81a92a"
+SRCREV = "c9469923f88cec496a9abce194d698fda0f3573d"
 
 UPSTREAM_CHECK_COMMITS = "1"
 
@@ -34,6 +34,8 @@ do_install:append() {
 	install -m 0755 ${S}/src/${GO_IMPORT}/contrib/aktualizr-toml-update ${D}${datadir}/fioconfig/handlers
 	install -m 0755 ${S}/src/${GO_IMPORT}/contrib/factory-config-vpn ${D}${datadir}/fioconfig/handlers
 	install -m 0755 ${S}/src/${GO_IMPORT}/contrib/renew-client-cert ${D}${datadir}/fioconfig/handlers
+	sed -i -e "s/SOTA_CLIENT=aktualizr-lite/SOTA_CLIENT=${SOTA_CLIENT}/g" \
+		 ${D}${datadir}/fioconfig/handlers/aktualizr-toml-update
 }
 
 # We need aktualizr because we uses its device gateway connectivity and keys,


Tim-Anderson (approved):
LGTM

MrCry0 (changes requested):
@caiotpereira please squash the commit that fixes the brackets.

also, we use rebase strategy instead of merging. Could you rebase your branch on top of the upstream `main`? The resulting PR should contain only one commit.

ricardosalveti:
Also please change the patch subject from meta-lmp-base to base only.

vkhoroz:
@caiotpereira https://github.com/foundriesio/fioconfig/pull/41 was merged.

Please, update this PR accordingly, and I will approve it.

mike-sul:
Also, this PR require another commit to bump the corresponding fioconfig version.

angolini (approved):
LGTM
when you are working on the other reviews, could you please add this sentence to the commit log?
"This pull request has been made based on the modifications found at https://github.com/foundriesio/fioconfig/commit/c9469923f88cec496a9abce194d698fda0f3573d."

quaresmajose (approved):

MrCry0 (approved):
LGTM.

In spite of a minor improvement that may be done:
I'd rewrite the commit message with an explanation of "why" instead of "what was done":
```
base: fioconfig: Support SOTA_CLIENT customization on aktualizr-toml-update handler

Bump fioconfig version and deploy a handler that supports customizing the SOTA_CLIENT.
```

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
```diff
@@ -11,7 +11,7 @@ SRC_URI = "git://${GO_IMPORT};protocol=${GO_IMPORT_PROTO};branch=main \
 	file://fioconfig.path \
 	file://fioconfig-extract.service \
 "
-SRCREV = "0b14975bbcfa85590432a9e71b1eb0125d81a92a"
+SRCREV = "c9469923f88cec496a9abce194d698fda0f3573d"
```
vkhoroz, line 14: Not sure if we usually do a _commit hash upgrade_ in a separate commit, or it is fine to combine it with code changes.

@ricardosalveti can give the guidance on this.

vkhoroz (approved):
LGTM

I like @MrCry0's idea. I'd make it even shorter:

```
base: fioconfig: Support SOTA_CLIENT customization on aktualizr-toml-update handler
```

The other sentence is _wordy details_.

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
```diff
@@ -24,14 +24,20 @@ GO_EXTRA_LDFLAGS = "-X ${GO_IMPORT}/internal.Commit=${SRCREV}"
 SYSTEMD_PACKAGES += "${PN}"
 SYSTEMD_SERVICE:${PN} = "fioconfig.service fioconfig-extract.service fioconfig.path"
 
+do_configure:append() {
+    sed -e "s/SOTA_CLIENT=aktualizr-lite/SOTA_CLIENT=${SOTA_CLIENT}/g" \
+        ${S}/src/${GO_IMPORT}/contrib/aktualizr-toml-update \
+        > ${S}/src/${GO_IMPORT}/contrib/aktualizr-toml-update-customized
+
```
vkhoroz, line 31: No need for this black line, right?

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
ricardosalveti: As in this case the only change involved is what allowed the recipe / packaging change, it is fine to have in a single commit.

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
ricardosalveti: Yeah, please drop the extra line.

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
```diff
@@ -24,14 +24,20 @@ GO_EXTRA_LDFLAGS = "-X ${GO_IMPORT}/internal.Commit=${SRCREV}"
 SYSTEMD_PACKAGES += "${PN}"
 SYSTEMD_SERVICE:${PN} = "fioconfig.service fioconfig-extract.service fioconfig.path"
 
+do_configure:append() {
```
ricardosalveti, line 27: Please perform this change on do_install:append as it it won't have any implication at the configure and compile steps (it is just a script file).

That way we can reuse the cache from configure and compile when changing SOTA_CLIENT.

ricardosalveti: Also, there is no need to create a custom file, just perform sed -i instead.

ricardosalveti:
Please also update the commit message:

`base: fioconfig: Bump SRC_URI and update the sed regex to replace the SOTA_CLIENT.`

Use `base: fioconfig: bump rev ...' instead.

SRC_URI is the same in the case, what got changed is the revision.

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
vkhoroz: @ricardosalveti we've discussed the `-i` thing in https://github.com/foundriesio/fioconfig/pull/41.

A problem with it is this: during development, once an `aktualizr-toml-update` file was updated in local environment, subsequent changes to SOTA_CLIENT have no effect (as the file content differs). The `-customized` intermediary mitigates this problem, as the original file remains unmodified.

vkhoroz: We can put the `-customized` thing into `/tmp`, but I have no idea what ramifications it might have; so IMHO it's not worth trying that.

ricardosalveti: Oh sure, for local dev I wouldn't suggest using sed -i indeed, but with yocto it is usually better to avoid creating an extra file unless done as part of the respective build steps (e.g. generate a new file in do_compile and later install this new file with the common name under do_install). This is needed as yocto cache works per task, and usually do_config is done just to prepare do_compile, which can generate whatever content that is needed, and later do_install is responsible for moving stuff create as part of do_compile into the deploy folder (which is later transformed into packages based on FILES).

ricardosalveti: Since here we're just replacing a script variable, we can do as do_install:append and reuse the task cache from configure and compile.

vkhoroz: I'll leave it to you guys, I'm personally fine either way. Just told you the concerns @caiotpereira expressed earlier regarding `-i`.

vkhoroz: Doing this inside `do_install:append` definitely makes sense to me.

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
caiotpereira: In this case, the SRCREV bump is aligned with the changes. since the changes are minor on `fioconfig` side and here, I don´t think it requires an extra PR.

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
caiotpereira: Previously I was using a `template` folder to avoid this situation. 
@ricardosalveti @mike-sul  should we move the new handler to the `template` folder or manipulate the `FILES:${PN}` to include the correct files?
What do you think about your suggestion?

quaresmajose: > Hello @ricardosalveti, I noticed that you used sed -i for the test, but this could potentially cause problems for those who build locally and decide to change the SOTA_CLIENT after the first build. To avoid any cache issues, we have opted to keep the template file and copy it with the same name as the previous implementation.

if the `SOTA_CLIENT`  change after the build the task that uses this variable needs to run again and if this not happens we need to add this variable to the hash using the `do_task[vardeps] += "SOTA_CLIENT"`. I will test it locally to see if it works.

quaresmajose (approved):

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
quaresmajose: Hello @ricardosalveti, I noticed that you used sed -i for the test, but this could potentially cause problems for those who build locally and decide to change the SOTA_CLIENT after the first build. To avoid any cache issues, we have opted to keep the template file and copy it with the same name as the previous implementation.

if the SOTA_CLIENT change after the build the task that uses this variable needs to run again and if this not happens we need to add this variable to the hash using the do_task[vardeps] += "SOTA_CLIENT". I will test it locally to see if it works.

quaresmajose: tested changing the install task and it works:
```
--- a/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
+++ b/meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
@@ -11,7 +11,7 @@ SRC_URI = "git://${GO_IMPORT};protocol=${GO_IMPORT_PROTO};branch=main \
        file://fioconfig.path \
        file://fioconfig-extract.service \
 "
-SRCREV = "0b14975bbcfa85590432a9e71b1eb0125d81a92a"
+SRCREV = "c9469923f88cec496a9abce194d698fda0f3573d"
 
 UPSTREAM_CHECK_COMMITS = "1"
 
@@ -34,6 +34,8 @@ do_install:append() {
        install -m 0755 ${S}/src/${GO_IMPORT}/contrib/aktualizr-toml-update ${D}${datadir}/fioconfig/handlers
        install -m 0755 ${S}/src/${GO_IMPORT}/contrib/factory-config-vpn ${D}${datadir}/fioconfig/handlers
        install -m 0755 ${S}/src/${GO_IMPORT}/contrib/renew-client-cert ${D}${datadir}/fioconfig/handlers
+       sed -i -e "s/SOTA_CLIENT=aktualizr-lite/SOTA_CLIENT=${SOTA_CLIENT}/g" \
+               ${D}${datadir}/fioconfig/handlers/aktualizr-toml-update
 }
```

To check if the task will re-run whan changing SOTA_CLIENT we can:

1. Building the default with `bitbake fioconfig` to rebuild the package and the sstate-cache.
2. change SOTA_CLIENT to use a difent value: `echo 'SOTA_CLIENT = "aktualizr"' >> conf/auto.conf`
3. check the sigantures with: `bitbake fioconfig --dump-signatures=printdiff`

```
Task fioconfig:do_install couldn't be used from the cache because:
  We need hash 57ab39a5ea88d43f312e2510a162a32998d0cfc5106cf361c38be33220f6041a, closest matching task was 229ebc24493e77a08df14f3fc944059684b6431dbc2eaf8238fafbe3950f7a0b
  basehash changed from dbfe15beea7b6bc891c1ab9f3788e7a70ac1efd9e6187507a86b311086d3d5b6 to 30f9bb8bcde7eb0f8d539de1410bdb193f8b5607f2c5b1b7d67e36bd7bc3a63b
  Variable SOTA_CLIENT value changed from 'aktualizr-lite' to 'aktualizr'
```

quaresmajose (changes requested):

meta-lmp-base/recipes-support/fioconfig/fioconfig_git.bb
vkhoroz: This looks like copy-on-write; clever and nice.

caiotpereira: Great @quaresmajose , thanks , nice idea!

quaresmajose (approved):
LGTM

ricardosalveti:
Will merge after the release.

mike-sul (approved):

PR #1223, opened 2023-07-10
bsp: u-boot-fio-mfgtool: imx8mn-lpddr4-evk: set CONFIG_SPL_FIT_IMAGE_…

ricardosalveti:
…TINY

Required in order to skip the recording of each loaded payload into the FDT, which might cause boot issues (Synchronous Abort).

Config got removed as a side effect of the u-boot update to 2022.04+lf-6.1.1-1.0.0-fio.


ricardosalveti:
Replaced by https://github.com/foundriesio/meta-lmp/pull/1224 (it also affects ddr4).

PR #1224, opened 2023-07-11
bsp: u-boot-fio-mfgtool: imx8mn: fix config regression

ricardosalveti:
Fix config regression as part of 8d1e06f6, which resulted in the simplified version to be applied to only one variant instead of both.


ricardosalveti:
Merging directly in order to create v91 RC3.

PR #1240, opened 2023-07-20
base: lmp-device-register: bump to 3a99d9b

ldts:
Relevant changes:
-3a99d9b options: debug info: fix source of tag and factory definition


ricardosalveti (approved):
LGTM

PR #1255, opened 2023-08-08
base: lmp: remove ptest from distro feauture

quaresmajose:
It will be better to dropping this as it build a lot of stuff used only in testing, when we decide to run on the target the unit test produced with ptest we can enable it again.


ricardosalveti (approved):
LGTM

PR #1257, opened 2023-08-09
systemd: build coredump

quaresmajose:
The `systemd-coredump` is needed to collect possible coredumps and `coredumpctl` to analize the collect data.


ricardosalveti:
@quaresmajose I would prefer having this isolate do a debug / dev mode, and not have coredump enabled by default (due stack traces, core dumps being stored locally, etc), can cause several other side effects on production systems.

We do have a doc showing how to enable that with dev mode (https://github.com/foundriesio/docs/pull/577/files), maybe just add that package config conditional line here by default?

quaresmajose:
funny and the dev mode documentation use as example the coredump.

PR #1261, opened 2023-08-13
bsp: move linux-lmp-ti-staging under dynamic-layers

ricardosalveti:
Recipe requires files from meta-ti-bsp, so only make it available when the layer is required by the user.


mike-scott (approved):
LGTM.  Strictly a movement change right?

MrCry0 (approved):

ricardosalveti:
> LGTM. Strictly a movement change right?

Yup, having it under dynamic makes it relevant only when the respective layer is also enabled.

PR #1270, opened 2023-08-30
Revert "base: import ntp 4.2.8p17 from meta-oe master d2a38a5ec5"

quaresmajose:
This reverts commit 97c1e279ad96840a28a2c06a282f371c2aec3373.

Upstream have fixed this in [1]: ntp: backport patch for 5 CVEs CVE-2023-26551/2/3/4/5

[1] https://git.openembedded.org/meta-openembedded/commit/meta-networking/recipes-support/ntp?h=kirkstone&id=3c3755a3da1aa93501664f2893fa110e66ce7b6e


ricardosalveti:
While it is nice that things got backported, I'm not so sure if we should downgrade the version here, that is usually not the best way moving this forward.

PR #1273, opened 2023-09-01
meta-lmp: update compose apps early start

Tim-Anderson:
The assemble system image to add containers to wic image was updated awhile back to include the docker content.

The early start can take advantage of that and just do a docker compose up similar to the logic for compose-apps.

This updates the early start app to detect and activate the app if it is already loaded into docker.

### DIFF
--- a/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
+++ b/meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
@@ -41,14 +41,29 @@ start_restorable_apps() {
 
 	if [ -f $default_apps_path ] ; then
 		for app in `cat $default_apps_path`; do
-			shortlist="${shortlist}${app},"
+			if [ -e "/var/sota/compose-apps/${app}" ] ; then
+				start_compose_app ${app}
+			else
+				shortlist="${shortlist}${app},"
+			fi
 		done
 		if [ ! -z $shortlist ] ; then
 			shortlist=${shortlist::-1}
                 fi
 	fi
 
-	aklite-apps run --apps "${shortlist}"
+	if [ -z "${shortlist}" ] ; then
+		all=`ls /var/sota/compose-apps 2> /dev/null`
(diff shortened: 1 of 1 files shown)


quaresmajose:
using `base: compose-apps-early-start: update early start` on the commit title are more alligned with what we usually use.

meta-lmp-base/recipes-support/compose-apps-early-start/compose-apps-early-start/compose-apps-early-start
```diff
@@ -41,14 +41,29 @@ start_restorable_apps() {
...
 
-	aklite-apps run --apps "${shortlist}"
+	if [ -z "${shortlist}" ] ; then
+		all=`ls /var/sota/compose-apps 2> /dev/null`
+		if [ -n "${all}" ] ; then
+			for app in ${all} ; do
+				start_compose_app ${app}
+			done
+		else
+			aklite-apps run --apps "${shortlist}"
```
mike-sul, line 62: iirc, `skopeo` (invoked by aklite-apps`) doesn't inject layers that already exists in the docker store. If so, then there  shouldn't be much difference between `aklite-apps run` and `docker compose up` if images/apps are preloaded to the docker store.

Tim-Anderson:
the problem is skopeo is serial and docker compose does the startup in
parallel

ricardosalveti:
> The assemble system image to add containers to wic image was updated awhile back to include the docker content.
> 
> The early start can take advantage of that and just do a docker compose up similar to the logic for compose-apps.
> 
> This updates the early start app to detect and activate the app if it is already loaded into docker.

Please add this description as part of the commit message as well.

ricardosalveti:
> the problem is skopeo is serial and docker compose does the startup in parallel

Won't both calls be serialized the same way?

Tim-Anderson:
docker runs the starting of the containers in parallel. The time difference is significant on a single core 32 bit processor

Tim-Anderson:
Added a reply

ricardosalveti (approved):
LGTM, just please update the commit message with your PR description before merging it.

PR #1283, opened 2023-09-15
lmp-el2go-auto-register: allow to disable "composeapp"

olheureu:
On a device running a dockerless LmP with EdgeLock 2GO [1], the "aktualizr-lite" was failing and exiting with an error message:

  `aktualizr-lite[991]: boost::filesystem::canonical: No such file or directory [generic:2]: "/usr/bin/docker"`

We could trace the error in "`aktualizr-lite`": it happens only when "`aktualizr-lite`" tried to upgrade the "compose apps" on a dockerless system. But "`aktualizr-lite`" can work: while our device is dockerless, we could register it manually with "`lmp-device-register`" and upgrade the firmware with "`aktualizr-lite`".

The difference between the successfull "`aktualizr-lite`" when manually registering the device and the error mentioned above is the use of EdgeLock 2GO (el2go). The "`lmp-el2go-auto-register`" script [2] starts the EdgeLock 2GO agent and creates a "`/var/sota/sota.toml`" configuration file for "`aktualizr-lite`". That configuration file does not support dockerless systems because it forces the package manager in "`ostree+compose_apps`" mode:
```
  [...]
  [pacman]
  type = "ostree+compose_apps"
  ostree_server = "https://{REPO_ID}.ostree.foundries.io:8443/ostree"
  packages_file = "/usr/package.manifest"
  [...]
```

On a dockerless system, the package manager should be in "`ostree`" mode, i.e. not trying to upgrade the docker containers. The config file speaks about "pacman `type`".

Fixed: made the package manager type an option that depends on the "`PACMAN_TYPE`" environment variable, and be "`ostree+compose_apps`" by default, for backward compatibility. Dockerless systems can set the "`PACMAN_TYPE`" variable to "`ostree`" in the script environment.

References:
- [1] "_Creating an LmP Build With EdgeLock 2GO_" <https://docs.foundries.io/91/user-guide/el2g.html#creating-an-lmp-build-with-edgelock-2go>
- [2] "_lmp-el2go-auto-register_" <https://github.com/foundriesio/meta-lmp/blob/main/meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register#L199>

### DIFF
--- a/meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
+++ b/meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
@@ -19,6 +19,7 @@ PIN = os.environ.get("PKCS11_PIN", "87654321")
 SO_PIN = os.environ.get("PKCS11_SOPIN", "12345678")
 SOTA_DIR = os.environ.get("SOTA_DIR", "/var/sota")
 HANDLERS = os.environ.get("HANDLERS", "aws-iot,aktualizr-lite")
+PACMAN_TYPE = os.environ.get("PACMAN_TYPE", "ostree+compose_apps")
 
 REPO_ID = os.environ["REPOID"]
 
@@ -196,7 +197,7 @@ repo_server = "https://{REPO_ID}.ota-lite.foundries.io:8443/repo"
 key_source = "file"
 
 [pacman]
-type = "ostree+compose_apps"
+type = "{PACMAN_TYPE}"
 ostree_server = "https://{REPO_ID}.ostree.foundries.io:8443/ostree"
 packages_file = "/usr/package.manifest"
 tags = "{tag}"


ricardosalveti (approved):
LGTM, we can merge once @doanac reviews it as well.

meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
```diff
@@ -19,6 +19,7 @@ PIN = os.environ.get("PKCS11_PIN", "87654321")
 SO_PIN = os.environ.get("PKCS11_SOPIN", "12345678")
 SOTA_DIR = os.environ.get("SOTA_DIR", "/var/sota")
 HANDLERS = os.environ.get("HANDLERS", "aws-iot,aktualizr-lite")
+PACMAN_TYPE = os.environ.get("PACMAN_TYPE", "ostree+compose_apps")
```
doanac, line 22: This is fine, but a user would still have to make additional changes to bitbake to get this set, right?

We could probably do this automagically with something like:
```
from shutil import which
def _has_docker() -> bool:
    return which("docker") is not None
```
Then you can know at runtime what packagemanager to assign.

ricardosalveti: Indeed, we should support users with docker and without necessarily wanting to use composeapp.

We can for sure reuse the composeapp packageconfig and make it a distro feature, having both registration related recipes using the same package config.

Since this is a script, the generation of default.env could be done at build time, otherwise we would need some sort of sed to have it hardcoded.

doanac: k. Running composeapp+ostree doesn't require apps, but this could be theoretically possible. I guess what we need to avoid is allowing someone to ask for composeapp+ostree if docker isn't on the system. Maybe the recipe can detect that.

ricardosalveti: > Still, a good approach could be to add a basic (empty) environment file in meta-lmp-base and In the .bb file, append the `PACMAN_TYPE` line to it. Then we can simply copy that basic file to meta-subscriber-overrides, add whatever variables to it, and set `FILESEXTRAPATH` in bbappend.

Yeah, we already provide an empty default.env that was created to facilitate the process to override it via meta-sub, as you suggested, so I think we can land the current PR (since it won't change the default behavior) and still allow users to create a custom default.env with a different PACMAN_TYPE.

If needed we can extend this in a more complete approach later.

ricardosalveti:
Will merge as this won't change the default behavior and still allow a more experienced user to control the usage better, we can improve to make it via pkgconfig and to validate if the scenario is valid at runtime in another pr.

PR #1284, opened 2023-09-16
Patch to sync extracted layer files

mike-sul:
It turned out that the current patch we have at https://github.com/foundriesio/meta-lmp/blob/main/meta-lmp-base/recipes-containers/docker/files/0001-Fsync-layer-once-extracted-to-file-system.patch doesn't work. The code branch at which that sync call is invoked is not executed if the docker engine runs as a root user, and the underlying filesystem has NO specific issues, see https://github.com/moby/moby/blob/30a57db97c5e25d1512ed986638d003813a6ab2b/daemon/graphdriver/overlay2/check.go#L20 for details. This is true for the LmP case, so so-called "naive diff/graph driver" is NOT used in our case https://github.com/moby/moby/blob/30a57db97c5e25d1512ed986638d003813a6ab2b/daemon/graphdriver/overlay2/overlay.go#L677.

This new patch places the sync invocation at a spot that doesn't depend on the graph driver type ('naive' vs. overlay). Specifically, the sync is called just before committing a layer to the 'DB' on the filesystem, which, in my belief, is the right place for it.

Ideally, we would require something like StartUnpacking() and Commit() in the graph driver interface to extract and save a layer in a transactional/atomic way. However, since we don't want to make extensive changes to the docker engine codebase, simply invoking the system 'sync' call before committing layer metadata should suffice for our needs.


mike-sul:
@mwasilew FYI.

doanac (approved):

meta-lmp-base/recipes-containers/docker/files/0001-layer-ensure-layer-files-are-synced-to-disk.patch
```diff
@@ -0,0 +1,55 @@
...
+into the Docker store. As a result, software running in the container
+based on the layer with 'broken' files will behave incorrectly.
+
+Invoking the sync system call is crucial to prevent this issue,
+as the incorrect behavior of software running inside a container
+with 'broken' files is often not immediately apparent.
+Container engineers may begin debugging the container's software, unaware
+that the root cause lies in the fact that some of the container files are
+simply 'broken'.
+
```
quaresmajose, line 21: can please add the patch status?

```
Upstream-Status: Pending
```
until it is not subited upstream.

quaresmajose (changes requested):

mwasilew:
I tested this patch with usual set of power cuts and so far I'm not able to corrupt any layers. +1

meta-lmp-base/recipes-containers/docker/files/0001-layer-ensure-layer-files-are-synced-to-disk.patch
mike-sul: Done. However, I am not sure moby will accept this patch as it is because (citation from https://github.com/moby/moby/pull/45900#issuecomment-1629672564):

```
We would not be able to accept a PR which uses syscall.Sync() as it is such a blunt instrument which could have a severe performance impact on workloads in production servers with big fs caches and uninterruptible power supplies, but a more targeted fsync of just the unpacked files and directories may be more palatable.
```

I kind of agree with him because doing `sync` for each layer write on a big servers with huge disks and lots of volumes supposedly will have a huge impact on performance.

I've already tried and checked doing `fsync` on each file as he suggested. However, it causes significant performance degradation on a real device (rpi3), layer extraction takes significantly more time.

I'll create a jira ticket for this one and do another fight round for this issue, maybe I'll come up with something that makes both embedded and data center servers worlds happy.

mike-sul: https://foundriesio.atlassian.net/browse/FFTK-2815

quaresmajose: Understant and agree with you as it is very specific for our use case, maybe `Upstream-Status: Inappropriate [lmp specific]` is better there.

quaresmajose (approved):
LGTM

ricardosalveti (approved):
LGTM, thanks for working on this one!

PR #1286, opened 2023-09-21
base: docker-credential-helper: Handle leading spaces

doanac:
Normally `/var/sota/sota.toml` has had TOML with no indentation. eg: 
```
[section]
key=val
```

There are situations where /var/sota/sota.toml can be indented:
```
[section]
  key=val
```
This changes our regular expression to handle either one. Discovered while testing meds.


doanac:
FYI - I think the cause of my situation was doing device certificate rotation. fioconfig's toml library does the indentation.

ricardosalveti (approved):
LGTM

PR #1288, opened 2023-09-22
base: rc: Patch docker to load image optimally

mike-sul:
This patch helps to improve the image loading process that occurs at the App installation step. Specifically, image layers are not copied from the skopeo/OCI store to the docker temporal directory before their extraction to the docker store.
Instead, image layers are read directly from the skopeo store.

Also, this patch fixes a few minor issues the loader/tarexporter implementation.

To leverage the patch [this change to aklite](https://github.com/foundriesio/aktualizr-lite/pull/273. should be utilized.

### DIFF
--- a/meta-lmp-base/recipes-containers/docker/docker-ce_%.bbappend
+++ b/meta-lmp-base/recipes-containers/docker/docker-ce_%.bbappend
@@ -7,6 +7,7 @@ SRC_URI:append = " \
     file://0001-registry-increase-TLS-and-connection-timeouts.patch;patchdir=src/import \
     file://0001-daemon-overlay2-write-layer-metadata-atomically.patch;patchdir=src/import \
     file://0001-layer-ensure-layer-files-are-synced-to-disk.patch;patchdir=src/import \
+    file://0001-tarexport-optimize-image-loading-on-local-host.patch;patchdir=src/import \
     file://daemon.json.in \
     file://docker.service \
 "
--- a/meta-lmp-base/recipes-containers/docker/files/0001-tarexport-optimize-image-loading-on-local-host.patch
+++ b/meta-lmp-base/recipes-containers/docker/files/0001-tarexport-optimize-image-loading-on-local-host.patch
@@ -1,0 +1,180 @@
+From d737cd24d2922c0bde11be945e19a0febbec42cb Mon Sep 17 00:00:00 2001
+From: Mike Sul <mike.sul@foundries.io>
+Date: Fri, 22 Sep 2023 12:49:07 +0200
+Subject: [PATCH] tarexport: Optimize image loading on local host
+
+This commit augments the image load manifest by introducing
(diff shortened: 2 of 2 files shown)


meta-lmp-base/recipes-containers/docker/files/0001-tarexport-Load-image-without-layer-copying.patch
```diff
@@ -0,0 +1,106 @@
...
+ }
+diff --git a/image/tarexport/tarexport.go b/image/tarexport/tarexport.go
+index 5bcad2265c..f53a1c4be6 100644
+--- a/image/tarexport/tarexport.go
++++ b/image/tarexport/tarexport.go
+@@ -21,6 +21,7 @@ type manifestItem struct {
+ 	Layers       []string
+ 	Parent       image.ID                                 `json:",omitempty"`
+ 	LayerSources map[layer.DiffID]distribution.Descriptor `json:",omitempty"`
++	LayerRootDir string `json:"layer_root_dir,omitempty"`
```
mike-sul, line 100: @doanac Just heads up. By extending the manifest passed in the TAR archive/stream to `/image/load` (`docker load` command) we kind of instruct the docker's tar exporter to load layer blobs from a specific local directory. As result, no need to pass layer blobs through the TAR stream in the `HTTP POST /image/load` body.

meta-lmp-base/recipes-containers/docker/files/0001-tarexport-Load-image-without-layer-copying.patch
```diff
@@ -0,0 +1,106 @@
...
+ 			}
+-			ref, ok := named.(reference.NamedTagged)
+-			if !ok {
+-				return fmt.Errorf("invalid tag %q", repoTag)
++			switch ref := named.(type) {
++				case reference.NamedTagged: {
++					l.setLoadedTag(ref, imgID.Digest(), outStream)
++					outStream.Write([]byte(fmt.Sprintf("Loaded image with tag: %s\n", reference.FamiliarString(ref))))
++				}
++				case reference.Canonical: {
```
mike-sul, line 49: @doanac Just heads up. This allows us to pass a digest/hashed URI and set for the loaded/exported image in the repository file.

mike-sul:
This https://github.com/foundriesio/aktualizr-lite/pull/273 is the corresponding change in aklite.

It works e2e as expected and I am finalizing/cleaning the implementation atm.

doanac (commented):
nice!

ricardosalveti (approved):
LGTM, not that familiar with the codebase but change seems sane.

Since the new option can be empty, it won't affect anything and we can land it before the ak-lite changes, right?

mike-sul:
> Since the new option can be empty, it won't affect anything and we can land it before the ak-lite changes, right?

Correct, and I verified it.

PR #1299, opened 2023-10-02
chore(fioconfig): Revert back when ommitting tag

StealthyCoder:
Just update the reference for `fioconfig` to make it so that when you remove overrides from `z-50-fioctl.toml` it will relinquish control back to whatever is in `/var/sota/sota.toml`.

@rsalveti can you verify all is in order?


ricardosalveti:
@StealthyCoder please just change the commit message to follow like done when bumping the fioconfig revision here (as in this case we're just changing the recipe to track a different revision, then we usually add the git history log as part of the commit message).

Example: https://github.com/foundriesio/meta-lmp/commit/00ced52b73eec5ffeb4241f778d24f1500e3d4a2

StealthyCoder:
@ricardosalveti I took your example and modified the commit message.

ricardosalveti (approved):
LGTM

PR #1301, opened 2023-10-03
Imx8mp/imx8mp switch to alternative boot upgrade procedure

igoropaniuk:
It turned out that both i.MX8MN and i.MX8MO don't have a possibility to control what A/B boot firmware set to boot, so we should always update A set.

This PR change from regular (which is broken now) to alternative boot upgrade procedure.


meta-lmp-bsp/recipes-bsp/u-boot/u-boot-ostree-scr-fit/imx8mp-lpddr4-evk/boot.cmd
```diff
@@ -16,14 +16,25 @@ setenv fit_addr 0x43800000
...
+
+setenv backup_primary_image 'echo "${fio_msg} backing up primary boot image set ..."; mmc dev ${devnum} 1 && mmc read ${loadaddr} 0x0 0x2000 && mmc dev ${devnum} 2 && mmc write ${loadaddr} 0x0 0x2000'
+setenv restore_primary_image 'echo "${fio_msg} restore primary boot image set ..."; mmc dev ${devnum} 2 && mmc read ${loadaddr} 0x0 0x2000 && mmc dev ${devnum} 1 && mmc write ${loadaddr} 0x0 0x2000'
+
+setenv update_primary_image1 'if test "${ostree_deploy_usr}" = "1"; then setenv image_path "${bootdir}/${bootloader_image}"; else setenv image_path "${ostree_root}/usr/lib/firmware/${bootloader_image}"; fi; setenv start_blk "${bootloader_s}";  run load_image; run update_image_boot0'
+setenv update_primary_image2 'if test "${ostree_deploy_usr}" = "1"; then setenv image_path "${bootdir}/${bootloader2_image}"; else setenv image_path "${ostree_root}/usr/lib/firmware/${bootloader2_image}"; fi; setenv start_blk "${bootloader2}";  run load_image; run update_image_boot0'
+
+setenv update_primary_image 'run update_primary_image1; run update_primary_image2'
+
+setenv check_secondary_boot 'setenv fiovb.is_secondary_boot "0"'
```
igoropaniuk, line 37: it's not possible to detect primary/secondary on imx8mp, so we just assume we're booting from primary always

angolini (approved):
I prefer if you could add on PR description the purpose of this PR. What does "alternative switch" means? Are you adding another alternative, or the first one? Why this PR is needed?

igoropaniuk:
@angolini added description, thanks!

ricardosalveti (approved):
LGTM

ricardosalveti:
Until we are able to detect that it booted from secondary we won't be able to test broken updates, right?

mwasilew:
@rsalveti IIUC the u-boot variable fiovb.is_secondary_boot will keep the information about the boot path. @igoropaniuk can confirm if I'm correct here.

igoropaniuk:
@igoropaniuk not really, `fiovb.is_secondary_boot` will be always `0`, as these SoCs don't have a mechanism to detect current boot path.

PR #1303, opened 2023-10-05
Improve common u-boot configs

MrCry0:
Test for this PR: https://github.com/foundriesio/lmp-manifest/pull/383

Build test **PASSED**


meta-lmp-base/recipes-bsp/u-boot/u-boot-fio-common.inc
```diff
@@ -12,12 +12,17 @@ UBOOT_REPO ??= "git://github.com/foundriesio/u-boot.git"
...
-    ${@bb.utils.contains('MACHINE_FEATURES', 'ebbr', 'file://lmp-ebbr.cfg', 'file://lmp-common.cfg', d)} \
+"
+
+# Handle lmp / lmp-mfgtool distro configs set
+SRC_URI:append = " \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'ebbr', 'file://lmp-common-ebbr.cfg', 'file://lmp-common.cfg', d)} \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'ebbr', 'file://lmp-ebbr.cfg', 'file://lmp.cfg', d)} \
     ${@bb.utils.contains('UBOOT_SIGN_ENABLE', '1', '', 'file://lmp-common-nosec.cfg', d)} \
-    ${@bb.utils.contains('MACHINE_FEATURES', 'ebbr', '', 'file://lmp.cfg', d)} \
 "
```
quaresmajose, line 22: I would rather have just one `ebbr`
```
# Handle lmp / lmp-mfgtool distro configs set
SRC_URI:append = " \
    ${@bb.utils.contains('MACHINE_FEATURES', 'ebbr',
        'file://lmp-common-ebbr.cfg', 'file://lmp-common.cfg', 'file://lmp-ebbr.cfg', 'file://lmp.cfg', d)} \
    ${@bb.utils.contains('UBOOT_SIGN_ENABLE', '1', '', 'file://lmp-common-nosec.cfg', d)} \
"
```

quaresmajose (approved):
LGTM

meta-lmp-base/recipes-bsp/u-boot/u-boot-fio-common.inc
angolini: What is the difference between `lmp-common-ebbr.cfg` and `lmp-ebbr.cfg`? Can we gather all ebbr configs in one cfg file only? (I'm asking because some times it's not possible)

angolini (approved):
I'm a little bit confused by the file naming. Shouldn't a "common" file be common for every configuration? Why do we need a "common-something" file?

Also, common and base together can be misleading. We use "base" because of "lmp-base".

Anyway, naming is always difficult, and sometimes it's impossible to avoid some decisions. This is only a set of suggestions.

meta-lmp-base/recipes-bsp/u-boot/u-boot-fio-common.inc
MrCry0: > What is the difference between `lmp-common-ebbr.cfg` and `lmp-ebbr.cfg`? Can we gather all ebbr configs in one cfg file only? (I'm asking because some times it's not possible)

@angolini 
lmp-common-ebbr.cfg - includes options common for all machines with enabled ebbr feature.
lmp-ebbr.cfg - an empty stub config file, like the same stub for lmp w/o ebbr feature `meta-lmp-base/recipes-bsp/u-boot/u-boot-fio/lmp.cfg`

We may need to rename lmp-common-ebbr.cfg to lmp-ebbr-common.cfg to move an accent. What do you think?

MrCry0:
> I'm a little bit confused by the file naming. Shouldn't a "common" file be common for every configuration? Why do we need a "common-something" file?
> 
> Also, common and base together can be misleading. We use "base" because of "lmp-base".
> 
> Anyway, naming is always difficult, and sometimes it's impossible to avoid some decisions. This is only a set of suggestions.


We can rename `lmp-common-ebbr.cfg` to `lmp-ebbr-common.cfg`, this naming looks more careful. What do you think?

meta-lmp-base/recipes-bsp/u-boot/u-boot-fio-common.inc
MrCry0: > I would rather have just one `ebbr`
> 
> ```
> # Handle lmp / lmp-mfgtool distro configs set
> SRC_URI:append = " \
>     ${@bb.utils.contains('MACHINE_FEATURES', 'ebbr',
>         'file://lmp-common-ebbr.cfg', 'file://lmp-common.cfg', 'file://lmp-ebbr.cfg', 'file://lmp.cfg', d)} \
>     ${@bb.utils.contains('UBOOT_SIGN_ENABLE', '1', '', 'file://lmp-common-nosec.cfg', d)} \
> "
> ```

Could you please point me to a manual with such syntax?
As I see, the current code doesn't support such multiply result groups :(
https://github.com/openembedded/bitbake/blob/master/lib/bb/utils.py#L1022

quaresmajose: > @quaresmajose
> Could you please point me to a manual with such syntax?
> As I see, the current code doesn't support such multiply result groups :(
> https://github.com/openembedded/bitbake/blob/master/lib/bb/utils.py#L1022

@MrCry0 I understand it wrong and don't see the commas impact so please don't consider my suggestion.

MrCry0:
> I'm a little bit confused by the file naming. Shouldn't a "common" file be common for every configuration? Why do we need a "common-something" file?
> 
> Also, common and base together can be misleading. We use "base" because of "lmp-base".
> 
> Anyway, naming is always difficult, and sometimes it's impossible to avoid some decisions. This is only a set of suggestions.

@angolini I renamed the files to more specific `lmp-base-common` and `lmp-ebbr-common`. It looks better, thanks for the suggestion!

ricardosalveti (approved):
LGTM, it would just be good to compare the final config before/after with every u-boot based board we support before merging, to make sure there is no unexpected side effect.

PR #1304, opened 2023-10-05
fix(aktualizr): Add User to Unit file

StealthyCoder:
In order for the `docker compose` commands to be executed correctly from within `aktualizr` the `$HOME` environment variable needs to be present. This can be done in two ways, either we add the HOME environment variable to the already existing `Environment=` settings or we set the `User`. 

All code in golang, so far as I have seemed to determine, uses the `$HOME` to determine the user home directory through methods like `UserHomeDir`. 

There are two other ways to get the data, one is through `os.user.Lookup`/`os.user.LookupId` and the other is `user.Current()`. Meaning we could also patch the `docker compose` code in order to use this particular method rather than setting a hardcoded user.


ricardosalveti (approved):
LGTM, thanks for the patch!

PR #1307, opened 2023-10-06
lmp-el2go-auto-register: enable generic usage for subscriber secure objects

arnout:
This PR makes various improvements to lmp-el2go-auto-register. The main goal is to make it more useful for a subscriber who wants to use EdgeLock 2GO to manage their own secure objects.

Currently only two secure objects are handled: Foundries and AWS IoT. The mechanism used for AWS IoT can however easily be applied more generically to any keypair and certificate that a subscriber wants to provision through EdgeLock 2GO. 

Subscriber keypairs may use an algorithm other than P-256, so support for that is added as well. However, this part is largely untested - in the end we used P-256 after all, so I didn't get around to test all the others, and our SE050 variant doesn't support P-384 and P-521.

We happen to use OID 0x5010 in our tests, and it turns out that the processing of labels didn't work correctly with smaller OIDs. The final patch fixes that bug.

The other patches are cleanups and enablers.

The documentation is updated accordingly: https://github.com/foundriesio/docs/pull/611

### DIFF
--- a/meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
+++ b/meta-lmp-base/recipes-support/lmp-el2go-auto-register/lmp-el2go-auto-register/lmp-el2go-auto-register
@@ -66,25 +66,27 @@ class FioCli:
         ]
         run_quietly(args)
 
+    @classmethod
+    def import_key(cls, slot: str, label: str):
+        oid = label.replace("SE_", "0x")
+        args = [
+            "fio-se05x-cli",
+            "--token-label",
+            "aktualizr",
+            "--import-key",
+            oid,
+            "--pin",
+            PIN,
+            "--id",
+            slot,
+            "--key-type",
+            "EC:prime256v1",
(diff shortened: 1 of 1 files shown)


doanac (approved):
this looks nice. We should probably have @mwasilew give it a quick test and then I think its good to merge.

mwasilew:
I'll try to test it next week.

arnout:
The documentation for this feature (https://github.com/foundriesio/docs/pull/611) was already merged...

ricardosalveti (approved):
LGTM, great set of changes, thanks for the contribution!

ricardosalveti:
@ldts one interesting one for you: https://github.com/foundriesio/meta-lmp/pull/1307/commits/382219a2cc1a6a28396e536a076ec348f5e5d8b9

ldts:
> @ldts one interesting one for you: [382219a](https://github.com/foundriesio/meta-lmp/commit/382219a2cc1a6a28396e536a076ec348f5e5d8b9)

yeah, I can do the conversion

PR #1314, opened 2023-10-18
base: systemd: use filter to get tpm2

quaresmajose:


MrCry0 (approved):
LGTM, I would just add a short commit message body in addition to the commit subject.

ricardosalveti (approved):
LGTM, same as pointed by @MrCry0, please add the reason why the change as part of the commit message (body).

ricardosalveti (approved):
LGTM

PR #1324, opened 2023-10-27
bsp: linux-lmp-fslc-imx: imx8mm-evk: fix compatible node on evka

ricardosalveti:
Upstream dts changes caused the evkb model to also be included in the evka dts file, causing a bug at userspace when trying to identify if hciattach should be executed or not (as it depends on evkb).


MrCry0 (approved):
LGTM

PR #1326, opened 2023-10-28
base: u-boot-fio-common: change config include order

ricardosalveti:
Make the machine specific configuration fragment (lmp-ebbr or lmp.cfg) be the last one in the fragments list so it can have higher priority.

This allows the machine specific config to replace any option that is set in another fragment that is parsed before it.


MrCry0 (approved):
LGTM

quaresmajose (approved):

PR #1333, opened 2023-10-31
base: lmp: add distro features to buildcfg vars

ricardosalveti:
Include distro features in pre-build configuration output in order to facilitate debugging from our CI logs.


meta-lmp-base/conf/distro/lmp.conf
```diff
@@ -5,6 +5,9 @@ DISTRO_NAME = "Linux-microPlatform"
 
 IMAGE_LINGUAS ?= "en-us"
 
+# Include distro features in pre-build configuration output
+BUILDCFG_VARS += "DISTRO_FEATURES"
```
quaresmajose, line 9: Why not in `meta-lmp-base/conf/distro/include/lmp.inc` and would be available in all distros

ricardosalveti: Yup, had that in my mind when changing but ended up changing at the wrong place :-)

ricardosalveti: Updated.

quaresmajose (approved):
LGTM

PR #1334, opened 2023-10-31
bsp: base-files: fstab: enable file system check on boot

ricardosalveti:
Enable file system check on boot mounts by default to avoid corrupted vfat partitions that can later block the boot process.


MrCry0 (approved):
.

MrCry0 (changes requested):
> Shouldn't we fix the `./meta-lmp-bsp/recipes-core/base-files/base-files/stm32mp15-disco/fstab` as well?

Nope :)

LGTM

MrCry0 (approved):

Tim-Anderson (commented):
LGTM

PR #1336, opened 2023-10-31
base: lmp: ota-esp is only needed when sota is enabled

quaresmajose:
Without this change lmp-base distro fails to build and other without sota DISTRO_FEATURES will still fail the same way.

### DIFF
--- a/meta-lmp-base/classes/lmp.bbclass
+++ b/meta-lmp-base/classes/lmp.bbclass
@@ -3,7 +3,9 @@ FIO_PUSH_CMD ?= "fiopush"
 FIO_CHECK_CMD ?= "fiocheck"
 SOTA_TUF_ROOT_DIR ?= "usr/lib/sota/tuf"
 
-IMAGE_FSTYPES += "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', ' ', d)}"
+IMAGE_FSTYPES_OTA ?= ""
+IMAGE_FSTYPES_OTA:sota = "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', '', d)}"
+IMAGE_FSTYPES += "${IMAGE_FSTYPES_OTA}"
 
 # Provided by meta-lmp-bsp or any other compatible BSP layer
 include conf/machine/include/lmp-machine-custom.inc


MrCry0 (approved):
LGTM

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -3,7 +3,7 @@ FIO_PUSH_CMD ?= "fiopush"
 FIO_CHECK_CMD ?= "fiocheck"
 SOTA_TUF_ROOT_DIR ?= "usr/lib/sota/tuf"
 
-IMAGE_FSTYPES += "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', ' ', d)}"
+IMAGE_FSTYPES:sota += "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', ' ', d)}"
```
ricardosalveti, line 6: Curious if :sota += will have bad implications here, did you check the final parsing?

Just wonder if we should instead use append here.

quaresmajose: Have see this on CI when testing, the += not work as you pointed. I have chnaged it to `append:sota`

ricardosalveti (approved):
LGTM

angolini:
What happens in DISTRO=lmp-base?

ricardosalveti:
> What happens in DISTRO=lmp-base?

I believe it still uses (or should use) esp with systemd-boot, but the current ota-esp is indeed sota specific because it depends on OTA_BOOT, which gets populated as the ostree-based image gets created.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -3,7 +3,7 @@ FIO_PUSH_CMD ?= "fiopush"
 FIO_CHECK_CMD ?= "fiocheck"
 SOTA_TUF_ROOT_DIR ?= "usr/lib/sota/tuf"
 
-IMAGE_FSTYPES += "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', ' ', d)}"
+IMAGE_FSTYPES:append:sota = "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', ' ota-esp', '', d)}"
```
ricardosalveti, line 6: There is only one problem here with append is that it will affect every image built by the build, including the initrds, so this won't really work without additional changes.

ricardosalveti: Unless you also add ota-esp to the remove list, e.g.:

`meta-lmp-base/recipes-core/images/initramfs-ostree-lmp-image.bb:IMAGE_FSTYPES:remove = "wic wic.gz wic.bmap wic.nopt ext4 ext4.gz"`

But for intel there is also core-image.

quaresmajose: yeah this fixes the `lmp-base` but I have see this circular dependency side effect so it breaks `lmp` maybe something like:
```
IMAGE_FSTYPES += " \
    ${@bb.utils.contains('DISTRO_FEATURES', 'sota', \
        ${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', '', d)} \
    '', d)} \
"
```

ricardosalveti: yeah, something along that line might work

quaresmajose: In the end I ended up adding a new variable IMAGE_FSTYPES_OTA_ESP for this propose. Teste with `lmp` and `lmp-base`.

quaresmajose:
> What happens in DISTRO=lmp-base?

This fix is mainly for lmp-base that is using do_image_ota when it shouldn't
```
INFO: recipe lmp-base-console-image-1.0-r0: task do_image_ota: Failed
```

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -3,7 +3,9 @@ FIO_PUSH_CMD ?= "fiopush"
 FIO_CHECK_CMD ?= "fiocheck"
 SOTA_TUF_ROOT_DIR ?= "usr/lib/sota/tuf"
 
-IMAGE_FSTYPES += "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', ' ', d)}"
+IMAGE_FSTYPES_OTA_ESP ?= ""
+IMAGE_FSTYPES_OTA_ESP:sota = "${@bb.utils.contains('EFI_PROVIDER', 'systemd-boot', 'ota-esp', '', d)}"
```
ricardosalveti, line 7: I would just make this variable a bit more generic, like LMP_OTA or simply OTA instead of OTA_ESP.

quaresmajose: `IMAGE_FSTYPES_OTA_ESP` renamed to `IMAGE_FSTYPES_OTA`.

ricardosalveti (approved):
LGTM

PR #1347, opened 2023-11-21
bsp: imx-boot: install all signed fw

MrCry0:
**NOTE, this change is backward compatible so we can apply these changes without bumping meta-freescale.**

The new version of imx-boot uses both signed_hdmi and signed_dp firmware for imx8m. Install them both.


quaresmajose (approved):

ricardosalveti (approved):
LGTM

PR #1348, opened 2023-11-21
bsp: u-boot-fio-mfgtool: imx8mp-lpddr4-evk: use imx-2022.04

MrCry0:
**WARNING! We still need to prove whether the flashing issue is a software issue, not a hardware one.**

Use earlier u-boot version imx-2022.04 to fix flashing I/O errors in some cases.


quaresmajose (approved):

ricardosalveti:
I'm fine having this as a temporary workaround, but it would still be good to try to understand better and isolate the changes.

MrCry0:
> I'm fine having this as a temporary workaround, but it would still be good to try to understand better and isolate the changes.

@ricardosalveti For now, the affected board even doesn't switch to the mmc boot, so we cannot rely on the hardware at this moment :(

PR #1349, opened 2023-11-23
Lmp disable gplv3 init install efi

angolini:
Use [1] to test.
I created one commit with the changes of the install script only to make it easier to review what the script was and how it is changed. The goal is to have this squashed before the merge.

Also, I created a new recipe and a new package name to make sure we can switch in the class file, however, there might be other better ways, I thought it is only simpler.

[1] https://docs.foundries.io/latest/reference-manual/linux/linux-wic-installer.html#testing-wic-image-installer-with-qemu-x86

### DIFF
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
@@ -138,17 +138,11 @@ if [ ! -e /etc/mtab ] && [ -e /proc/mounts ]; then
     ln -sf /proc/mounts /etc/mtab
 fi
 
-disk_info=$(parted -s ${device} unit mb print 2>&1 || true)
-if echo ${disk_info} | grep -q "unrecognised disk label"; then
-	echo "No valid disk label found, creating one"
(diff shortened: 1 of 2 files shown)


quaresmajose:
Why not only have one `initramfs-module-install-efi` version? maintain two different versions of the same thing will increase the maintenance and the probability to diverge. I prefer changing the current version to use fdisk/gdisk instead of parted if possible.

ricardosalveti:
Yeah, we can simplify this one by changing the default version instead.

ricardosalveti:
I will update the pr to include the changes against initramfs-module-install-efi (as we discussed before you went off).

ricardosalveti:
Updated but still not yet tested.

quaresmajose (approved):
LGTM
maybe we try to submit it upstream as the package is mainly this script.

quaresmajose:
Already send the first part to change the licenses.
https://lists.openembedded.org/g/openembedded-core/message/191350

We can also do the same here.

ricardosalveti:
A part of it is specific to our setup, but we can indeed upstream the change of tooling used.

ricardosalveti:
Updated and validated with qemu, working fine.

MrCry0 (approved):

PR #1355, opened 2023-12-04
mfgtools: disable hw wdog and move 8mp to use standard imx-boot

ricardosalveti:
As done for imx8mn, move mfgtools build to use standard imx-boot as that simplifies testing and the flashing logic (since we don't need to decouple u-boot.itb in a separated image).

Also disables hw watchdog as it was getting triggered while flashing imx8mp devices, and it is not needed for mfgtools.


MrCry0 (approved):
LGTM

quaresmajose (approved):

igoropaniuk:
IMO, the problem is that imx fastboot driver  implementation in U-Boot doesn't update watchdog timer.

So we still need to address that in u-boot source tree

MrCry0:
> IMO, the problem is that imx fastboot driver implementation in U-Boot doesn't update watchdog timer.
> 
> So we still need to address that in u-boot source tree

Yes, we can address this issue after the release.

PR #1359, opened 2023-12-08
base: rs: custom-sota-client: Add User to Unit file

mike-sul:
In order for the `docker compose` commands to be executed correctly from within the custom sota client the `$HOME` environment variable needs to be present.
This can be done in two ways, either we add the HOME environment variable to the already existing `Environment=` settings or we set the `User`.


mike-sul:
@StealthyCoder FYI

ricardosalveti (approved):
LGTM

PR #1366, opened 2023-12-12
recipes-support: make early-start only run to completion once

Tim-Anderson:
The way customers use this feature means that it can be several reboots before the device is registered. The early-start was running on every boot which causes delays in startup. This makes it so it only runs once and on all subsequent reboots dockerd will start the containers.


mike-sul (approved):

ricardosalveti (approved):
LGTM

PR #1380, opened 2024-01-02
base: distro: lmp: keep ext4 image

igoropaniuk:
Don't remove/keep ext4 image in deploy folder, as it might be needed by other scripts responsible for flashing a target.


ricardosalveti (approved):
LGTM

PR #1385, opened 2024-01-07
Fix the aklite image loading issue for v92 (base: rs: Bump aklite with fix for v92 838e53c)

mike-sul:
Relevant changes:
- 838e53c dockerclient: Calculate archive size correctly
- 6fda4d5 dockerclient: Handle tar creation errors


ricardosalveti (approved):
LGTM, thanks for fixing it!

PR #1386, opened 2024-01-08
base: add basic configs for cve-check

ricardosalveti:
Add basic cve-check configs and extra exclusions for lmp.


quaresmajose (approved):
LGTM

PR #1394, opened 2024-01-12
base: rs: aktualizr: Remove composectl from packageconfig

mike-sul:
The new utility is rather like `skopeo` not `aklite-apps` so no need to have the package config for it.


quaresmajose (approved):

ricardosalveti (approved):
LGTM

PR #1395, opened 2024-01-12
base: rc: composectl: Bump version 5419dab

mike-sul:
Relevant changes:
- 5419dab check: Use read-only store to copy from local store


ricardosalveti (approved):
LGTM

PR #1397, opened 2024-01-16
bsp: dynamic-layers: stm32-mfgtool-files: change sed separator

igoropaniuk:
Use ':' separator instead of '/' for sed 's' command, as for example LMP_MFGTOOL_FLASHLAYOUT value itself might also have slashes.

Fixes: 91eb22094("bsp: lmp-mfgtool: provide dynamic configuration for flashlayout class")


quaresmajose (approved):
LGTM

PR #1399, opened 2024-01-17
bsp: lmp-machine-custom: tegra: disable wic images if not used

quaresmajose:
The generation of the wic images on tegra is only supported when TEGRA_EXTERNAL_ROOTFS_WIC is enabled.


ricardosalveti (approved):
LGTM

PR #1400, opened 2024-01-17
Revert "bsp: stm32-mfgtool-files: keep track of the variables used"

quaresmajose:
This reverts commit 7600a79bd17504c0544a8bd5152740def380f884.

I have tested this localy and it works without the do_compile[vardeps] and so the do_compile task runs when the LMP_MFGTOOL_FLASHLAYOUT or LMP_FLASHLAYOUT_BOARD_NAME change.


ricardosalveti (approved):
LGTM

PR #1401, opened 2024-01-18
base: libnvidia-container: rdepend on tegra-libraries-cuda

quaresmajose:
Unblock meta-tegra bump and can be reverted when https://github.com/OE4T/meta-tegra/pull/1456 was merged

### DIFF
--- a/meta-lmp-bsp/dynamic-layers/tegra/recipes-containers/libnvidia-container/libnvidia-container_1.10.0.bbappend
+++ b/meta-lmp-bsp/dynamic-layers/tegra/recipes-containers/libnvidia-container/libnvidia-container_1.10.0.bbappend
@@ -1,0 +1,1 @@
+RDEPENDS:${PN}:append = " tegra-libraries-cuda"


meta-lmp-base/recipes-containers/libnvidia-container/libnvidia-container_1.10.0.bbappend
```diff
@@ -0,0 +1 @@
+RDEPENDS:${PN}:append:tegra = " tegra-libraries-cuda"
```
ricardosalveti, line 1: Have it under dynamic layers, otherwise bitbake will complain when meta-tegra is not there.

Might also not need the 'tegra' override here since it is only relevant for tegra anyway.

quaresmajose: right! fixed

ricardosalveti (approved):
LGTM

PR #1402, opened 2024-01-18
base: lmp-signing: fix append when vardeps is none

quaresmajose:
The task flag vardeps can be empty and in this case it is none.
Consequently this means that we cannot add anything to the task vardeps when it is empty.
We should also match any full string present on vardeps and not sub-strings so varname_hash should be checked on vardeps list and not string.


ricardosalveti (approved):
LGTM

PR #1403, opened 2024-01-19
base: lmp-device-auto-register: add support for HSM

mwasilew:
This patch adds support for device auto registration with use of HSM. Auto registration script will source the following variables from /etc/sota/hsm:
 * HSM_MODULE
 * HSM_PIN
 * HSM_SOPIN Variables will be used as values for the corresponding parameters of lmp-device-register.


quaresmajose (approved):

ricardosalveti (approved):
LGTM

ricardosalveti:
@mwasilew would be nice to have this in our docs as well.

mwasilew:
OK, I'll add a PR for the docs

PR #1408, opened 2024-01-25
base: fioconfig: Bump version

StealthyCoder:
Pulls in fix that uses correct reference for SOTA_DIR.


ricardosalveti (approved):
LGTM, don't forget to update the release notes with the known issue for this one, pointing to this commit (for customers wanting to fast track the fix).

PR #1417, opened 2024-02-05
bsp: linux-lmp-rpi: 5.15.92 -> 6.1.74

quaresmajose:


igoropaniuk (approved):
LGTM

ricardosalveti:
I'm not sure this will work correctly in kirkstone, as the kernel usually needs to be updated together with the firmware blobs, and we're still using 5.15 basically because that is the supported version in meta-rpi kirkstone.

MrCry0 (approved):

quaresmajose:
I have not test this and have it mainly to fix main-next that tracks the layer meta-rpi master. I will drop and keep this only on main-next.

PR #1418, opened 2024-02-05
Cleanup the home directory as it is not preserved by ostree

quaresmajose:


MrCry0 (approved):

meta-lmp-base/recipes-samples/images/lmp-image-common.inc
```diff
@@ -104,6 +104,6 @@ python () {
 
 EXTRA_USERS_PARAMS = "\
 groupadd ${LMP_USER}; \
-useradd -p '${LMP_PASSWORD}' ${LMP_USER}; \
```
ricardosalveti, line 107: Can you check if it gets created when trying to log-in via ssh first (before any serial console login)?

AFAIK it wasn't really able to create automatically when calling via ssh, same issue we have after running recovery.

quaresmajose: I have tested it only the the serial with lmp and lmp-base. I need to check the ssh too

quaresmajose: Tested with the lmp-base-console-image and it also works over ssh.

quaresmajose:
> @quaresmajose Could you also include this related commit to your PR? [3699011](https://github.com/foundriesio/meta-lmp/commit/3699011969d7ba28f3c49159fa1ea6b772283ff1)

We don't need to remove the /home directory content anymore because with this PR they should be clean at the end of the image do_rootfs task. 

In the near future I would like to make this not clean /home and /var a failure, changing the meta-updater [OSTREE_RMDIR_HELPER_MSGTYPE](https://github.com/uptane/meta-updater/blob/0bd2045d6efe771c6516cb06849c346e8b052210/classes/image_types_ostree.bbclass#L25C1-L25C28) to bberror in the helper ostree_rmdir_helper.

ricardosalveti (approved):
LGTM

PR #1420, opened 2024-02-09
meta-lmp-base: allow lmp-device-auto-register to use hostname

mwasilew:
This patch enables lmp-device-auto-register to use /etc/hostname as FoundriesFactory device registration name. It is done by setting build time variable LMP_AUTO_REGISTER_USE_HOSTNAME. If this variable is empty lmp-device-auto-register will use randomly generated UUID.


quaresmajose (approved):

MrCry0 (approved):

ricardosalveti (approved):
LGTM

PR #1422, opened 2024-02-20
Add initial support for splitting var as a separated partition

ricardosalveti:


meta-lmp-base/classes/lmp.bbclass
```diff
@@ -125,10 +125,18 @@ IMAGE_CMD:ota:append () {
...
+	if [ "${OSTREE_SPLIT_VAR}" = "1" ]; then
+		rm -rf ${OTA_VAR}
+		mv ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var ${OTA_VAR}
+		mkdir -p ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var
+	fi
 }
 OTA_BOOT = "${WORKDIR}/ota-boot"
-do_image_ota[dirs] += "${OTA_BOOT}"
-do_image_ota[cleandirs] += "${OTA_BOOT}"
+OTA_VAR = "${WORKDIR}/ota-var"
```
MrCry0, line 137: Wouldn't it be more logical and readable to place setting vars above using them?

MrCry0 (approved):
LGTM with a minor note.

meta-lmp-base/classes/lmp.bbclass
```diff
@@ -125,10 +125,18 @@ IMAGE_CMD:ota:append () {
 	if [ "${APP_PRELOAD_WITHIN_OE_BUILD}" = "1" ]; then
 		preload_apps
 	fi
+
+	# Split content /var into a separated folder so it can be consumed by WKS separately
+	if [ "${OSTREE_SPLIT_VAR}" = "1" ]; then
+		rm -rf ${OTA_VAR}
+		mv ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var ${OTA_VAR}
+		mkdir -p ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var
```
quaresmajose, line 133: Using `mv` will fail if we run this task more than one time because on the second one the `${OSTREE_OSNAME}/var` is clean. It should be better to copy all the files with `cp -a` and we can drop the `mkdir`.

MrCry0: We don't need to have the content in `${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var` after this code.

But I agree we may need to move the content only once like
```
if [ "${OSTREE_SPLIT_VAR}" = "1" ] && ! -d ${OTA_VAR}; then
		mv ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var ${OTA_VAR}
		mkdir -p ${OTA_SYSROOT}/ostree/deploy/${OSTREE_OSNAME}/var
fi
```

meta-lmp-base/classes/lmp.bbclass
ricardosalveti: Funny enough most tasks end up defining the variables and cleaning actions after the task itself is defined (if you see the other tasks in this class). We can probably have a separated commit to move things around if needed (here I would prefer to just add the new variable to avoid a larger diff).

meta-lmp-base/classes/lmp.bbclass
ricardosalveti: OTA_SYSROOT is already defined for dir and cleandirs:
```
do_image_ota[dirs] = "${OTA_SYSROOT}"
do_image_ota[cleandirs] = "${OTA_SYSROOT}"
```

So a re-run wouldn't be an issue here.

mkdir might not necessarily be needed indeed, will check.

ricardosalveti:
@Tim-Anderson mind reviewing the resize-helper related changes?

Tim-Anderson (approved):
this removes any validation of partition and simply expands the last partition of the device. If that is what you want with no validation that it is the right partition then that should work.

ricardosalveti:
> this removes any validation of partition and simply expands the last partition of the device. If that is what you want with no validation that it is the right partition then that should work.

Yeah, the idea is to expand the latest, which in our standard setup ends up being basically the rootfs. Once we add another patition (e.g. var), it expands that one instead.

PR #1425, opened 2024-02-20
bsp: dynamic-layers: u-boot-ti-staging: drop bootcount

igoropaniuk:
Drop legacy bootcount support as it's handled now in boot.cmd script.


MrCry0 (approved):

ricardosalveti:
Don't we also need to do a similar change for am64xx?

ricardosalveti:
> @ricardosalveti I haven't added support for boot firmware updates yet for am64xx

right, it is indeed not using boot common yet.

ricardosalveti (approved):
LGTM

PR #1426, opened 2024-02-21
base: layer.conf: exclude os-release siggen dependencie on the initramfs ostree-lmp-recovery

quaresmajose:
This patch is a follow-up of 24d1016e, the initramfs-ostree-lmp-recovery is used on the machine qemuarm64-secureboot and without this patch the initramfs is rebuilded which consequently triggers a kernel rebuild.


ricardosalveti (approved):
LGTM

PR #1430, opened 2024-02-29
bsp: tf-a-fio-st: fix fip image deploy path

igoropaniuk:
Fix image deploy path which is used STM32_ROT_SIGN_ENABLE is enabled.


ricardosalveti (approved):
LGTM

PR #1438, opened 2024-04-06
bsp: freescale: cairo: fix compatibility with 1.16

MrCry0:
The commit [1] breaks compatibility with cairo < 1.18. LmP is still based on OE kirkstone, that uses cairo 1.16. Return back compatibility.

[1]
commit 0e478bde ("cairo: Drop the removed packageconfigs")


quaresmajose:
Have this bbappend version specific to only be valid on version 1.16 is better in my opinion, so we don't forget to drop it when we switch to the 1.18

MrCry0:
> Have this bbappend version specific to only be valid on version 1.16 is better in my opinion, so we don't forget to drop it when we switch to the 1.18

@quaresmajose sure, we'll drop it during switching to scarthgap.

MrCry0:
> Have this bbappend version specific to only be valid on version 1.16 is better in my opinion, so we don't forget to drop it when we switch to the 1.18

@quaresmajose you're completely right, we have no other custom settings for cairo for now.

quaresmajose (approved):

PR #1442, opened 2024-04-12
base: wpanusb: drop recipe

quaresmajose:
- we don't use it anymore
- also drop it from lmp-feature-nat64 images


ricardosalveti (approved):
LGTM

PR #1443, opened 2024-04-14
linux-lmp: refix setting a subdir for dts files

MrCry0:
We need to support kernel versions with and without vendor subdirs
for arm architecture dts files. Replace a static prefix with a
dynamic one set according to a kernel version.

### DIFF
--- a/meta-lmp-base/classes/lmp-staging.bbclass
+++ b/meta-lmp-base/classes/lmp-staging.bbclass
@@ -9,8 +9,15 @@ LMPSTAGING_INHERIT_KERNEL_MODSIGN = ""
 
 LMPSTAGING_LOCK_TO_AVOID_OOM = "clang-native rust-native rust-llvm-native"
 
+LMPSTAGING_KERN_ADD_ST_SUBDIR = ""
+
 python __anonymous() {
     pn = d.getVar('PN')
+
+    if pn == "linux-lmp" or pn == "linux-lmp-rt":
+        (major_ver, minor_ver, rest) = d.getVar('LINUX_VERSION').split(".")
+        if int(major_ver) > 6 or (int(major_ver) == 6 and int(minor_ver) >= 4):
+            d.setVar('LMPSTAGING_KERN_ADD_ST_SUBDIR',  'st/')
 
     if bb.data.inherits_class('module', d):
         d.appendVar('DEPENDS', ' virtual/kernel')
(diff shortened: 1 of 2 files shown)


meta-lmp-base/classes/lmp-staging.bbclass
```diff
@@ -12,6 +12,13 @@ LMPSTAGING_LOCK_TO_AVOID_OOM = "clang-native rust-native rust-llvm-native"
 python __anonymous() {
     pn = d.getVar('PN')
 
+    d.setVar('LMPSTAGING_KERN_ADD_ST_SUBDIR',  '')
```
quaresmajose, line 15: move this variable outside the function with the default version make the things more clear in my opinion

MrCry0: Do you mean to move outside the function but leave it in the stage class?

quaresmajose: yes, I mean together with the LMPSTAGING_INHERIT_KERNEL_MODSIGN and LMPSTAGING_LOCK_TO_AVOID_OOM.

ricardosalveti: I think this was a generic change done by the kernel for all arm32 targets, right? If so we can have the variable defined as arm32 instead of ST.

MrCry0: > I think this was a generic change done by the kernel for all arm32 targets, right? If so we can have the variable defined as arm32 instead of ST.

For now, we need to handle only an ST vendor case, as NXP boards are handled in meta-freescale. As a prefix is vendor-specific, I'd keep it as is. Anytime we can rename it to a generic one if we have to expand a solution for other vendors.

MrCry0: > yes, I mean together with the LMPSTAGING_INHERIT_KERNEL_MODSIGN and LMPSTAGING_LOCK_TO_AVOID_OOM.

Done, thanks!

quaresmajose (approved):
LGTM

meta-lmp-base/classes/lmp-staging.bbclass
ricardosalveti: > > I think this was a generic change done by the kernel for all arm32 targets, right? If so we can have the variable defined as arm32 instead of ST.
> 
> For now, we need to handle only an ST vendor case, as NXP boards are handled in meta-freescale. As a prefix is vendor-specific, I'd keep it as is. Anytime we can rename it to a generic one if we have to expand a solution for other vendors.

True, we need to know the prefix.

ricardosalveti (approved):
LGTM

PR #1445, opened 2024-04-16
base: weston-init: Add Upstream-Status

quaresmajose:
It's a patch to configure the behavior, so it's not supposed to be upstreamed.


igoropaniuk (approved):
LGTM

ricardosalveti (approved):
LGTM

PR #1453, opened 2024-05-07
linux-lmp: rework a versioning of patches

MrCry0:
It simplifies the versioning and prepares lmp to support 2 kernel versions of linux kernel versions with a lot of patches like linux-lmp-fslc-imx.

I removed switching linux-lmp to 6.6 as it needs to bump meta-freescale* as well, we can switch to 6.6 once all changes for all linux-lmp* are in place.

### DIFF
--- a/meta-lmp-bsp/recipes-kernel/linux/linux-lmp_%.bbappend
+++ b/meta-lmp-bsp/recipes-kernel/linux/linux-lmp_%.bbappend
@@ -1,0 +1,5 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}/${KSHORT_VER}:${THISDIR}/${PN}:"
+
+SRC_URI:append:stm32mp15-disco = " \
+    ${@bb.utils.contains('MACHINE_FEATURES', 'se05x', "file://0001-FIO-internal-arch-arm-dts-stm32mp157c-dk2-enable-I2C.patch", '', d)} \
+"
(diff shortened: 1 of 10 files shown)


meta-lmp-bsp/recipes-kernel/linux/linux-lmp_%.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}/${KSHORT_VER}:${THISDIR}/${PN}:"
```
angolini, line 1: What happens if you use PV here instead?

MrCry0: @angolini the standard PV for kernel includes SRCPV which is changed each time we bump a particular kernel. Usually patches can be applied without changes to any kernel with the same major/minor version, like 6.6. So this change allows us to use patchsets tuned for different major/minor kernel with the same patch names.

MrCry0: Yes, we have a lot of patches for linux-lmp-fslc-imx 6.1 and 6.6, and saving the short version of PV allows us to store different versions of the same patches for different kernel versions in the fine structure like
/linux-lmp/6.1/
/linux-lmp/6.6/
or
/linux-lmp-fslc-imx/6.1/
/linux-lmp-fslc-imx/6.6/

angolini (approved):

angolini:
I'm closing this PR as we already migrated to kernel 6.6 and there are conflicts. But if it's needed in the future, please reopen.

PR #1454, opened 2024-05-10
Bump aklite and composectl

mike-sul:
- Add callback support to the offline updater.
- Add bundle metadata support to the offline updater. The bundle metadata are signed by TUF targets role key and contains description of the bundle, i.e. tag, bundle targets, whether it is prod or CI bundle, etc. The offline updater checks the bundle metadata signature and use information found in it to determine whether an update should be permitted, possible, and what targets can be installed.
- Move app functionality to check whether app is fetched, installed, and running from aklite to composectl.

### DIFF
--- a/meta-lmp-base/recipes-containers/composeapp/composectl_git.bb
+++ b/meta-lmp-base/recipes-containers/composeapp/composectl_git.bb
@@ -1,13 +1,13 @@
 DESCRIPTION = "A CLI utility to manage compose apps"
 
 SECTION = "devel"
-LICENSE = "Apache-2.0"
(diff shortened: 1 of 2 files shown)


doanac:
@caiotpereira ^ FYI - this will wind up in release notes

mike-sul:
This is a doc on the bundle metadata signing https://ci.foundries.io/projects/fio-docs/builds/2648/docs/artifacts/html/user-guide/offline-update/offline-update.html#signing-the-offline-update-bundle.

detsch (approved):
LGTM. Did some testing with qemu, and all worked as expected.

PR #1464, opened 2024-05-24
base: systemd: switch to cgroupv2 by default

ricardosalveti:
There is no reason anymore to keep using cgroupv1 (every component using cgroups is already v2 compatible), so switch to v2 by default, as done on every major linux distribution.

https://github.com/opencontainers/runc/blob/main/docs/cgroup-v2.md


mike-scott:
Did you mean "There is no reason anymore to keep using cgroupv1" in the description?

MrCry0 (approved):
With Mike's comment +1, LGTM

ricardosalveti:
> Did you mean "There is no reason anymore to keep using cgroupv1" in the description?

Yup, updated.

PR #1475, opened 2024-06-10
bsp: k3: conditionally copy boot artifacts to boot

igoropaniuk:
Set OSTREE_DEPLOY_USR_OSTREE_BOOT for K3 machines only when LUKS is enabled.


angolini (approved):

ricardosalveti:
So this was created as a way to facilitate boot firmware updates as we didn't support updating those files automatically as part of the OTA process (easy to locate the files to update to).

But we can make it conditional to luks as well, not an issue.

ricardosalveti (approved):

PR #1479, opened 2024-06-27
bsp: meson: fixup the imx meson backported

quaresmajose:
Using the DEFAULT_PREFERENCE on the meson_0.63.3.imx.bb does not work in this case because
we cannot use the DEFAULT_PREFERENCE to choose versions in lower priority layers,
which are basically almost all of them because we have a very high priority value on the
BBFILE_PRIORITY.
So we need to also set the default PREFERRED_VERSION_meson with the version available
on oe-core because we can't use the DEFAULT_PREFERENCE.
Another issue fixed was the overrides used imx8-nxp-bsp and imx9-nxp-bsp that are not defined
on the IMX bsp layer, we can use the imx-generic-bsp which is common to all IMX.


ricardosalveti (approved):
LGTM

PR #1483, opened 2024-07-09
se05x: split agent into a separated package

ricardosalveti:


quaresmajose (approved):

angolini (approved):
LGTM

I understand we previously did not install the agent, that's why we go from 2 lib to 2 lib plus a directory in the package files definition

ricardosalveti:
> I understand we previously did not install the agent, that's why we go from 2 lib to 2 lib plus a directory in the package files definition

The agent itself was installed as part of the package, and it was available in bindir, which is why it wasn't explicitly described in FILES.

PR #1491, opened 2024-07-16
bsp: stm32mp15-disco: fix size of u-boot-env partition

igoropaniuk:
Fix the size of u-boot-env partition, which should be 0x4000 as defined in CONFIG_ENV_SIZE, otherwise U-Boot uses the last 0x4000 bytes to place the environment [1].

[1] https://github.com/u-boot/u-boot/blob/v2024.04/env/mmc.c#L124


ricardosalveti (approved):
LGTM

PR #1493, opened 2024-07-17
base: cryptfs: force filesystem check before resize

Tim-Anderson:
Make sure the FS is clean before conversion. Force the fsck

### DIFF
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/cryptfs
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/cryptfs
@@ -13,7 +13,7 @@ cryptfs_enabled() {
 
 e2fsck_check() {
 	if [ -n "`which e2fsck`" ]; then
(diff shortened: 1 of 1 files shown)


MrCry0 (approved):
LGTM

meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/cryptfs
```diff
@@ -13,7 +13,7 @@ cryptfs_enabled() {
 
 e2fsck_check() {
 	if [ -n "`which e2fsck`" ]; then
-		fsckout=`e2fsck -p -v ${1}`
+		fsckout=`e2fsck -p -v -f ${1}`
```
ricardosalveti, line 16: Here it will cause every boot to run with a forced fsck, while we only really need before resizing.

Can you change to accept a parameter and only call with -f on the block that performs the resizing?

ricardosalveti (approved):
LGTM

PR #1495, opened 2024-07-26
Capsules intall fix

igoropaniuk:


meta-lmp-base/recipes-sota/fioefi/fioefi_0.1.bb
```diff
@@ -3,6 +3,7 @@ LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
 SRC_URI = "file://fioefi.sh.in"
+RDEPENDS:${PN}:append:tegra = " lmp-boot-firmware"
```
quaresmajose, line 6: for all target we don't need the `tegra` machine override because this way it will only work in that arch

igoropaniuk: oops, that's a typo after copy-pasting, thanks!

quaresmajose (approved):
LGTM

PR #1502, opened 2024-08-14
linux-lmp-fslc-imx-xeno4: move patches specific to the xeno4 recipe

ricardosalveti:
Isolate patches under the linux-lmp-fslc-imx-xeno4 folder in order to make the patches specific to the base kernel version used by it.


angolini (approved):
good catch!

PR #1507, opened 2024-08-26
[RFC] base: systemd-boot: support rollback event

igoropaniuk:
Check BootChainFwCurrent-781e084c-a330-417c-b678-38e696380cb9 EFI variable, which provides a hint if we need to rollback and boot use fallback ostree sysroot deployment.

### DIFF
--- a/meta-lmp-base/recipes-core/systemd/systemd-boot_%.bbappend
+++ b/meta-lmp-base/recipes-core/systemd/systemd-boot_%.bbappend
@@ -1,5 +1,11 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
 # Ostree handles the default boot configuration
 RDEPENDS:${PN}:remove:sota = "virtual-systemd-bootconf"
+
+SRC_URI:append = " \
+    file://0001-boot-add-support-for-ostree-rollback.patch \
+"
 
 # Install systemd-boot at expected path for tools such as bootctl
 do_install:append() {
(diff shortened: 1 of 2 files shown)


igoropaniuk:
I'll recreate this PR on top of scarthgap branch after initial feedback received
Also I simplified a bit fallback implementation, so it's now just provides a hint what entry should be booted (like it's done with `LoaderEntryOneShot`)

@ricardosalveti

ricardosalveti:
> Technically we can just re-use this code for any target, considering that some Dxe driver provides us information about FW slot booted via EFI variable

Right, this EFI variable UID is specific to Nvidia, so we would need something vendor specific for other implementations as well.

Is BootChainFwCurrent used to indicate that the firmware itself performed a rollback after a capsule update? I guess we can relate both because in our case (with ostree) we only update the firmware when the OS is also updated (as the capsule will be part of the image itself).

meta-lmp-base/recipes-core/systemd/systemd-boot_%.bbappend
```diff
@@ -1,6 +1,12 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
 # Ostree handles the default boot configuration
 RDEPENDS:${PN}:remove:sota = "virtual-systemd-bootconf"
 
+SRC_URI:append = " \
```
quaresmajose, line 6: just a minor suggestion to keep it only for sota `SRC_URI:append:sota = " \`

angolini (approved):

igoropaniuk:
Closing this PR in favor of a different implementation based on ESRT table

PR #1508, opened 2024-08-27
base: UEFI Secure Boot Provisioning

ldts:
Simplify the Secure Boot key provisioning process by adding a systemd-boot entry wich uses the [efitools](https://github.com/mjg59/efitools.git) EFI program [LockDown.efi](https://github.com/mjg59/efitools/blob/master/LockDown.c).

image

On selecting that entry, the keys are provisioned after which the system  resets.

Users can then boot into Linux with a secure boot enabled chain.

### DIFF
--- a/meta-lmp-base/recipes-bsp/efitools/efitools/lockdown.conf
+++ b/meta-lmp-base/recipes-bsp/efitools/efitools/lockdown.conf
@@ -1,0 +1,2 @@
+title UEFI Secure Boot Provisioning
+efi /LockDown.efi
--- a/meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
+++ b/meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
@@ -13,6 +13,7 @@ SRC_URI += " \
     file://0001-Enable-RISC-V-build.patch \
     file://build-keys-for-lockdown-only.patch \
     file://allow-local-auths.patch \
+    file://lockdown.conf \
 "
 
 COMPATIBLE_HOST = "(i.86|x86_64|arm|aarch64|riscv64).*-linux"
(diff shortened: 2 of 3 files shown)


MrCry0 (approved):

ldts:
Using the default keys provided with lmp-tools on tag 94 will fail with output:

image

ldts:
This PR can be merged at any time - no dependencies. It just needs valid keys.

meta-lmp-base/recipes-bsp/efitools/efitools/0001-lockdown-systemd-boot-conf.patch
```diff
@@ -0,0 +1,23 @@
...
+new file mode 100755
+index 0000000..4d7a943
+--- /dev/null
++++ b/lockdown.conf
+@@ -0,0 +1,2 @@
++title UEFI Secure Boot Provisioning
++efi /LockDown.efi
+\ No newline at end of file
+-- 
+2.34.1
```
ricardosalveti, line 22: You don't need a patch for this, you can add the file directly at SRC_URI.

ldts: sure, I thought the patch will be better, anyway, let me update

ldts:
@rsalveti anything else you want to add before I merge

ricardosalveti:
Nops, but don't merge just yet, we're in the process of releasing v94.1, holding on merges until we're good with it.

ldts:
> Nops, but don't merge just yet, we're in the process of releasing v94.1, holding on merges until we're good with it.

just ping me whenever

PR #1510, opened 2024-08-29
bsp: core: base-files: automount vfat boot partition

vanmaegima:
Automount[1] the boot partition to minimize vfat partition corruption risks.

This mounts the partition on demand and unmounts after use.

[1] https://www.freedesktop.org/software/systemd/man/latest/systemd.automount.html#


ricardosalveti (approved):
LGTM

PR #1513, opened 2024-09-11
[RFC] base: ostree: add support systemd-boot automatic boot assesment

igoropaniuk:
Add support for Automatic Boot Assessment [1].
Boot entries are now created with an additional suffix, which represents the amount of maximum tries for boot counting.

[1] https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/


meta-lmp-base/recipes-extended/ostree/ostree/0006-Add-support-systemd-boot-automatic-boot-assesment.patch
```diff
@@ -0,0 +1,148 @@
+From a9e35b76fd5e47b9bc29a7113d97a4d98a973e75 Mon Sep 17 00:00:00 2001
+From: Igor Opaniuk <igor.opaniuk@foundries.io>
+Date: Wed, 11 Sep 2024 18:03:10 +0200
+Subject: [PATCH] Add support systemd-boot automatic boot assesment
```
ricardosalveti, line 4: s/assesment/assessment/

ricardosalveti:
Change itself looks ok, is bless boot enabled by default? Just trying to understand if once we land this patch everything will keep working the same way, until we move the confirmation to aktualizr-lite or similar.

igoropaniuk:
> is bless boot enabled by default

yes, I've tested in on `intel-corei7-64`, systemd service configurations (`systemd-bless-boot.service`, `boot-complete.target`) and tools are installed to rootfs by default (`systemd-bless-boot`)

Systemd-boot EFI app handles all `+n` suffixes in boot entries automatically (no need to enable anything explicitly )

meta-lmp-base/recipes-extended/ostree/ostree/0006-Add-support-systemd-boot-automatic-boot-assesment.patch
```diff
@@ -0,0 +1,148 @@
...
+
+Add support for Automatic Boot Assessment [1].
+Boot entries are now created with an additional suffix, which
+represents the amount of maximum tries for boot counting.
+
+[1] https://systemd.io/AUTOMATIC_BOOT_ASSESSMENT/
+Upstream-Status: Inappropriate [lmp specific]
```
ricardosalveti, line 11: We should probably set as pending here, as this is something we want to try to upstream once there is support for systemd.

quic-roshs:
Shouldn't we be checking if the bootloader used is systemd-boot before appending the counter to the conf file? 
This might be needed if we are planning to upstream the patch.

igoropaniuk:
@quic-roshs this behavior is described in Bootloader Specification (https://uapi-group.org/specifications/specs/boot_loader_specification/#boot-counting; as well as OSTree also follows it https://ostreedev.github.io/ostree/deployment/#the-system-boot), so this is not systemd-boot specific adjustments

ricardosalveti:
> @quic-roshs this behavior is described in Bootloader Specification (https://uapi-group.org/specifications/specs/boot_loader_specification/#boot-counting; as well as OSTree also follows it https://ostreedev.github.io/ostree/deployment/#the-system-boot), so this is not systemd-boot specific adjustments

Is u-boot (uEnv) based boot flows reusing the same code path?

igoropaniuk:
@ricardosalveti 

> Is u-boot (uEnv) based boot flows reusing the same code path?

yes, to create the final uEnv.txt it parses boot entries that was created before by the function I adjusted in this PR with `create_config_from_boot_loader_entries()` function (https://github.com/ostreedev/ostree/blob/main/src/libostree/ostree-bootloader-uboot.c#L102) .
The change of boot entries naming doesn't somehow affect the contents of the final `uEnv.txt`

ricardosalveti:
Needs a rebase.

quaresmajose:
merged manually

PR #1514, opened 2024-09-23
base: lmp: efi: check for sbsign sbverify errors.

ldts:
Report errors during secure boot sign/verify.


quaresmajose (commented):
just a minor suggestion for the commit, as we are changing a bbclass it's better to use her name:
```
base: kernel-lmp-efi: check for sbsign sbverify errors
```

PR #1520, opened 2024-10-15
U-boot 2024.04 and rpi4/5 support

ricardosalveti:
Also fixed scr-fit for the qemu targets in order to boot test with the new u-boot.


meta-lmp-base/recipes-bsp/u-boot/u-boot-fio-mfgtool_2024.04.bb
```diff
@@ -2,7 +2,7 @@ SUMMARY = "Produces a Manufacturing Tool compatible U-Boot"
 DESCRIPTION = "U-Boot recipe that produces a Manufacturing Tool compatible \
 binary to be used in updater environment"
 
-require recipes-bsp/u-boot/u-boot-fio_2023.04.bb
+require recipes-bsp/u-boot/u-boot-fio_${PV}.bb
```
quaresmajose, line 5: I believe we can drop all the u-boot-fio-mfgtool and switch to change the default u-boot-fio config using distro overrides like we did in https://github.com/foundriesio/meta-lmp/pull/1519

quaresmajose (approved):
LGTM

PR #1522, opened 2024-10-25
base: init-install-efi: installer: support encrypted rootfs

ldts:
Add support for a CI encrypted rootfs USB installer.

### DIFF
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
@@ -9,6 +9,20 @@
 
 PATH=/sbin:/bin:/usr/sbin:/usr/bin
 
+check_secure_boot() {
+    if [ ! -d /sys/firmware/efi/efivars ]; then
+        echo "EFI vars sysfs mount point not found"
+	exit 1
+    fi
+
+    efi_secure=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal`
+    efi_mode=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SetupMode --print-decimal`
+    if [ "${efi_secure}" != "1" ] || [ "${efi_mode}" != "0" ]; then
+	echo "UEFI SecureBoot not enabled, installation aborted"
+	exit 1
+    fi
+}
+
 # Recommended ESP partition size is 512m
 boot_size=512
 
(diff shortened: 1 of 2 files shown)


ldts:
getting rid of "fiopassphrase" for the installer and the initramdisk should be done in a different commit.
having it first here allows users already using "fiopassphrase" to test the installer and TPM2 luks reencryption onf the next boot.

ldts:
Enforce that the PK has been set before installing.

meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
```diff
@@ -9,6 +9,20 @@
 
 PATH=/sbin:/bin:/usr/sbin:/usr/bin
 
+check_secure_boot() {
+    if [ ! -d /sys/firmware/efi/efivars ]; then
+        echo "EFI vars sysfs mount point not found"
+	exit 1
+    fi
+
+    efi_secure=`hexdump /sys/firmware/efi/efivars/SecureBoot-* | head -n1 | awk '{print $4}'`
```
quaresmajose, line 18: the efivar tool should be more bullet proof instead of parsing binary strings.
It has no dependencies, so its installation size is super small.

```
# Whether the platform firmware is operating in Secure boot mode (1) or not (0)
efi_secure=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal`
# Whether the system should require authentication on SetVariable() requests to Secure Boot policy variables (0) or not (1)
efi_mode=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SetupMode --print-decimal`
```

The `8be4df61-93ca-11d2-aa0d-00e098032b8c` is a global variable and part of the spec
```
efivar --list-guids|grep 8be4df61-93ca-11d2-aa0d-00e098032b8c
{8be4df61-93ca-11d2-aa0d-00e098032b8c}  {global}        efi_guid_global EFI Global Variable
```

meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi_%.bbappend
```diff
@@ -2,4 +2,4 @@ FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
 
 # Prefer gptfdisk instead of parted
 RDEPENDS:${PN}:remove = "parted dosfstools"
-RDEPENDS:${PN} += "efibootmgr gptfdisk"
+RDEPENDS:${PN} += "efibootmgr gptfdisk systemd-crypt"
```
quaresmajose, line 5: `RDEPENDS:${PN} += "efivar"` to use the efivar tool on the ramdisk

ldts: absolutely! makes sense - I shouldnt have followed the code that we have in LmP :) . will refactor that as well

ldts:
@mike-scott please could you check https://foundriesio.atlassian.net/browse/FFTK-3548 
I think we need your go-ahead before merging this.

ricardosalveti (approved):
Fine with the changes, we need to see if we can find a way to test this as well with @mwasilew.

mwasilew:
At the moment we're not testing the installer at all due to lack of some "preseed" support. Same applies to installing encrypted rootfs.

PR #1525, opened 2024-10-29
efitools: fix do_deploy and keys check

quaresmajose:


quaresmajose:
Also requires https://github.com/foundriesio/lmp-manifest/pull/463

ldts (approved):
${D} can be changed and used as a temporary storage however the new data will not be packed. 
This is not a problem in this case since this firmware is being picked from the deploy folder. However - after our conversation-  I understand it is formally incorrect and confusing to the reader using ${D} that way.

ldts:
@quaresmajose the commit message is wrong. UnLock.efi is created at compile time. It is however _signed_ at install time

quaresmajose:
> @quaresmajose the commit message is wrong. UnLock.efi is created at compile time. It is however _signed_ at install time

right, I will change it to  _signed_ at install time

quaresmajose:
> ${D} can be changed and used as a temporary storage however the new data will not be packed. This is not a problem in this case since this firmware is being picked from the deploy folder. However - after our conversation- I understand it is formally incorrect and confusing to the reader using ${D} that way.

This is not true, ${D} is not a temporary storage and is the image folder of the recipe.
Only the do_install task should create or modify content on that folder.

${D} is the output folder of the do_install
${B} is the output folder of the do_compile
${DEPLOYDIR} is the output folder of the do_deploy

In a perfect world D/B/DEPLOYDIR should be read/write only for the mentioned tasks.
but that's not what happens and these directories can be modified at any time in any task,
just because this is possible doesn't mean it's correct.

there a lot of directories used by bitbake:
```
T = "${WORKDIR}/temp"
D = "${WORKDIR}/image"
S = "${WORKDIR}/${BP}"
B = "${S}"
```
more https://git.yoctoproject.org/poky/plain/meta/conf/bitbake.conf

quaresmajose:
Fixed with https://github.com/foundriesio/meta-lmp/pull/1526

PR #1526, opened 2024-10-29
Unlock efi

ldts:


ldts:
efitools: 

case1
-------
- efitools builds userspace tools and lockdown/unlock EFI programs.
- the efi programs only make sense when providing LmP keys (UEFI_SIGN_ENABLE)
- they can be discarded otherwise as they provide no value for LmP
case2
--------
- The efi applications when built with UEFI_SIGN_ENABLE are installed from the deploy folder
- They are not needed in install
- It should be ok therefore to build and copy them to a temporary folder (WORKDIR) during deploy

ldts:
please let me know if anything is unclear. I'd like to get this in.

meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
```diff
@@ -43,10 +43,15 @@ python do_prepare_local_auths() {
     dir = d.expand('${UEFI_SIGN_KEYDIR}/')
 
     import shutil
+    import os
```
angolini, line 46: typos in the commit message

meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
```diff
@@ -14,10 +14,11 @@ SRC_URI += " \
     file://build-keys-for-lockdown-only.patch \
     file://allow-local-auths.patch \
     file://lockdown.conf \
-    file://unlock.patch \
-    file://unlock.conf \
 "
 
+# UnLock needs the user keys
```
angolini, line 19: I would add a little bit more on the commit log as it looks like to me that the patch does more that only build the keys
And I'm not sure I understand what "build a key" is
Do we need to verify if the key files exists?

meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
ldts: hi daiane, which commit?

meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
ldts: the tool (unlock.efi) requires that the user provides the UEFI keys in order for the tool to be built - because the keys are later on embedded/compiled_in in the coff image.

ldts:
@angolini please let me know if this is enough and thanks for reviewing

meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
angolini: maybe my github displayed an old version when I commented. It's fixed now

angolini (approved):
perfect, thanks <3

PR #1527, opened 2024-10-29
base: initramfs: check for secure boot

ldts:
Using efivar --print-decimal returns an integer.

### DIFF
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/cryptfs_tpm2
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/cryptfs_tpm2
@@ -7,7 +7,7 @@ cryptfs_check_tpm2() {
 	# Check for SecureBoot support as PCR 7 differs based on its state
 	efi_secure=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal`
 	efi_mode=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SetupMode --print-decimal`
-	if [ "${efi_secure}" != "1" ] || [ "${efi_mode}" != "0" ]; then
+	if [ "${efi_secure}" -ne 1 ] || [ "${efi_mode}" -ne 0 ]; then
 		fatal "UEFI SecureBoot not enabled (required due PCR 7)"
 	fi
 
--- a/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
+++ b/meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh
@@ -17,7 +17,7 @@ check_secure_boot() {
 
     efi_secure=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal`
     efi_mode=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SetupMode --print-decimal`
-    if [ "${efi_secure}" != "1" ] || [ "${efi_mode}" != "0" ]; then
+    if [ "${efi_secure}" -ne 1 ] || [ "${efi_mode}" -ne 0 ]; then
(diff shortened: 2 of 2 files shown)


ldts:
@ricardosalveti please could we merge this. I have noticed errors on qemu with the previous validation code. thx.

meta-lmp-base/recipes-core/initrdscripts/initramfs-framework/cryptfs_tpm2
```diff
@@ -7,7 +7,7 @@ cryptfs_check_tpm2() {
 	# Check for SecureBoot support as PCR 7 differs based on its state
 	efi_secure=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal`
 	efi_mode=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SetupMode --print-decimal`
-	if [ "${efi_secure}" != "1" ] || [ "${efi_mode}" != "0" ]; then
+	if [ "${efi_secure}" -ne 1 ] || [ "${efi_mode}" -ne 0 ]; then
```
igoropaniuk, line 10: I think you should drop quotation marks for the first operand, same in `meta-lmp-base/recipes-core/initrdscripts/initramfs-module-install-efi/init-install-efi.sh`

ldts: it doesnt really matter in this particular case - all bash variables are strings anyway. it is safer and good practice to keep the quotes in case there are spaces and so on...

ricardosalveti:
> @ricardosalveti please could we merge this. I have noticed errors on qemu with the previous validation code. thx.

Which errors? Looks like this won't change the actual functionality of the code.

ldts:
I added some traces for you so you can see how bash works. Notice the delimiters around the variable

```
if [ "${efi_secure}" != "1"] || [ "${efi_mode}" != "0" ]; then
       echo "UEFI SecureBoot not enabled (required due PCR 7) ==>  ${efi_secure} ${efi_mode}"
       echo "efi_secure=[$efi_secure]"
       echo "efi_mode=[$efi_mode]"
        exit 1
fi

```

This generates the following error on output even though secure boot was enabled (secure = 1, mode = 0)

```
UEFI SecureBoot not enabled (required due PCR 7) ==>  1  0 
efi_secure=[1 ]
efi_mode=[0 ]
```
basically the trailing space makes the difference. 

the sooner we merge, the better :)

igoropaniuk:
@ldts 

just cross-checked on my laptop, you're right, --print-decimal actually prints a value with an additional trailing space:

```
# testvar=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal` && echo "%$testvar%"
%0 %
```

imo, why not to trim it with xargs for example (so it looks much more obvious what kind of an issue the commit addresses + add that information in commit message):
```
# testvar=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal | xargs` && echo "%$testvar%"
%0%
```

ldts:
> @ldts

> 
> imo, why not to trim it with xargs for example (so it looks much more obvious what kind of an issue the commit addresses + add that information in commit message):
> 
> ```
> # testvar=`efivar --name=8be4df61-93ca-11d2-aa0d-00e098032b8c-SecureBoot --print-decimal | xargs` && echo "%$testvar%"
> %0%
> ```

Using standard solutions to standard problems is usually the better thing.

MrCry0 (commented):
It seems a fix for another commit. Could you please add a "Fixed:" field?

quaresmajose (approved):
LGTM

When I suggested using efivar and tested it, I didn't see the existence of this extra space at the end.

ldts:
> LGTM
> 
> When I suggested using efivar and tested it, I didn't see the existence of this extra space at the end.

yeah dont worry about it :) all we had to do was to protect against these sort of issues which funny enough are not uncommon - I should have noticed

PR #1528, opened 2024-10-31
base: rs: Bump ostreeuploader ver 5cd2cf9

mike-sul:
Relevant changes:
- 5cd2cf9 check: Reduce a number of goroutine to check file existing
- e318913 fiopush: Treat summary update failure as error
- 5991933 push: Decrease number of push goroutines


ricardosalveti (approved):
LGTM

PR #1534, opened 2024-11-08
bsp: dynamic-layers: lmp-boot-firmware: use L4T_VERSION

igoropaniuk:
Use L4T_VERSION variable, and convert it to the same format as it's currenlty exposed via ESRT table:

$ cat /sys/firmware/efi/esrt/entries/entry0/fw_version 2295296


MrCry0 (approved):

igoropaniuk:
@quaresmajose @ricardosalveti good to merge?

PR #1535, opened 2024-11-12
base: systemd-boot display order

ldts:
By using the sort-key[1] label in the conf files, this PR positions ostree deployments above UEFI provisioning applications.


quaresmajose (approved):
LGTM
tested locally and is booting again

PR #1541, opened 2024-11-21
bsp: u-boot-fio: am62xx-evm: drop bootcount config

igoropaniuk:
Drop CONFIG_BOOTCOUNT_* configs, as this is now handled in boot.cmd instead.


ricardosalveti (approved):
LGTM

mwasilew:
I think there is a typo in the commit message. Should be am64

PR #1554, opened 2025-01-15
bsp: alsa-state-stm32mp1: drop bbappend

quaresmajose:
The plan is to include it as a partner layer, this way we can better handle the versioning.


ricardosalveti (approved):
LGTM

PR #1565, opened 2025-01-27
base: kernel-lmp-fitimage: align changes with scarthgap

ricardosalveti:
Align with the kernel-fitimage changes available in scarthgap.

This will help us upstreaming the changes as we can easily generate a diff based on what is available in scarthgap.

### DIFF
--- a/meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
+++ b/meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
@@ -38,7 +38,7 @@ fitimage_emit_section_kernel() {
                 kernel-${2} {
                         description = "Linux kernel";
                         data = /incbin/("${3}");
-                        type = "kernel";
+                        type = "${UBOOT_MKIMAGE_KERNEL_TYPE}";
                         arch = "${UBOOT_ARCH}";
                         os = "linux";
                         compression = "${4}";
@@ -109,6 +109,7 @@ EOF
 	fi
 }
 
+#
 # Emit the fitImage ITS u-boot script section
 #
 # $1 ... .its filename
@@ -233,10 +234,12 @@ EOF
 # $6 ... config ID
 # $7 ... loadable ID (LmP specific)
 # $8 ... default flag
+# $9 ... default DTB image name
 fitimage_emit_section_config() {
 
 	conf_csum="${FIT_HASH_ALG}"
 	conf_sign_algo="${FIT_SIGN_ALG}"
+	conf_padding_algo="${FIT_PAD_ALG}"
 	if [ "${UBOOT_SIGN_ENABLE}" = "1" ] ; then
 		conf_sign_keyname="${UBOOT_SIGN_KEYNAME}"
 	fi
@@ -249,6 +252,7 @@ fitimage_emit_section_config() {
 	config_id="${6}"
 	loadable_id="${7}"
 	default_flag="${8}"
+	default_dtb_image="${9}"
 
 	# Test if we have any DTBs at all
 	sep=""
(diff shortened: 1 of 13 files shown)


ricardosalveti:
This fixes boot on the qemuarm targets, still validating with other platforms.

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
```diff
@@ -278,7 +300,7 @@ fitimage_emit_section_config() {
 	if [ -n "${dtb_image}" ]; then
```
quaresmajose, line 300: this sounds like it should be `dtb_image_sect` as it will be the value used above on the fdt_line var

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
```diff
@@ -480,26 +511,60 @@ fitimage_assemble() {
...
+
+			# Set the default dtb image if it exists in the devicetree.
+			if [ ${FIT_CONF_DEFAULT_DTB} = $DTB ];then
+				default_dtb_image=$(echo "$DTB" | tr '/' '_')
+			fi
+
+			DTB=$(echo "$DTB" | tr '/' '_')
+
+			# Skip DTB if we've picked it up previously
+			echo "$DTBS" | tr ' ' '\n' | grep -xq "$DTB" && continue
```
quaresmajose, line 532: using `&& continue` is prone to fail if the script run with `set -e` and the first part of the expression fail.
the saffer is using `if $exp; then continue;  fi`

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
```diff
@@ -480,26 +511,60 @@ fitimage_assemble() {
...
+		$(find "${EXTERNAL_KERNEL_DEVICETREE}" -name '*.dtbo' -printf '%P\n' | sort); do
+			# Set the default dtb image if it exists in the devicetree.
+			if [ ${FIT_CONF_DEFAULT_DTB} = $DTB ];then
+				default_dtb_image=$(echo "$DTB" | tr '/' '_')
+			fi
+
+			DTB=$(echo "$DTB" | tr '/' '_')
+
+			# Skip DTB/DTBO if we've picked it up previously
+			echo "$DTBS" | tr ' ' '\n' | grep -xq "$DTB" && continue
```
quaresmajose, line 552: same `if $exp; then continue;  fi`

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
```diff
@@ -480,26 +511,60 @@ fitimage_assemble() {
...
+
 			DTBS="${DTBS} ${DTB}"
+
+			# Also skip if a symlink. We'll later have each config section point at it
+			[ $(symlink_points_below $DTB "${EXTERNAL_KERNEL_DEVICETREE}") ] && continue
```
quaresmajose, line 557: same `if $exp; then continue;  fi`

ricardosalveti:
@quaresmajose these are just changes coming from kernel-fitimage, I decided to not change and simply sync so we can then compare and upstream whatever that is specific to us.

If we change common pieces further, it will increase the diff.

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
ricardosalveti: In the end dtb_image_sect depends on dtb_image, and why it should work.

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
ricardosalveti: Yup, something we can suggest upstream.

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
ricardosalveti: Same.

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
ricardosalveti: Same.

meta-lmp-base/classes/kernel-lmp-fitimage.bbclass
quaresmajose: I will do that

quaresmajose:
Understand I will test this PR in my scarthgap factory to see the impact on some other machines

angolini:
I'm still worried when we are going to need to config FIT_CONF_PREFIX again

looks like it's not now?

ricardosalveti:
> looks like it's not now?

yeah, it is not needed anymore, and we can also remove it from upstream since that was due an old u-boot, which is not supported in master

quaresmajose:
I am testing it now on the factory

quaresmajose (approved):
LGTM
all good on the testing factory

PR #1566, opened 2025-01-28
efitools: generate the revocation keys if not present

ldts:
Simplify the backport of the UEFI Secure Boot revocation keys

For background information:
https://foundriesio.atlassian.net/browse/FFTK-3738


ricardosalveti:
Jira link is internal to foundries, please explain the reasoning instead as part of the commit message.

mike-scott:
@ldts I was able to get this to work with the following diff (vs. your patch):
```diff --git a/meta-lmp-base/recipes-bsp/efitools/efitools_git.bb b/meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
index 102ce298..85a58629 100644
--- a/meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
+++ b/meta-lmp-base/recipes-bsp/efitools/efitools_git.bb
@@ -50,8 +50,13 @@ python do_prepare_local_auths() {
         file = _ + '.auth'
         src = dir + file
         if not os.path.isfile(src):
-            bb.fatal("File '%s' not found!" % src)
-        shutil.copyfile(src, d.expand('${S}/') + file)
+            # Don't require no* auth files, generate empty on demand
+            if file.startswith('no'):
+                os.system('echo "" > %s' % (d.expand('${S}/') + file))
+            else:
+                bb.fatal("File '%s' not found!" % src)
+        else:
+            shutil.copyfile(src, d.expand('${S}/') + file)

 }
 addtask prepare_local_auths after do_configure before do_compile```

**Comment by mike-scott** [comment]
Discussion in Slack reveals why my solution won't work.  Thanks @ldts

File: meta-lmp-base/recipes-bsp/efitools/efitools_git.bb [hunk 1]
```diff
@@ -45,14 +45,28 @@ python do_prepare_local_auths() {
...
     for _ in ('PK', 'KEK', 'DB', 'DBX', 'noPK', 'noKEK'):
         file = _ + '.auth'
-        src = dir + file
+        src = os.path.join(dir, file)
         if not os.path.isfile(src):
-            bb.fatal("File '%s' not found!" % src)
-        shutil.copyfile(src, d.expand('${S}/') + file)
-
+            if _ in ('noPK', 'noKEK'):
+                bb.note("Generating '%s' in WORKDIR" % file)
```
quaresmajose, line 54: I don't know is it's better to use a _warning_ there instead of `note` otherwise it will go unnoticed during the build and only be noticeable looking at the entire build log which is not usually done.

ldts: right, I chose a note because IMO users should know what keys they have.

ldts:
@quaresmajose um your comment disappeared...I guess I pressed the wrong button.

ldts:
so, could the accountable entity (you/ricardo/dainae?) confirm if a warning is preferred over a note?

quaresmajose:
> so, could the accountable entity (you/ricardo/dainae?) confirm if a warning is preferred over a note?

I didn't quite understand why it was generated at runtime but after offline explanation I got it.
so it's fine as you have it as it is and doesn't need to be a warning.

quaresmajose (approved):
LGTM

PR #1567, opened 2025-01-29
base: rc: aktualizr: Fix Yocto TMPDIR warning message

detsch:
asn1c tool places the full absolute path for the asn1 files inside comments of the generated source files. Make them relative to ${S} to avoid the warning messages during do_package_qa.

### DIFF
--- a/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
+++ b/meta-lmp-base/recipes-sota/aktualizr/aktualizr_%.bbappend
@@ -38,6 +38,13 @@ do_compile:append:lmp() {
 }
 
 do_install:prepend:lmp() {
+    # asn1c tool places the full absolute path for the asn1 file at the header comment of the generated source files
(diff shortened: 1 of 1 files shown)


angolini:
I understand and agree with https://github.com/foundriesio/meta-lmp/pull/1567#issuecomment-2621479806, but the reason for having it now is that the later PR was not working.

So I would accept this change as an intermediate amend to the problem, and have it reworked properly for future aklite versions (using newer versions of generation tool, or changing that tool)

It looks odd to worry so much about a path on a comment, but the reason behind it is reproducibility.

angolini (approved):

detsch:
> I understand and agree with https://github.com/foundriesio/meta-lmp/pull/1567#issuecomment-2621479806, but the reason for having it now is that the later PR was not working.

That is correct, the change in akltualizr was not enough to fix the issue. Best would be to make asn1c not generate the full path in the comment, but this is not an option.

PR #1568, opened 2025-01-29
base: systemd-boot: display order

ldts:
By using the sort-key [1] label in the conf files, this PR positions ostree deployments above UEFI provisioning applications.

Setting the default sort-key enforces the new systemd-boot style ordering patch.

### DIFF
--- a/meta-lmp-base/recipes-core/systemd/systemd-boot/efi-boot-display-sort-key-default.patch
+++ b/meta-lmp-base/recipes-core/systemd/systemd-boot/efi-boot-display-sort-key-default.patch
@@ -1,0 +1,33 @@
+From 01f64cfa48f5a2b38f35dac1b18482a2707fce0e Mon Sep 17 00:00:00 2001
+From: Jorge Ramirez-Ortiz <jorge@foundries.io>
+Date: Wed, 29 Jan 2025 15:43:31 +0100
+Subject: [PATCH] [PATCH] boot: efi: enable new style display order
+
+The UEFI provisioning applications should be displayed after all other
+entries.
+
+To avoid defaulting to the old display order - causing the provisioning
+applications to raise to the top of the display menu - move undefined
+configs to the top of the display.
+
+Upstream-Status: Inappropriate [lmp specific]
+
+Signed-off-by: Jorge Ramirez-Ortiz <jorge@foundries.io>
+---
+ src/boot/efi/boot.c | 1 +
+ 1 file changed, 1 insertion(+)
+
+diff --git a/src/boot/efi/boot.c b/src/boot/efi/boot.c
+index 83207a5afe..688d9454a2 100644
+--- a/src/boot/efi/boot.c
++++ b/src/boot/efi/boot.c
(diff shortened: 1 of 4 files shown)


igoropaniuk (approved):
LGTM

angolini (approved):

meta-lmp-base/recipes-core/systemd/systemd-boot/efi-boot-display-sort-key-default.patch
```diff
@@ -0,0 +1,33 @@
...
+
+diff --git a/src/boot/efi/boot.c b/src/boot/efi/boot.c
+index 83207a5afe..688d9454a2 100644
+--- a/src/boot/efi/boot.c
++++ b/src/boot/efi/boot.c
+@@ -1430,6 +1430,7 @@ static void boot_entry_add_type1(
+         *entry = (BootEntry) {
+                 .tries_done = -1,
+                 .tries_left = -1,
++                .sort_key = xstrdup16(L"00"),
```
ricardosalveti, line 28: I'm concerned about changing systemd-boot here, and changing the default behavior, as it might bite us in the future.

How is it currently ordering the files without sort key? Can't we name the files in a way they won't show up first?

ricardosalveti: If changing systemd-boot I would prefer an upstream-compatible solution instead (and a backport instead of a lmp-specific patch).

ldts: This is a temporary/simple solution until we adapt the new systemd-boot sorting scheme globally. The problem is our default entries do not define sort-key hence ending at the end of the list under the secure boot ones (the algorithm gives precedence to the new mechanism)

We will not need this default once all of our entries use sort-key but that might take some releases

IMO -the way I have thought about this- our entries could use sort-key "00" for our releases, sort-key "01" for provision and "02" revoke UEFI keys. Then within 00 we should just use the version.

If we dont take a step in this direction by setting a default it will be difficult to ever transition since users will always have legacy entries which will end up in the bottom of the display.

igoropaniuk:
```
$ ls /boot/loader/entries
doprovision_lock+0.conf    doprovision_unlock+0.conf  ostree-1.conf
```

igoropaniuk: Screenshot from 2025-01-29 17-59-37

ldts: I like that - but I dont think both exclude each other (we could have both).  I think we should take steps towards sort-keys hence the need for the default regardless.

regardless, I am ok to withdraw this PR if you want to proceed with that. let me know

PR #1572, opened 2025-01-29
bsp: intel-corei7-64: add 0 tries left suffixes

igoropaniuk:
According to boot entries sorting rules in bootloader spec [1]:
- Entries which are subject to boot counting and are marked as “bad”, should be sorted later than all other entries. Entries which are marked as “indeterminate” or “good” (or were not subject to boot counting at all), are thus sorted earlier.

Mark unlock/lockdown boot entries as bad by adding +0 suffix (0 tries left), as a result the entries will be still shown (so it's possible to select them manually) and always be sorted late.

### DIFF
--- a/meta-lmp-bsp/scripts/lib/wic/plugins/source/bootimg-sota-efi.py
+++ b/meta-lmp-bsp/scripts/lib/wic/plugins/source/bootimg-sota-efi.py
@@ -128,7 +128,7 @@ class BootimgSotaEFIPlugin(SourcePlugin):
 
             # list of tuples (src_name, dst_name)
             deploy_files = []
-            for src_entry in re.findall(r'[\w;\-\./\*]+', boot_files):
+            for src_entry in re.findall(r'[\w;\-\.\+/\*]+', boot_files):
                 if ';' in src_entry:
                     dst_entry = tuple(src_entry.split(';'))
                     if not dst_entry[0] or not dst_entry[1]:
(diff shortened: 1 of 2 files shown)


ldts (commented):
LGTM

igoropaniuk:
I found an issue in building the final wic image, `scripts/lib/wic/plugins/source/bootimg-sota-efi.py` parses that `+` incorrectly, need more time to address this

mike-scott:
> I found an issue in building the final wic image, `scripts/lib/wic/plugins/source/bootimg-sota-efi.py` parses that `+` incorrectly, need more time to address this

Yep, I was about to post this here.  Glad you caught it.

igoropaniuk:
@mike-scott fixed the issue

meta-lmp-bsp/scripts/lib/wic/plugins/source/bootimg-sota-efi.py
```diff
@@ -128,7 +128,7 @@ def do_configure_partition(cls, part, source_params, creator, cr_workdir,
 
             # list of tuples (src_name, dst_name)
             deploy_files = []
-            for src_entry in re.findall(r'[\w;\-\./\*]+', boot_files):
+            for src_entry in re.findall(r'[\w;\-\.\+/\*]+', boot_files):
```
igoropaniuk, line 131: @ricardosalveti does it make sense to upstream that change in the original recipe `./scripts/lib/wic/plugins/source/bootimg-efi.py`?

igoropaniuk: I've sent one https://lists.openembedded.org/g/openembedded-core/message/210442

ricardosalveti: Yup, thanks for sending upstream.

ricardosalveti (approved):
LGTM

PR #1576, opened 2025-02-06
bsp: lmp-machine-custom: keep TI 6.6 in bsp 11.00.01

quaresmajose:
We found some issues on ti-6_12 who is the one that comes by default in the bsp 11.00.01 so still with ti-6_6 for now with the bsp 11.00.01


quaresmajose:
The requirement of this change is to bump the Ti bsp to the latest 11.00.01.

angolini (approved):

PR #1587, opened 2025-02-18
efitools: move the UEFI secure boot provisioning to a bbappend

quaresmajose:
This readd the EFI secure boot provisioning specific only to intel-x86-common machines.

This avoid to use it in CI with overrides
```
UEFI_SIGN_ENABLE:intel-x86-common = "1"
```
and make it possible to just use
```
UEFI_SIGN_ENABLE = "1"
```
that will only work where it is needed, for now just in intel-x86-common


MrCry0:
@quaresmajose Shoudln't the include file name ends with `.inc`? `efitools-UEFI-secure-boot-provisioning.inc` instead of just `efitools-UEFI-secure-boot-provisioning`

quaresmajose:
yes I forgot to push the last change. also have changed the `include` to `require` to fail if the file name is wrong, which is the case you are reporting

MrCry0 (approved):
LGTM, not tested

ldts:
https://github.com/foundriesio/meta-lmp/blob/9853caeaf0de9c6cec2e7e8e510557c971ddb97c/meta-lmp-base/recipes-bsp/efitools/efitools_git.bb#L76 I would just provide empty files in place of the signed ones with a warning when the compiled  files do not have COFF headers. Then if we ever want to add that support is simple to bring it back.

quaresmajose:
I prefer to carry our changes using a bbappend because it becomes easier to understand what we have changed. This will simplify the process of updating the main recipe or even transitioning to using this recipe if it is submitted to oe-core/meta-oe https://github.com/Wind-River/meta-secure-core/issues/26

We can still do the changes to provide empty files in place of the signed ones in this new bbappend.

ldts:
right, we should just make clear (in the commit ?) that efitools is functionally agnostic: will work on any UEFI supported platform independently of the architecture.

The problem is that the makefile doesn't add the COFF header to non-x86 builds - which is why we are discovering the error now and not when the recipe was introduced years ago.

angolini (approved):

quaresmajose:
> right, we should just make clear (in the commit ?) that efitools is functionally agnostic: will work on any UEFI supported platform independently of the architecture.
> 
> The problem is that the makefile doesn't add the COFF header to non-x86 builds - which is why we are discovering the error now and not when this buggy recipe was introduced years ago.

I have added a mention about this on the commit message

PR #1590, opened 2025-02-20
bsp: tegra: systemd-boot: double dtb buffer size

quaresmajose:
This is probably due to fact systemd-boot does not allocate any extra space for the device tree. Without the modification, because there is not any extra space, the UEFI will fail applying the overlay changes from its .dtbo files

Fixes:
```
EFI stub: ERROR: Invalid header detected on UEFI supplied FDT, ignoring ... 
EFI stub: Generating empty DTB
```
References:
https://github.com/anduril/jetpack-nixos/issues/111


ricardosalveti:
Can we test on the lab to see if this fixes the issue?

quaresmajose:
I have tested it on scarthgap factory which has the necessary machinery and tests it in the lab

ricardosalveti (approved):

igoropaniuk (approved):
LGTM

igoropaniuk:
I'll add this also to my backlog, maybe we can somehow address that in the upstream also

quaresmajose:
yeah something good to address upstream

PR #1593, opened 2025-02-21
base: lmp: facilitate changing the default user

quaresmajose:
The default user created in LmP is 'fio' defined on the variable LMP_USER. Changing it also requires a static ID user tables change so it's easier if these tables with the default user can be easily rewritten. To simplify this process we will add the new tables just for the default user.

 files/lmp-passwd-table-default
 files/lmp-group-table-default

To change the default user we need to change the LMP_USER and the new table files according to the desired new user name created.


angolini (approved):
LGTM, let's see what CS has to say
And @kprosise , please be aware that this will probably require change in the documentation

ricardosalveti (approved):

meta-lmp-base/conf/distro/include/lmp.inc
```diff
@@ -144,8 +144,13 @@ ERROR_QA:remove = "version-going-backwards"
 # Required to avoid removing libraries used during signing
 RM_WORK_EXCLUDE += "make-mod-scripts"
 
+# changing the default user also requires overriding the bellow files with the same user and group
```
vanmaegima, line 147:
```suggestion
# changing the default user also requires overriding the below files with the same user and group
```

vanmaegima (approved):
minor comment, lgtm

vanmaegima:
For the doc change, it may fit well here https://docs.foundries.io/latest/user-guide/lmp-customization/lmp-customization.html#lmp-users-and-groups

PR #1603, opened 2025-03-13
systemd updates for scarthgap

ricardosalveti:
- add pstore
- add conditional for pni-names
- enable pni-names by default


ricardosalveti:
@mwasilew this should enable predictable interface naming on scarthgap, as it was enabled in kirkstone.

quaresmajose (approved):
LGTM

PR #1630, opened 2025-04-16
bsp: lmp-machine-custom: intel-corei7-64: make the provision only on sota

quaresmajose:
This breaks the do_image_wic on the lmp-base distro variant

```
| + BUILDDIR=/lmp/build-lmp-base-intel-corei7-64-scarthgap PSEUDO_UNLOAD=1 wic create /lmp/build-lmp-base-intel-corei7-64-scarthgap/tmp-lmp_base/work/intel_corei7_64-lmp-linux/lmp-base-console-image/1.0/systemd-bootdisk-microcode.wks --vars /lmp/build-lmp-base-intel-corei7-64-scarthgap/tmp-lmp_base/sysroots/intel-corei7-64/imgdata/ -e lmp-base-console-image -o /lmp/build-lmp-base-intel-corei7-64-scarthgap/tmp-lmp_base/work/intel_corei7_64-lmp-linux/lmp-base-console-image/1.0/build-wic/ -w /lmp/build-lmp-base-intel-corei7-64-scarthgap/tmp-lmp_base/work/intel_corei7_64-lmp-linux/lmp-base-console-image/1.0/tmp-wic |
| INFO: Creating image(s)...
|
| ERROR: _exec_cmd: install -m 0644 -D /lmp/build-lmp-base-intel-corei7-64-scarthgap/deploy/images/intel-corei7-64/0.conf /lmp/build-lmp-base-intel-corei7-64-scarthgap/tmp-lmp_base/work/intel_corei7_64-lmp-linux/lmp-base-console-image/1.0/tmp-wic/hdd/boot/0.conf returned '1' instead of 0 | output: install: cannot stat '/lmp/build-lmp-base-intel-corei7-64-scarthgap/deploy/images/intel-corei7-64/0.conf': No such file or directory
```

PR #1631, opened 2025-04-21
    bsp: fix oe-core upstream wic bootimg-efi plugin

quaresmajose:
This includes the necessary fixes to make the wic bootimg-efi.py plugin work
with the '+' symbol in the IMAGE_EFI_BOOT_FILES.

We will use our copy of the bootimg-efi renamed to lmp-bootimg-efi
which includes the fix [1].
We will also use our copy of the meta-intel systemd-bootdisk-microcode.wks.in
renamed to lmp-systemd-bootdisk-microcode.wks.in on the intel-corei7-64 machines.
This one uses the wic plugin lmp-bootimg-efi instead of the upstream bootimg-efi.

This is a temporary workaround and can be reverted when the proposed backport [1]
lands the scarthgap branch. Since the plugins are read in the order they appear in
BBLAYERS this is the solution that was found to fix the issue.
[1] https://lists.openembedded.org/g/openembedded-core/message/215044


quaresmajose:
replaces https://github.com/foundriesio/meta-lmp/pull/1630

ricardosalveti:
Backported change now available in oe-core scarthgap.

PR #1650, opened 2025-07-03
bsp: linux-lmp.inc: makes it mandatory

quaresmajose:
On the lmp base layer the linux-lmp.inc already uses the 'require' so the build fail when the file not found.
The same should also happen in the lmp bsp layer.


MrCry0 (approved):

quaresmajose:
> did you happen to face a build error somehow?

No, I don't see any error and I belive we don't have any. But this `linux-lmp.inc` file is a requirment and can't be optional.

angolini:
> No, I don't see any error and I belive we don't have any. But this linux-lmp.inc file is a requirment and can't be optional.

I agree with the requirement, I was only curious to see if there was something else missing!

angolini (approved):

ricardosalveti (approved):

PR #1655, opened 2025-07-31
base/bsp: drop TI bsp

quaresmajose:
The plan is to include it as a partner layer, this way we can better handle versioning.

Searching (find and grep) for the following keywords:
* ti
* am6
* k3
* beagleplay


meta-lmp-bsp/conf/machine/include/lmp-machine-custom.inc
```diff
@@ -1,22 +1,5 @@
 # LMP specific configuration
 
-# Beaglebone
-PREFERRED_PROVIDER_virtual/bootloader:beaglebone-yocto = "u-boot-fio"
```
ricardosalveti, line 4: beagle-yocto is actually from meta-yocto, should be removed separately.

angolini (approved):
I'm ok this this change, however I understood we would keep beagle (as a community board)

quaresmajose:
I have keep it the beaglebone and drop the just beagleplay

PR #1659, opened 2025-08-19
sysctl-hang-crash-helper: use UNPACKDIR and drop PACKAGE_ARCH

ricardosalveti:
Use UNPACKDIR in do_install and drop PACKAGE_ARCH since this is expected to be all (configuration file).


quaresmajose (approved):
LGTM

PR #1660, opened 2025-08-19
base: lmp: drop custom toolchain options, switch to gcc / default

ricardosalveti:
Switch back to the default oe-core toolchain (gcc) as clang still requires a substantial amount of fixes when supporting third party BSP vendors.


mike-scott:
@ricardosalveti Does this also need to be groomed? https://github.com/foundriesio/meta-lmp/blob/main/meta-lmp-base/conf/distro/include/non-clangable.inc

angolini (approved):

PR #1667, opened 2025-09-01
Switch generic-arm64 to systemd-boot

ricardosalveti:


angolini (approved):
I would only add a description to the PR saying the motivation for this PR

PR #1670, opened 2025-09-03
bsp: beaglebone-yocto: drop bt / wl firmware and related scripts

ricardosalveti:
Drop bluetooth and wifi firmware and scripts which are provided by meta-ti, as it is now removed from the standard manifest list.


quaresmajose:
There are still the meta-lmp-bsp/dynamic-layers/meta-ti-extras/ that only contains the wlconf bbappend. We can also drop  the full directory tree.

quaresmajose (approved):
LGTM

angolini (approved):

PR #1676, opened 2025-10-01
bsp: fixup imx drop

quaresmajose:
Fixup the IMX drop in last commit.
- The u-boot-fio-bsp-common.inc was removed
- Remove extra "


mike-scott (approved):
LGTM

PR #1720, opened 2026-03-14
base: compose-apps-early-start: avoid composectl call with empty apps list

MrCry0:
In start_restorable_apps(), when shortlist is empty (all default apps are already present locally) and the compose-apps directory is also empty, the code fell into an else branch and executed:
```
composectl run --apps ""
```
      
Since shortlist is empty, there are no apps to restore. Calling composectl with an empty --apps argument is pointless and may trigger an error that fails the service.

Replace this branch with a log message to make the no-op explicit.


angolini (approved):
Can you describe how you have tested it?

quaresmajose:
> well, this was replaced by #1719 and #1720

There are one commit in this PR1719 that was not merged, only the one in #1719 was merged.

